PR #42, opened 2021-03-24
Expand test case for lines with interpolations

shocklateboy92:
Another bug report dressed as a pull request :zany_face: 
Details are in the commit message.

Don't know if the fix is as straight-forward as the last one - still haven't had a chance to dive into your code much yet.
Let me know if there's anything you want to discuss re. the logic.


belav:
This fix ended up being easy, but it took me a while to figure out where the problem was. The logic for what fits on a line was just ignoring the ForceFlat doc type so all interpolated strings were considered size 0 when the DocPrinter was deciding where to line break. It'll now throw an exception if a doc type isn't explicitly handled to avoid any issues like this going forward.

belav (approved):

PR #45, opened 2021-03-26
Fix files with backslashes being generated on Linux

shocklateboy92:
While there are a lot of backslashes in csproj files everywhere, msbuild seems to do a decent job at converting them automatically.
The ones in CS files directly on the other hand, cause files with literal backslashes in their names to be generated.
```
fernie@tiberius /tmp/scratch/csharpier/Src (git)-[master] % git status 
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   CSharpier.Tests/EncodingTests/USC2LEBOM.actual.cst
        modified:   CSharpier.Tests/Samples/AllInOne.json
        modified:   CSharpier.Tests/Samples/Scratch.json

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        "../Src\\CSharpier.Parser\\SyntaxNodeJsonWriter.generated.cs"
        "../Src\\CSharpier\\SyntaxNodeComparer.generated.cs"
        "../Src\\CSharpier\\SyntaxNodeJsonWriter.generated.cs"

no changes added to commit (use "git add" and/or "git commit -a")
```

AFAIK, Windows accepts `/` as well as `\`, so I think it makes much more sense to use forward slashes. However, you should validate that before merging this in. If it's not the case, let me know and I'll update them to use `Path.DirectorySeparator` instead.

### DIFF
--- a/Src/Worker/GeneratePrinterSwitch.cs
+++ b/Src/Worker/GeneratePrinterSwitch.cs
@@ -47,7 +47,7 @@ namespace CSharpier
 
             var csharpDirectory = Path.Combine(
                 rootDirectory.FullName,
-                @"CSharpier\Printer");
+                "CSharpier/Printer");
             foreach (var file in new DirectoryInfo(csharpDirectory).GetFiles())
             {
                 var name = file.Name.Replace(".cs", "");
--- a/Src/Worker/SyntaxNodeComparerGenerator.cs
+++ b/Src/Worker/SyntaxNodeComparerGenerator.cs
@@ -31,7 +31,7 @@ namespace Worker
                 .ToList();
 
             var fileName =
-                directory.FullName + @"\CSharpier\SyntaxNodeComparer.generated.cs";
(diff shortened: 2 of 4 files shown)


belav:
The forward slashes work on windows.
Another potential thing that may not work on linux -
When one of the tests fail, I have it set up so that https://github.com/VerifyTests/DiffEngine will pop open winmerge automatically. That makes it much easier to see what differences causes one of the tests to fail, and to modify the [TestName].cst file with the changes in [TestName].actual.cst
I modified the way winMerge launches in CSharpier.Tests/TestFiles/SetUpFixture so that the right side is read only instead of the left side. You may need to do something similar for another tool since I'm assuming winmerge isn't available on linux.
You can find the default arguments the tools are launched with in https://github.com/VerifyTests/DiffEngine/tree/master/src/DiffEngine/Implementation

Issue #48, opened 2021-03-27
Support --check


belav:
Add support for a --check option.
This can be used to validate that files are already formatted but will not make any changes to files.

PR #49, opened 2021-03-27
GH-48 adding option for --check

belav:


Src/CSharpier/CommandLineOptions.cs
```diff
@@ -23,6 +24,10 @@ public static RootCommand Create()
                     Arity = ArgumentArity.ZeroOrOne,
                     Description = "A path to a directory containing files to format or a file to format. If a path is not specified the current directory is used"
                 }.LegalFilePathsOnly(),
+                new Option(
```
shocklateboy92, line 27: Have you considered using [System.CommandLine](https://github.com/dotnet/command-line-api/blob/main/docs/Your-first-app-with-System-CommandLine-DragonFruit.md)? It does some real _magic_!

PR #52, opened 2021-03-27
Ditching files that aren't needed in git

belav:


.gitignore
```diff
@@ -25,4 +25,4 @@ nupkg
 dist
 
 *.doctree.txt
-apiKey.txt
```
shocklateboy92, line 28: You sure you meant to get rid of this? Seems like a bad thing to accidentally commit.

belav: I had it stored in a text file at first until I realized I just needed to add it to github as a secret, so I don't think it'll be coming back

PR #57, opened 2021-03-28
Handle non trivial tuple expressions (e.g. with await)

shocklateboy92:


Src/CSharpier/CodeFormatter.cs
```diff
@@ -38,7 +38,7 @@ public CSharpierResult Format(string code, Options options)
                 cancellationToken: cancellationToken
             );
             var syntaxNode = await syntaxTree.GetRootAsync(cancellationToken);
-            if (!(syntaxNode is CompilationUnitSyntax rootNode))
+            if (syntaxNode is not CompilationUnitSyntax rootNode)
```
shocklateboy92, line 41: Unrelated change that snuck in.

Src/CSharpier/Printer/ArgumentListSyntax.cs
```diff
@@ -7,25 +7,39 @@ public partial class Printer
...
+                    )
+                )
+                : PrintArgumentListLikeSyntax(
+                    node.OpenParenToken,
+                    node.Arguments,
+                    node.CloseParenToken
+                );
+        }
+
+        private Doc PrintArgumentListLikeSyntax(
```
shocklateboy92, line 25: Not sure where you'd want to extract shared code between syntax tree expressions. I put it here because it's the most related thing I could think of, but feel free to suggest elsewhere.

belav: Right now it is a mix between methods in PrinterHelpers and methods directly in the Printer.cs class.
I think I lean towards keeping them in PrinterHelper files. At some point PrinterHelpers didn't exist, but the Printer.cs file was getting too big so I pulled things out as I worked on them.

shocklateboy92: Okay, I moved it to `PrinterHelpers/`. We should find a naming convention that lets us easily distinguish the methods that handle a Roslyn syntax node vs these helpers.

Personally, I'm not a big fan of the term "helper", because it's such a generic term that doesn't really mean anything. But I don't have any real suggestions yet, so I'll go with it for now.

belav (approved):

Issue #22, opened 2021-03-01
Leading/trailing trivia Formatting


belav:
We lose new lines in places we may want them, and insert them in places we probably shouldn't.
We also gain extra new lines at times.

We lose the new line after foo
```c#
#define foo

namespace Namespace
```

We add an extra new line before #pragma each time we format - https://github.com/belav/csharpier/issues/21
```c#
namespace Namespace
{
#pragma
    public class ClassName { }
}
```

There are some forms of comments we want inline, something like this breaks I think.
```c#
this.DoSomething(1, /* no break */ 2);
```

belav:
https://github.com/belav/csharpier/issues/32 is related, and maybe can be handled in this issue.

PR #59, opened 2021-03-29
GH-22 figuring out a (kind of ugly) solution to get new lines working in trivia

belav:


belav:
There are more edge cases I missed, ugh

shocklateboy92 (approved):
Okay, I read through your long ass comment and all I got from it is that this logic is rather finicky. :face_with_head_bandage:

Here's a rubber stamp.

belav:
I've been debating redoing how to print trivia completely. But don't really have a good idea of how it would work, and this got things basically working the way they should.

Issue #17, opened 2021-03-01
Query Expression Formatting


belav:
Review formatting of longer query expressions and determine how they should break.

PR #60, opened 2021-03-29
GH-17 query expression formatting

belav:
See test file for examples. I don't often use query expressions myself so these may require some more tweaking.

### DIFF
--- a/Src/CSharpier/Printer/JoinClauseSyntax.cs
+++ b/Src/CSharpier/Printer/JoinClauseSyntax.cs
@@ -7,21 +7,39 @@ namespace CSharpier
         private Doc PrintJoinClauseSyntax(JoinClauseSyntax node)
         {
             return Group(
-                this.PrintSyntaxToken(node.JoinKeyword, " "),
-                this.PrintSyntaxToken(node.Identifier, " "),
-                this.PrintSyntaxToken(node.InKeyword, " "),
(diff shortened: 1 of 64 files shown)


Src/CSharpier/Printer/JoinClauseSyntax.cs
```diff
@@ -6,24 +6,26 @@ public partial class Printer
...
-                node.Into != null
-                    ? Concat(
-                        SpaceIfNoPreviousComment,
-                        this.PrintSyntaxToken(node.Into.IntoKeyword, " "),
-                        this.PrintSyntaxToken(node.Into.Identifier)
-                    )
-                    : Doc.Null
+                Indent(
+                    Line,
+                    this.PrintSyntaxToken(node.OnKeyword, " "),
```
shocklateboy92, line 16: This might be somewhat annoying, but I think this could would be much easier to parse for humans if you used `afterTokenIfNoTrailing: " "` as a named parameter. It's not really intuitive without looking at the definition.

shocklateboy92 (approved):
Seems legit, though I did have one suggestion that would make this code annoying to write but easier to read.

shocklateboy92:
Also, does this close #17 ? If so, you should really link it in your commit message:
https://stackoverflow.com/a/6742691

Issue #32, opened 2021-03-14
Block with just a comment doesn't indent the comment


belav:
```c#
if (someBoolean)
{
// code using value
}
```
The comment is leading trivia on the closing brace. Currently leading trivia ends up indented at the same level as the token it is on.
Because we are changing line breaks/indenting we can't really depend on the existing whitespace in front of the comment.
This probably needs to be handled as a special case somehow. There may be other special cases like this.

belav:
Figure out https://github.com/belav/csharpier/issues/22 first just in case that affects this.

PR #62, opened 2021-03-29
GH-32 indent comments inside closing brace

belav:


Src/CSharpier/DocPrinter.cs
```diff
@@ -71,53 +77,53 @@ private Indent MakeIndent(Indent indent, Options options)
                 }
             }
 
-            flushSpaces();
+            FlushSpaces();
 
-            void addTabs(int count)
+            void AddTabs(int count)
```
shocklateboy92, line 82: Did you also port these local functions from prettier? It might be more C# like to make them instance methods and extract this into its own class. What do you think?

Src/CSharpier/DocPrinter.cs
```diff
@@ -71,53 +77,53 @@ private Indent MakeIndent(Indent indent, Options options)
...
             {
                 value += new string('\t', count);
                 length += options.TabWidth * count;
             }
 
-            void addSpaces(int count)
+            void AddSpaces(int count)
             {
                 value += new string(' ', count);
```
shocklateboy92, line 90: Also, you should probably a `StringBuilder()` rather than string, since all you do are appends.
Also, I think it'll probably be faster to `Append(char)` in a loop than allocate a new string of the right length:
https://www.meziantou.net/stringbuilder-performance-pitfalls.htm#use-append-char-inst

shocklateboy92 (commented):
I didn't do the rubber stamp this time, as a forcing function for me to dig into the code more.
I did however, make a lot of suggestions (including comments about `StringBuilder` without doing any performance benchmarks :innocent:).

Feel free to address in a separate PR (or ignore completely) all the comments _except_ for the one I made about your new `AddExtraIndent` property.

Src/CSharpier.Tests/TestFiles/Comments/_CommentsTests.cs
```diff
@@ -20,5 +20,10 @@ public void MethodComments()
         {
             this.RunTest("Comments", "MethodComments");
         }
+        [Test]
```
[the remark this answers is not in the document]
belav: I've been debating that recently. It doesn't look like prettier does. And if I'm reading the black docs correctly, I don't think they do either, except in some specific situations. So I'm kinda leaning towards no it shouldn't.

Src/CSharpier/Doc.cs
```diff
@@ -92,6 +92,7 @@ public class LeadingComment : Doc
     {
         public CommentType Type { get; set; }
         public string Comment { get; set; } = string.Empty;
+        public bool AddExtraIndent { get; set; }
```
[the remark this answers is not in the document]
belav: For some reason I didn't think to try just using Indent. Using it did require some changes but this keeps the core code much cleaner.
I also played around with LeadingComment, and it looks like it can probably go away. With the changes I made https://github.com/belav/csharpier/issues/22 it seems like it isn't necessary. It wasn't completely straightforward to remove them, so I created a new issue for it - https://github.com/belav/csharpier/issues/67

Src/CSharpier/DocPrinter.cs
belav: Looks like StringBuilder actually has .Append(' ', count)

Src/CSharpier/DocPrinter.cs
belav: A good chunk of this was ported right from prettier. I've been debating doing some major refactoring to it because I always get lost inside of it. But I think first I should figure out if any of the other features from prettier are needed.

Src/CSharpier/DocPrinter.cs
```diff
@@ -63,11 +63,11 @@ private Indent MakeIndent(Indent indent, Options options)
                             AddSpaces(options.TabWidth);
                         }
                         break;
-                    case "stringAlign":
-                        Flush();
-                        value += part.Number;
-                        // TODO 2 huh? length += part.n.length;
-                        break;
+                    // case "stringAlign":
```
shocklateboy92, line 66: Did you forget about this, or are you leaving it as a todo reminder? 
If it's the latter, I think you're better of just having a TODO comment. Commented-out-code just isn't very useful with revision control.

Src/CSharpier/CodeFormatter.cs
```diff
@@ -173,11 +173,11 @@ private string PrintDocTree(Doc document, string indent)
                         indent + "    "
                     ) + ")";
                 case LeadingComment leadingComment:
-                    return indent + "LeadingComment(" + leadingComment.Comment + ", CommentType." + (leadingComment.Type == CommentType.SingleLine
+                    return indent + "LeadingComment(\"" + leadingComment.Comment + "\", CommentType." + (leadingComment.Type == CommentType.SingleLine
```
shocklateboy92, line 176: You might find these easier to do with string interpolations.

Also, is this meant to output valid C# code that you can paste into unit tests or something? If not, you might find it easier to use single quotes rather than double; that way you don't have to escape the quotes all the time.

shocklateboy92 (approved):

PR #75, opened 2021-04-10
Adding ifBreak doc type

### LINKED ISSUE #34: If Statement with long expression
Currently when an if statement has a long expression, it formats like
```c#
if (
    !this.model.TranslatableFields.TryGetValue(
        fieldName,
        out var dictionary
    )
)
{
    return;
}
```
I think it should do this
```c#
if (
    !this.model.TranslatableFields.TryGetValue(
        fieldName,
        out var dictionary
    )
) {
    return;
}

if (shortStatement)
{
    return;
}
```
or
```c#
if (
    !this.model.TranslatableFields.TryGetValue(
        fieldName,
        out var dictionary
    ))
{
    return;
}

if (shortStatement)
{
    return;
}
```

This applies to other syntax nodes, like for loops and while loops.

belav:
changing if, while, foreach and using to use SpaceBrace formatting.
closes #34


Src/CSharpier/DocPrinter.cs
```diff
@@ -11,6 +11,9 @@ namespace CSharpier
     // taken from prettier 2.2.1 or so
     public static class DocPrinter
     {
+        [ThreadStatic]
+        private static Dictionary<string, PrintMode>? groupModeMap;
```
belav, line 15: It feels a little weird to have a ThreadStatic like this. I could also change DocPrinter to not be static to avoid this.

Src/CSharpier/GroupIdGenerator.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+
+namespace CSharpier
+{
+    public static class GroupIdGenerator
```
belav, line 7: I'm not thrilled with this class.
I made it static because if we start to split apart printer, it is probably going to move towards static methods.
If we have to instantiate a bunch of different classes in each print method that is going to get obnoxious.
How this gets used also feels a little fragile, but I wasn't sure how else to get a groupId inside of BlockSyntax.
Although, maybe I should just add a method overload like so.... yeah duh. Possibly naming it better while I'm at it.
```c#
private Doc PrintBlockSyntax(BlockSyntax node) => PrintBlockSyntax(node, null);

private Doc PrintBlockSyntax(BlockSyntax node, string? groupId)
{
...
```

Src/CSharpier/Doc.cs
```diff
@@ -25,6 +42,13 @@ public HardLine()
             ) { }
     }
 
+    public class IfBreak : Doc
+    {
+        public Doc FlatContents { get; set; } = Docs.Null;
+        public Doc BreakContents { get; set; } = Docs.Null;
+        public string? GroupId { get; set; }
```
belav, line 49: Maybe this should just be a Guid?

PR #80, opened 2021-04-11
arrow expression methods

### LINKED ISSUE #58: Lambda Body Method Formatting
```c#
// currently formats like
private Doc MethodWithParameters(
    SomeClass one,
    SomeClass two,
    SomeClass three) =>
    CallSomethingElseWithALongNameThatForcesABreak(one, two, three);

// maybe like this
private Doc MethodWithParameters(
    SomeClass one,
    SomeClass two,
    SomeClass three) =>
        CallSomethingElseWithALongNameThatForcesABreak(one, two, three);
// or this is how prettier does it, which I think I prefer
const doSomething = (
  jasdjklfasjkldflajskdflkasdlfjkasdf,
  jalsdfkasdlkfalksflksdf,
  kjladslkfasjfkasdfl
) => doSomethingElse();

// but also don't indent more when
private Doc ShortMethod(SomeClass one) =>
    CallSomeOtherLongMethodSoThatItBreaks(one);

// but probably always break when
private Doc MethodWithParameters(
    SomeClass one,
    SomeClass two,
    SomeClass three
) => CallSomething(one, two, three);
```

belav:
closes #58


belav:
I left some tests failing based on thinking we should use SpaceBrace for Methods etc. If true, we'll need to finish up https://github.com/belav/csharpier/issues/34 before we can finish this up.

PR #84, opened 2021-04-12
Getting a basic version of grouping binary expressions working.

belav:
Playing around with an idea for how to start pulling apart the partial class


belav:
Some of this is a little ugly with how SyntaxNodes and Printer interact, but the idea is that it is just temporary and provides a way to start moving away from the partial class without having to do it all at once.
I'm also not sure about the naming and it's currently a little inconsistent.
BinaryExpression.Print(node)
vs
SyntaxNodes.Print(node)
SyntaxTokens.Print(token)
Maybe all the class names should be pluralized? It sounds a little weird, but we can't name it SyntaxNode and SyntaxToken without colliding with existing names.
Or we can just come up with different names for SyntaxNodes and SyntaxTokens. Node.Print, Token.Print

PR #105, opened 2021-04-17
Testing out what it might be like to pass IList<Doc> down to each method to simplify doc tree

belav:
If we pass IList<Doc> to each method, we can simplify the doc tree. We won't end up with things like
```
Docs.Concat(
    Docs.Concat(
        "one"
    ),
    Docs.Concat(
        "two"
    )
)
// would instead end up
Docs.Concat(
    "one",
    "two"
)
```
I did a little POC to see what some methods may look like, and benchmarked the changes.
The changes shaved ~15% off of the benchmark, but the benchmark was not reading/writing files.
When I included reading/writing files, it shaved 2%.
I didn't quite take it as far as I could with the code I was using. So theoretically it would be a tiny bit faster.
Being forced to pass IList<Doc> to methods also makes the code a lot less clean.
I'm thinking it isn't worth the time and ugly code just to make things a little bit faster and make our doc tree a little bit easier to read.

### DIFF
--- a/Src/CSharpier/Printer/ParameterListSyntax.cs
+++ b/Src/CSharpier/Printer/ParameterListSyntax.cs
@@ -1,5 +1,7 @@
+using System.Collections.Generic;
 using CSharpier.DocTypes;
 using CSharpier.SyntaxPrinter;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace CSharpier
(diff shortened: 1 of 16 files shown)


Src/CSharpier/Printer/ParameterListSyntax.cs
```diff
@@ -10,24 +12,24 @@ public partial class Printer
...
+                    Docs.Line,
+                    innerDocs
+                );
+
+                docs.Add(Docs.Indent(innerDocs));
+                docs.Add(Docs.SoftLine);
+            }
+            SyntaxTokens.Print(node.CloseParenToken, docs);
+
+            return Docs.GroupWithId(groupId ?? string.Empty, docs);
```
belav, line 32: This is pretty gross looking now.

PR #114, opened 2021-04-18
Fixing issue with ANSI files

### LINKED ISSUE #109: NonEnglish characters cause file to not format.
When running from the command line \AutoMapper\src\UnitTests\Internationalization.cs fails to compile
When pasting the contents of it into the playground, it compiles and formats.

belav:
Abstracted away file reading so that unit tests can read files the same way as the CLI
Moved file encoding tests to same style of unit tests, they weren't actually testing how the CLI read files previously.
Modified unit tests to fail on compilation errors which uncovered some bugs.
closes #109


Src/CSharpier/FileReader.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                defaultedEncoding = true;
+                encoding = Encoding.Default;
+            }
+
+            var fileContents =
+                await File.ReadAllTextAsync(
+                    file,
+                        // this fixes an issue with ANSI encoded files like csharpier-repos\AutoMapper\src\UnitTests\Internationalization.cs
+                        encoding.CodePage
```
shocklateboy92, line 34: Unrelated to this PR I suppose: why is this argument indented? Because of the comment above?

shocklateboy92 (approved):

Src/CSharpier/FileReader.cs
belav: The comment does seem to cause it. It may have to do with the changes I made a while ago to get comments to indent when they are attached to a closing brace. I created https://github.com/belav/csharpier/issues/116 for it.

Issue #43, opened 2021-03-24
Support for options in a config file.


belav:
When there are options available for formatting - https://github.com/belav/csharpier/issues/10 - it should be possible to specify those options in a config file. That will ensure that if csharpier is run via command line, build task, or some other way, it will use the same set of options without having to duplicate them.

belav:
This is partially implemented, just so that we can ignore some files in csharpier-repos.
When we implement it for real, we should also probably support looking for a file in parent directories.
Scan up until you find one or until you hit the root.
If we do that though, the current exclude option won't work, because it expects paths relative to where the file is.

belav:
We may want to implement ignored/exluded files using a .csharpierignore file instead. Or we could potentially support both, if the file exists in either exclude of is in a .csharpierignore, then ignore it.
We should also support YAML as well as JSON to give the people options.

PR #126, opened 2021-04-19
add way to ignore files, option for skip-write, some refactoring

belav:


Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -0,0 +1,323 @@
...
+                    Interlocked.Increment(ref this.UnformattedFiles);
+                }
+            }
+
+            cancellationToken.ThrowIfCancellationRequested();
+            Interlocked.Increment(ref this.Files);
+
+            if (!this.Check)
+            {
+                // purposely avoid async here, that way the file completely writes if the process gets cancelled while running.
```
shocklateboy92, line 227: Do you know if that's true?

Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -0,0 +1,323 @@
...
+                .Substring(this.RootPath.Length + 1);
+
+            if (
+                this.ConfigurationOptions.Exclude.Contains(normalizedFilePath)
+            ) {
+                return true;
+            }
+
+            return normalizedFilePath.EndsWith(".g.cs")
+            || normalizedFilePath.EndsWith(".cshtml.cs")
```
shocklateboy92, line 299: These should have been indented by csharpier. Probably a bug.

Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -0,0 +1,323 @@
...
+                    }
+                }
+            }
+
+            PrintResults();
+
+            return ReturnExitCode();
+        }
+
+        private async Task DoWork(
```
shocklateboy92, line 103: `DoWork` doesn't seem like a great name.
How about `FormatFile()`?

Src/CSharpier/ConfigurationOptions.cs
```diff
@@ -0,0 +1,40 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Newtonsoft.Json;
+
+namespace CSharpier
+{
+    public class ConfigurationOptions
+    {
+        public HashSet<string> Exclude { get; set; }
```
shocklateboy92, line 10: If configuration options only have `Exclude`, why not just have a `.csharpierignore` file? 

Then like `.gitignore`s, there can be multiple in a project, and scoped down to a given directory.

Src/CSharpier/ConfigurationOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public static ConfigurationOptions Create(string rootPath)
+        {
+            var directoryInfo = new DirectoryInfo(rootPath);
+
+            var nextPath = Path.Combine(directoryInfo.FullName, ".csharpierrc");
+            if (File.Exists(nextPath))
+            {
+                return JsonConvert.DeserializeObject<ConfigurationOptions>(
```
shocklateboy92, line 32: Relatively minor point, but I would much prefer yaml over json - it doesn't require a quotes/commas and _allows comments_ in the spec, so all editors support them.

.github/workflows/format_repositories.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       -
```
shocklateboy92, line 18: Nit: you don't need a blank link when starting a new object in an array in yaml. I think this would look better:
```yaml
    - run: dotnet build csharpier/Src/CSharpier/CSharpier.csproj -c release
```

Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -0,0 +1,323 @@
...
+        }
+
+        private static void PrintResultLine(string message, int count)
+        {
+            Console.WriteLine(
+                PadToSize(message + ": ", 80) + ReversePad(count + "  ")
+            );
+        }
+
+        private bool IgnoreFile(string filePath)
```
shocklateboy92, line 287: This whole thing could be much simpler if you used a [Matcher](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.filesystemglobbing.matcher?view=dotnet-plat-ext-5.0)

You can construct one, add `**/*.cs` as an include, add your `**/*.g.cs` (etc. along with all the user specified ones) as excludes, and just call evaluate. It doesn't go to disk and expects you to provide the list of files, but that should be easy since you already have them all in memory already.

Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -0,0 +1,323 @@
...
+            if (fileReaderResult.DefaultedEncoding)
+            {
+                Console.WriteLine(
+                    $"{GetPath()} - unable to detect file encoding. Defaulting to {fileReaderResult.Encoding}"
+                );
+            }
+
+            cancellationToken.ThrowIfCancellationRequested();
+
+            CSharpierResult result;
```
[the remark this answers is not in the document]
belav: Looks that way.
If I add this below it, it results in a build error
result.Code = "x";

Src/CSharpier/CommandLineFormatter.cs
belav: Yeah. Not every time, but it probably happened every other time I cancelled the process when I was testing.

Src/CSharpier/ConfigurationOptions.cs
belav: Good call. This was just my quick hack so I could get the format repos workflow passing. I updated https://github.com/belav/csharpier/issues/43 with the idea

Src/CSharpier/ConfigurationOptions.cs
belav: I hate that json doesn't allow comments but don't use yaml much so usually default to json. I think we could pretty easily support both and updated https://github.com/belav/csharpier/issues/43

shocklateboy92:
You should definitely use yaml more :D
I think you'll start to really like it - for hand-written files, yaml is
much nicer than json.

Src/CSharpier/ConfigurationOptions.cs
shocklateboy92: You should definitely use YAML more. :grin: 
I think you'll start to like it - for hand-written files, it's much nicer than JSON.

PR #138, opened 2021-04-24
move solution, adding solution items, minor cleanup

belav:


shocklateboy92 (commented):
Seems Legit. Though, I don't use Visual Studio so I have no feedback regarding the changes to stuff in the `.sln`

PR #139, opened 2021-04-24
formatting with 0.9.0

belav:


shocklateboy92 (commented):
Okay, the more I look at the resulting changes, the more I think that the operator for a binary expression should be in the new line.

There's no indent in these, so when skimming through it isn't immediately apparent that the new line is a continuation of the previous line's content. 
Putting the operator at the beginning of the line does many things in one go. It:
- Gives us a "semi-indent" for free (using less characters as well)
- Let's you know this is part of the previous statement while skimming through
- Makes it immediately apparent _how_ this is related to the previous line. (E.g. `+` or `&&` etc.)

PR #145, opened 2021-04-25
more work on ditching the partial class

belav:


Src/CSharpier/DocTypes/Doc.cs
```diff
@@ -8,6 +11,164 @@ public class Doc
...
+            {
+                Contents = contents.Length == 0
+                    ? contents[0]
+                    : Concat(contents),
+            };
+        }
+
+        public static Group Group(List<Doc> contents)
+        {
+            return new()
```
shocklateboy92, line 100: Minor point, but I think these might look better with the expression body syntax

PR #153, opened 2021-04-26
breaking before binary operator

### LINKED ISSUE #152: Break before BinaryOperator
We've decided we want to break before BinaryOperators.
See discussion at https://github.com/belav/csharpier/issues/117

belav:
closes #152


belav:
merging because it's simple and another PR depends on it

PR #185, opened 2021-05-06
Implementing auto line ending

### LINKED ISSUE #147: Support "auto" for endOfLine
For the endOfLine option implemented in https://github.com/belav/csharpier/issues/10 we can add

"auto" - Maintain existing line endings (mixed values within one file are normalised by looking at what’s used after the first line)

belav:
closes #147

### DIFF
--- a/Src/CSharpier/ConfigurationFileOptions.cs
+++ b/Src/CSharpier/ConfigurationFileOptions.cs
@@ -15,7 +15,7 @@ namespace CSharpier
         public int PrintWidth { get; init; } = 100;
         public int TabWidth { get; init; } = 4;
         public bool UseTabs { get; init; }
-        public string EndOfLine { get; init; } = "lf";
+        public string EndOfLine { get; init; } = "auto";
 
         public static ConfigurationFileOptions Create(string rootPath, IFileSystem fileSystem)
         {
--- a/Src/CSharpier/DocPrinter/DocPrinter.cs
+++ b/Src/CSharpier/DocPrinter/DocPrinter.cs
@@ -145,14 +145,13 @@ namespace CSharpier.DocPrinter
             return false;
         }
 
-        public static string Print(Doc document, PrinterOptions printerOptions)
+        public static string Print(Doc document, PrinterOptions printerOptions, string endOfLine)
         {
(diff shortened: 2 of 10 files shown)


Src/CSharpier/ConfigurationFileOptions.cs
```diff
@@ -15,7 +15,7 @@ public class ConfigurationFileOptions
         public int PrintWidth { get; init; } = 100;
         public int TabWidth { get; init; } = 4;
         public bool UseTabs { get; init; }
-        public string EndOfLine { get; init; } = "lf";
+        public string EndOfLine { get; init; } = "auto";
```
shocklateboy92, line 18: Why did you make this a string instead of just putting the enum value here?
(Both the Json/Yaml parsers should be able to handle that)

Src/CSharpier/DocPrinter/DocPrinter.cs
```diff
@@ -194,7 +193,7 @@ void Push(Doc doc, PrintMode printMode, Indent indent)
                         {
                             // TODO 1 new line stuff
                             TrimOutput(output);
-                            output.Append(newLine + command.Indent.Value);
+                            output.Append(endOfLine + command.Indent.Value);
```
shocklateboy92, line 196: Nit: it's faster to call `StringBuilder.Append()` twice than doing a string concat in its argument

Src/CSharpier/Program.cs
```diff
@@ -67,14 +67,16 @@ CancellationToken cancellationToken
...
-                    ? "\n"
-                    : configurationFileOptions.EndOfLine == "crlf"
-                            ? "\r\n"
-                            : throw new Exception(
-                                    "Unhandled value from EndOfLine options "
-                                    + configurationFileOptions.EndOfLine
-                                )
+                EndOfLine = configurationFileOptions.EndOfLine == "auto"
+                    ? EndOfLine.Auto
+                    : configurationFileOptions.EndOfLine == "lf"
```
shocklateboy92, line 72: This will probably look much better with pattern matching (`match {}` statement) than nested ternary operators.

PR #191, opened 2021-05-14
Formatting uses proper line endings on verbatim strings

belav:
I'm not super happy with either of these, and neither of them really solves the issue of line endings in verbatim strings affecting print width. One just happens to fix the unit test I wrote for it.


Src/CSharpier/SyntaxPrinter/Token.cs
```diff
@@ -58,7 +57,20 @@ public static Doc PrintWithSuffix(SyntaxToken syntaxToken, Doc suffixDoc)
 
             ShouldSkipNextLeadingTrivia = false;
 
-            docs.Add(syntaxToken.Text);
+            if (
+                syntaxToken.Kind() == SyntaxKind.StringLiteralToken
+                && syntaxToken.Text.StartsWith("@")
+            ) {
+                var lines = syntaxToken.Text.Split(
```
shocklateboy92, line 64: The more I think about it, the more convinced I'm becoming that we should leave verbatim strings alone. 
We don't know what environment/context the user code is being used in, and changing its behavior could introduce bugs for them. Also, csharpier's job is to change the code's formatting, not its behavior.

For that, I think the simplest approach would be to create a `StringDoc` containing the entire `Text` of the `StringLiteralToken` and wash our hands of it. 

And then change the tests to expect them to be "unformatted" essentially.

belav: I disagree. When a user writes a verbatim string, they aren't really specifying the line endings. Their editor is doing that for them. If they want specific line endings they would be doing something like this. In this case CSharpier won't change anything.
```c#
@"one\ntwo\r\nthree"
```

I also dug into how prettier deals with template literals, which are kind of like a verbatim interoplated string. When I format a file back and forth between crlf and lf in prettier, it affects the template literal.
image
If CSharpier doesn't touch the line endings in a verbatim string, then a file could end up with mixed line endings. Which is what I was ending up with a week ago
image

belav: I realized today that .editorconfig supports specifying line endings.
When I use the following in .editorconfig
```
end_of_line = lf
```
And format a file that contains a verbatim string with crlf, they get replaced with lf.
Based on the fact that both prettier and dotnet-format adjust line endings in string literals that span multiple lines, I think csharpier replacing them is the right approach.

shocklateboy92: I suppose. I forgot that git commits LF by default.
Okay, let's normalize them then.

PR #195, opened 2021-05-15
Fixing extra space in pattern matching

### LINKED ISSUE #167: PatternMatching extra space
```c#
if (
    doc is HardLine { SkipBreakIfFirstInGroup: true }  &&
    newGroup
) {
```

belav:
closes #167


shocklateboy92 (commented):
Not sure what the "before" case was like, but I like the "after".

PR #199, opened 2021-05-16
properties with modifiers shouldn't break unless they hit width

### LINKED ISSUE #188: Formatting of auto implemented properties with access modifiers
```csharp
//Is
public class ClassName {
    public string ShortPropertyName {
        get;
        private set; 
    }
}
//ShouldBe
public class ClassName {
    public string ShortPropertyName { get; private set; }
}
```

The latter seems more idiomatic

belav:
closes #188


shocklateboy92 (commented):
LGTM :+1:

Issue #181, opened 2021-05-06
CSharpierIgnore from parent directories should be respected.


belav:
One example of why this is an issue - when a file watcher is used in rider, it doesn't respect the ignore/options that are configured.

PR #200, opened 2021-05-16
Add package with build tasks to format on build

shocklateboy92:
My coworkers are (pretty much) guaranteed forget to run the dotnet tool command, so I want to run it for them automatically during build.

This creates a nuget package that injects a build task that formats before build. Assuming you're fine with the approach, you'll need to update the metadata of the package and update the github actions to publish it too.


belav:
As someone who is guilty of forgetting to prettier my files all the time at work without a file watcher, I love it.

Personally, I'd want to run this for debug, and then run it with --check during release. Although doing the --check as a build step might be a little weird because it makes it harder for someone to run release code locally.

My original thought was we should add some properties to support situations like that.
```xml
<PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
  <CSharpier_Disable>true</CSharpier_Disable>// maybe CSharpier_Enable?
  <CSharpier_Check>true</CSharpier_Check>
</PropertyGroup>
```

On the other hand you can always run the check as a separate github action, like `.github/workflows/validatePR.yml`, so maybe we don't need to worry about adding properties right now.

I believe this depends on https://github.com/belav/csharpier/issues/181 to support .csharpierrc and .csharpierignore, which I was going to work on in the next day or two.

belav:
I think this is what is needed to publish the msbuild project. I haven't tested out trying to reference this nuget package in a project yet.
The readme should also get updated, but I'm about to split that apart

Src/CSharpier.MsBuild/CSharpier.MsBuild.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <Import Project="../../CSharpier.Targets"/>
+    <PropertyGroup>
+        <TargetFramework>net5.0</TargetFramework>
```
shocklateboy92, line 4: You need the `netstandard2.0` here. Otherwise it can't be used to format projects that don't also target `net5.0` themselves

belav:
closes #218

PR #206, opened 2021-05-17
switching to webpack + fixing issue with SyntaxTree rerendering json tree way too often

belav:


shocklateboy92 (commented):
Why don't you just use create react app? It gives pretty much everything we need out of the box they will continue maintain our webpack config for us.

Doing all this by hand takes a lot of work and I don't think it'll really provide any tangible benefit...

PR #207, opened 2021-05-17
Support looking for config and ignore files in parent directories

### LINKED ISSUE #181: CSharpierIgnore from parent directories should be respected.

belav:
closes #181

### DIFF
--- a/Src/CSharpier/CommandLineFormatter.cs
+++ b/Src/CSharpier/CommandLineFormatter.cs
@@ -6,7 +6,6 @@ using System.IO.Abstractions;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using Ignore = Ignore.Ignore;
 
 namespace CSharpier
 {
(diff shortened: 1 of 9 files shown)


Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -20,44 +19,93 @@ public class CommandLineFormatter
...
+                    $"The path of {commandLineOptions.DirectoryOrFile} does not appear to point to a directory or a file."
+                );
             }
 
+            var configurationFileOptions = ConfigurationFileOptions.Create(
+                baseDirectoryPath,
+                fileSystem
+            );
+
+            var (ignoreFile, exitCode) = await CSharpier.IgnoreFile.Create(
```
shocklateboy92, line 69: So, returning a tuple with the result and a flag representing success (as an int) isn't a great pattern. 

One way would be to have multiple types (each representing a different status code, along with the result for the success case) that inherit from a common base, and return said base. Then you can pattern match on the different subtypes and proceed accordingly.

However, in this case it looks like you're only using the values `0` and `1` to represent success and failure. In this case, the nullability of the result already achieves that quite well, so the whole tuple is redundant. You can just return `1` below if the `ignoreFile` is `null`.

Src/CSharpier/ConfigurationFileOptions.cs
```diff
@@ -17,58 +17,68 @@ public class ConfigurationFileOptions
...
-                    fileSystem.File.ReadAllText(path)
-                );
-            }
-
-            ConfigurationFileOptions ReadYaml(string path)
+        public static ConfigurationFileOptions Create(
+            string baseDirectoryPath,
+            IFileSystem fileSystem
+        ) {
+            while (true)
```
shocklateboy92, line 24: So, instead of doing `baseDirectoryPath = directoryInfo.Parent.FullName` below, you can probably do `directoryInfo = directoryInfo.Parent`. That way, you can have a `while (directoryInfo != null)` instead of `while (true)` which I prefer to avoid whenever possible.

Src/CSharpier/ConfigurationFileOptions.cs
```diff
@@ -17,58 +17,68 @@ public class ConfigurationFileOptions
...
+                var jsonExtensionPath = potentialPath + ".json";
+                if (fileSystem.File.Exists(jsonExtensionPath))
                 {
-                    return ReadJson(potentialPath);
+                    return ReadJson(jsonExtensionPath, fileSystem);
                 }
 
-                return ReadYaml(potentialPath);
-            }
+                var yamlExtensionPaths = new[] { potentialPath + ".yaml", potentialPath + ".yml" };
```
shocklateboy92, line 45: Then you can use that `directoryInfo` like so:
```cs
directoryInfo
  .EnumerateFiles(".csharpierrc*", SearchOption.TopDirectoryOnly)
  .Select(file => /* do pattern matching here with the suffix */)
```
to avoid multiple calls to disk.

Src/CSharpier/IgnoreFile.cs
```diff
@@ -0,0 +1,78 @@
+using System;
+using System.IO.Abstractions;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CSharpier
+{
+    public class IgnoreFile
+    {
+        protected global::Ignore.Ignore Ignore { get; }
```
shocklateboy92, line 10: Where is `global::Ignore.Ignore` defined in? Is it part of some weird library? Some ancient framework API?

shocklateboy92 (commented):
Some suggestions that (while not critical) I think would improve code quality

PR #208, opened 2021-05-17
fixing bug with verbatim strings losing empty lines

belav:


Src/CSharpier/SyntaxNodeComparer.cs
```diff
@@ -60,10 +60,10 @@ await NewSyntaxTree.GetRootAsync(cancellationToken)
             }
 
             var message =
-                "    Original: "
+                "----------------------------- Original: "
```
shocklateboy92, line 63: This might look better with a `string.Format()` than a bunch of `+=`

PR #213, opened 2021-05-18
dealing with loops without braces

### LINKED ISSUE #202: Loops without braces
I had foreach loop in my code without braces 
```cs
foreach (var c in word.Original)
    switch (c)
    {
        case '(':
        case '（':
            bracketCount++;
            break;
        case ')':
        case '）':
            bracketCount--;
            break;
    }
```

and csharpier did this to it:
```cs
foreach (var c in word.Original)switch (c)
{
    case '(':
    case '（':
        bracketCount++;
        break;
    case ')':
    case '）':
        bracketCount--;
        break;
}
```

Should we make csharpier deal with conditions/loops without braces? Or should we just have it add braces automatically?

belav:
closes #202
I did this for for, foreach, if/else if, do and while. I can't think of any other places this would be needed.

### DIFF
--- a/Src/CSharpier.Tests/TestFiles/IfStatement/IfStatements.cst
+++ b/Src/CSharpier.Tests/TestFiles/IfStatement/IfStatements.cst
@@ -20,12 +20,25 @@ public class ClassName
             return;
         }
 
-        if (true)
-            DoSoemthing();
-        else if (false)
-            DoSomething();
+        if (true) DoSomething();
+        else if (false) DoSomething();
         else
             DoSomething;
+
+        if (someLongCondition != someOtherCondition)
+            DoSomeLongMethodCall111111111111();
+        else if (someLongCondition == someOtherCondition)
+            DoSomeLongMethodCall111111111111();
+        else
+            DoSomeLongMethodCall1111111111111111111111111111111111111111();
+
+        if (
+            longStatementName
+            && longerStatementName
+            && evenLongerStatementName
+            && superLongStatementName
+        )
+            return;
 
         if (!true) { }
 
--- a/Src/CSharpier.Tests/TestFiles/IfStatement/IfStatements.cst
+++ b/Src/CSharpier.Tests/TestFiles/IfStatement/IfStatements.cst
@@ -26,11 +26,11 @@ public class ClassName
             DoSomething;
 
         if (someLongCondition != someOtherCondition)
-            DoSomeLongMethodCall111111111111();
+            DoSomeLongMethodCall____________();
         else if (someLongCondition == someOtherCondition)
-            DoSomeLongMethodCall111111111111();
+            DoSomeLongMethodCall____________();
         else
-            DoSomeLongMethodCall1111111111111111111111111111111111111111();
+            DoSomeLongMethodCall________________________________________();
 
         if (
             longStatementName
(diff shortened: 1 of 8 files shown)


Src/CSharpier.Tests/TestFiles/IfStatement/IfStatements.cst
```diff
@@ -20,13 +20,26 @@ public class ClassName
...
+        else if (false) DoSomething();
         else
             DoSomething;
 
+        if (someLongCondition != someOtherCondition)
+            DoSomeLongMethodCall111111111111();
+        else if (someLongCondition == someOtherCondition)
+            DoSomeLongMethodCall111111111111();
+        else
+            DoSomeLongMethodCall1111111111111111111111111111111111111111();
```
shocklateboy92, line 33: Not really a point, but I really like padding length by adding more `o`s to `Long` than adding `1`s to the end.
That way, when I pronounce it in my head as a really loooooooong method, I _feel_ the weight of the function call.

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/ForEachStatement.cs
```diff
@@ -20,34 +20,36 @@ public static Doc Print(ForEachStatementSyntax node)
...
-                            Node.Print(node.Type),
-                            " ",
-                            Token.Print(node.Identifier),
-                            " ",
-                            Token.Print(node.InKeyword),
-                            " ",
-                            Node.Print(node.Expression)
+                    Doc.Group(
+                        Token.PrintWithoutLeadingTrivia(node.AwaitKeyword),
+                        node.AwaitKeyword.Kind() != SyntaxKind.None ? " " : Doc.Null,
```
shocklateboy92, line 25: Have you considered if this whole thing might look better with pattern matching?

shocklateboy92 (approved):

Src/CSharpier.Tests/TestFiles/IfStatement/IfStatements.cst
belav: I was on board with it..... except rider complains about LooooooooooongMethod() being misspelled and seeing things underlined with red triggers my OCD.
How about ending with ________________ like this? It doesn't have the weight, but I think it makes it easier to read without getting distracted by all the extra 11111111111111111111s or the red squiggles

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/ForEachStatement.cs
belav: Pattern matching is still newish to me so I wasn't even thinking about using it here. My thoughts on doing it here after I changed it.
The `is not` is more typing, but I don't mind the way it looks and it kinda reads better. Although changing it caused csharpier to break in an ugly way. I'm thinking csharpier should probably break it like so instead
```
node.AwaitKeyword.Kind() is not SyntaxKind.None 
    ? " " : Doc.Null, 
// or maybe
node.AwaitKeyword.Kind() is not SyntaxKind.None 
    ? " "
    : Doc.Null, 
```

I read a discussion the other day where someone pointed out that pattern matching is really nice for something like
```
if (!(someObject is SomeType))
// vs
if (someObject is not SomeType)
```
I did use pattern matching for gross condition in another issue just now, and I'm starting to drink the pattern matching kool-aid.

shocklateboy92: When I said pattern matching, I meant the funky new match expression
```cs
  " ",
  node.AwaitKeyword.Kind() switch {                                                                                                                                                                                                            
      not SyntaxKind.None => Token.Print(node.ForEachKeyword),                                                                                                                                                                                 
      SyntaxKind.None => Token.PrintWithoutLeadingTrivia(node.ForEachKeyword)                                                                                                                                                                  
  },
  " ",
```

But at the time I thought the new instance of the ternary operator was a nested condition. Looking at the code in more detail I see that they are in fact separate arguments to the `Doc.Group()` call. Thus, I don't think there's a significantly nicer way of expressing that than your original code.

PR #227, opened 2021-05-24
Fixing bug with directive following a trailing comment getting extra indentation

### LINKED ISSUE #216: #endregion is getting indented more on each format
In some situations #endregion is getting an extra indent each time a file is formatted
runtime/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/SignatureHelper.cs
runtime/src/coreclr/System.Private.CoreLib/src/System/RuntimeType.CoreCLR.cs
runtime/src/coreclr/System.Private.CoreLib/src/System/TypeNameParser.cs

belav:
closes #216

### DIFF
--- a/Src/CSharpier.Tests/TestFiles/Directives/Directives.cst
+++ b/Src/CSharpier.Tests/TestFiles/Directives/Directives.cst
@@ -20,8 +20,8 @@ class ClassName { }
 
 public class ClassName
 {
-    #region Trailing comment does not indent more than it should #endregion
-    private int x; // trailing comment
+    #region Region
+    private int x; // trailing comment here shouldn't give extra indent to the endregion after it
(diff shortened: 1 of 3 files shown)


Src/CSharpier.Tests/TestFiles/Directives/Directives.cst
```diff
@@ -20,6 +20,10 @@ class ClassName { }
 
 public class ClassName
 {
+    #region Trailing comment does not indent more than it should #endregion
```
shocklateboy92, line 23: Does that `#endregion` end that region at the end of that line?
Or if that's meant to be be part of the comment you should reword it.

shocklateboy92 (approved):

Issue #214, opened 2021-05-18
SwitchRexpression Pattern Matching is not breaking


belav:
The 3rd part of this is way past the printWidth but doesn't break.
```c#
var member = expression switch
{
    MemberExpression { Expression: Expression target, Member: MemberInfo propertyOrField } => new Member(
        expression,
        propertyOrField,
        target
    ),
    MethodCallExpression { Method: var instanceMethod, Object: Expression target } => new Member(
        expression,
        instanceMethod,
        target
    ),
    MethodCallExpression { Method: var extensionMethod, Arguments: { Count: > 0 } arguments } when extensionMethod.Has<ExtensionAttribute>() => new Member(
        expression,
        extensionMethod,
        arguments[0]
    ),
    _ => default
};
```

belav:
I believe the problem is more with the IsPattern, see
```c#
            if (
                !(node
                    is PrefixUnaryExpressionSyntax { Operand: ParenthesizedExpressionSyntax { Expression: IsPatternExpressionSyntax { Pattern: DeclarationPatternSyntax, } isPattern, }, } notExpression)
            ) {
                return;
            }
```

belav:
This will hopefully be resolved in #154

belav:
The original issue was resolved with #382

PR #229, opened 2021-05-24
formatting with 0.9.4

belav:


Src/CSharpier/CSharpier.csproj
```diff
@@ -1,5 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
-	<Import Project="../../CSharpier.Build.props"/>
+	<Import Project="../../CSharpier.Build.props" />
```
shocklateboy92, line 2: Is this the right level of indent?

Src/CSharpier/SyntaxPrinter/Token.cs
```diff
@@ -62,7 +62,7 @@ public static Doc PrintWithSuffix(SyntaxToken syntaxToken, Doc suffixDoc)
                 && syntaxToken.Text.StartsWith("@"))
                 || (syntaxToken.Kind() == SyntaxKind.InterpolatedStringTextToken
                 && syntaxToken.Parent!.Parent
-                    is InterpolatedStringExpressionSyntax { StringStartToken:  { RawKind: (int)SyntaxKind.InterpolatedVerbatimStringStartToken }  } )
+                    is InterpolatedStringExpressionSyntax { StringStartToken: { RawKind: (int)SyntaxKind.InterpolatedVerbatimStringStartToken } })
```
shocklateboy92, line 65: Is there a bug tracking breaking these?

Src/CSharpier.Tests/SyntaxNodeComparerTests.cs
```diff
@@ -156,12 +158,14 @@ public DropdownAttribute(bool ignoreIfNotPresent)
         [Test]
         public void MissingSemiColon()
         {
-            var left = @"public enum Enum
+            var left =
+                @"public enum Enum
```
shocklateboy92, line 162: Okay, the more I think about it, the more I realize: _there is no good way_ to format multiline strings! :stuck_out_tongue:

Src/CSharpier/SyntaxPrinter/Token.cs
belav: I believe this is the same problem https://github.com/belav/csharpier/issues/214

Src/CSharpier.Tests/SyntaxNodeComparerTests.cs
belav: The recent PR I might actually make it easier to keep the `@"public enum Enum` on the same line, although it may depend on implementing ConditionalGroup

PR #230, opened 2021-05-24
adding ability to display differences when --check fails

### LINKED ISSUE #182: File that fails check should give some indication of the formatting that was missing.
When the --check option is used, the output only contains if a file was not formatted. It doesn't contain any indication of what part of the file wasn't formatted.
Theoretically you can just run csharpier on the file, but in the case where the --check appears to be a false positive, not having a message means it is harder to track down why it failed.

belav:
closes #182

### DIFF
--- a/Src/CSharpier/Utilities/StringDiffer.cs
+++ b/Src/CSharpier/Utilities/StringDiffer.cs
@@ -4,7 +4,11 @@ namespace CSharpier.Utilities
 {
     public static class StringDiffer
     {
-        public static void PrintDifference(string expected, string actual, IConsole console)
+        /// <summary>
+        /// Given two strings that are different, this will print the first different line it encounters
+        /// along with a single line before and after that different line.
+        /// </summary>
+        public static void PrintFirstDifference(string expected, string actual, IConsole console)
         {
             if (expected == actual)
             {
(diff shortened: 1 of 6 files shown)


Src/CSharpier/Utilities/StringDiffer.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            console.WriteLine(
+                "The file contained different line endings than formatting it would result in."
+            );
+        }
+
+        private static string? MakeWhiteSpaceVisible(string? value)
+        {
+            return value?.Replace(' ', (char)183).Replace('\t', '→');
```
shocklateboy92, line 72: Any particular reason you didn't put character 183 as a literal here like you did with `'→'`?

Src/CSharpier/Utilities/StringDiffer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+                console.WriteLine(
+                    $"----------------------------- Expected: Around Line {line} -----------------------------"
+                );
+                if (previousExpectedLine != null)
+                {
+                    console.WriteLine(MakeWhiteSpaceVisible(previousExpectedLine));
+                }
+                console.WriteLine(MakeWhiteSpaceVisible(expectedLine));
+                var nextExpectedLine = expectedReader.ReadLine();
```
shocklateboy92, line 42: Maybe leave a comment somewhere in/around this function that if there are multiple lines that are different, it only prints around the first divergence. That would have made this code much quicker to parse.

shocklateboy92 (approved):

PR #231, opened 2021-05-24
Pass parents around in syntaxNodeComparer so we can get better errors when things are missing

### LINKED ISSUE #160: Missing nodes in SyntaxNodeComparer need better reporting.
When SyntaxNodeComparer compares the following
```
public class ConstructorWithBase
{
    public ConstructorWithBase(string value)
        : base(value)
    {
        return;
    }
}
```
```
public class ConstructorWithBase
{
    public ConstructorWithBase(string value)
    {
        return;
    }
}
```

It ends up with the following message. Which isn't as helpful as it could be.
```
Original: Around Line 3
{
    public ConstructorWithBase(string value)
        : base(value)
    {
        return;
    }
}
    Formatted: Missing
```

The problem is because the root Compare method is comparing null to a node, which means they are not equal. But it has no context for where the null node is.
SyntaxNodeComparer probably needs to be modified to keep track of parents in the stack, and passing them to the compare methods.
See the ignored unit test.

belav:
closes #160


shocklateboy92 (approved):
LGTM

Issue #239, opened 2021-05-30
Better generic ObjectCreationExpressionSyntax


belav:
The current formatting is
```c#
    public static Dictionary<Type, string> PropertiesByType = new Dictionary<
        Type,
        string
    >();
```
I think it would be possible to format like this, but this may require conditionalGroups. See https://github.com/belav/csharpier/issues/7
```c#
    public static Dictionary<Type, string> PropertiesByType =
        new Dictionary<Type, string();
```

shocklateboy92:
I don't get why conditional groups would be needed. I thought the logic in doc for nested groups is to try breaking the outer groups before inner ones.

So if we change the printer to something loosely resembling:
```cs
Doc.Group(
    Node.Print(/* the stuff on the left */),
    "=",
    Node.Print(/* the stuff on the right */)
)
```
Would we not get the desired effect?
Same goes for #250

belav:
This appears to have been resolved at some point

Issue #7, opened 2021-02-26
InvocationExpressions need some work


belav:
Some examples

```c#
                var extension = GetExtensionForLanguageType(
                    (FrontEndResourceLanguage)Enum.Parse(
                        typeof(FrontEndResourceLanguage),
                        this.Language));

            this.Address1 = addressFields.FirstOrDefault(
                field => field.FieldName == "Address1");

// Invocation inside IfStatement
            if (customerOrder == null || !customerOrder.FulfillmentMethod.EqualsIgnoreCase(
                FulfillmentMethod.PickUp.ToString()))

```

belav:
Printing InvocationExpressions well for all the variations looks like it will be a lot of work. See https://github.com/prettier/prettier/blob/main/src/language-js/print/member-chain.js for how complex it could get.

belav:
I made some progress on formatting simpler cases than the ones above.

Here is another one that needs some work
```c#
        this.Address1 = addressFields.Where(
                field => field.FieldName == "Address1")
            .Where(field => yeah)
            .Where(field => yeah);
```

belav:
Another example
```c#
    roleNames
        .ToList()
        .ForEach((role) =>
            this.adminContextMock
                .Setup((ctx) => ctx.IsUserInRole(role))
                .Returns(false)
        );
```

belav:
This will require conditionalGroup, run it in prettier to see how they do it.
```c#
        var superLongMethodNameForceLine = someFactoryName
            .SuperLongMethodNameForceLine()
            .SomeOtherReallyLongMethodName();
```

belav:
Another edge case
```c#
    var linePosition________________________ = diagnostics[
            i
        ].Location.GetLineSpan().StartLinePosition;
```

belav:
this seems like extra indentation
```
            var compilation = await project.WithCompilationOptions(compilationOptions)
                    .GetCompilationAsync();
```

belav:
Possibly already covered above with the `diagnostics[` edge case
```
                AttributeCollection attributes = TypeDescriptor.GetProperties(modelType)[
                        propertyName
                    ].Attributes;

                JsonIgnoreAttribute jsonIgnoreAttribute = (JsonIgnoreAttribute)attributes[
                        typeof(JsonIgnoreAttribute)
                    ];

**Comment by belav** [comment]
And
```
            string signedRequest = htmlHelper.ViewContext.HttpContext.Request.Params[
                "signed_request"
            ];

[...]

Issue #238, opened 2021-05-29
ParenthesizedLambdaExpressions needs some work.


belav:
```c#
        var task = Task.Factory.StartNew(async () =>  {
                return await new WebClient().DownloadStringTaskAsync(
                    "http://example.com"
                );
            });
// should probably be
        var task = Task.Factory.StartNew(async () =>
        {
            return await new WebClient().DownloadStringTaskAsync(
                "http://example.com"
            );
        });
```

belav:
This got mostly resolved, so the only part left is to clean up the indentation.

shocklateboy92:
You think something like this might look better? To make it clear that the braces aren't for the outer level statement, but rather the lambda being passed in?
```cs
        var task = Task.Factory.StartNew(async () =>
            {
                return await new WebClient().DownloadStringTaskAsync(
                    "http://example.com"
                );
            }
        );
```

belav:
It actually formats really close to that now
```c#
        var task = Task.Factory.StartNew(
            async () =>
            {
                return await new WebClient().DownloadStringTaskAsync(
                    "http://example.comddddddddddddddddddd"
                );
            }
        );
```
Prettier formats it like this
```c#
var task = Task.Factory.StartNew(async () => {
    return await new WebClient().DownloadStringTaskAsync(
        "http://example.comddddddddddddddddddd"
    );
});
```

But I don't think we want to have the brace on the same line as `=>`

We could maybe move the `)` up, which would be this, but I don't think I love that.
```c#
        var task = Task.Factory.StartNew(async () =>
            {
                return await new WebClient().DownloadStringTaskAsync(
                    "http://example.com"
                );
            });
```
Which leaves us with doing what you said
```c#
        var task = Task.Factory.StartNew(async () =>
            {
                return await new WebClient().DownloadStringTaskAsync(
                    "http://example.com"
                );
            }
        );

// however if there are multiple parameters

        var task = Task.Factory.StartNew(
            async () => 
            {
                return await new WebClient().DownloadStringTaskAsync(
                    "http://example.com"
                );
            },
            someOtherParameter
        );

PR #240, opened 2021-05-30
Improving formatting of generics + variable declarations.

belav:
This started as just cleaning up the formatting of a generic dictionary in a field but attempting to get that cleaned up broke a number of other test cases without adding all this logic into VariableDeclaration.
This will allow us to handle edge cases better. The previous simple logic handled basic cases well, but didn't handle edge cases cleanly.
I split a couple of things off into separate tickets.
https://github.com/belav/csharpier/issues/239 could be better, but this is how prettier currently formats things.
https://github.com/belav/csharpier/issues/7 I added another edge case to this which will probably require conditional groups.


Src/CSharpier.Tests/TestFiles/FieldDeclaration/FieldDeclarations.cst
```diff
@@ -30,4 +30,29 @@ public class ClassName
...
     };
+
+    // leading comments
+    private static const string SomeName = "1";
+
+    public SomeClass SomeName; // trailing comments
+
+    public System.Text.StringBuilder NamespacedField;
+
+    public static Dictionary<Type, string> PropertiesByType = new Dictionary<
```
shocklateboy92, line 41: I think it makes sense to break on the assignment first, i.e.
```
public static Dictionary<Type, string> PropertiesByType =
    new Dictionary<Type, string>();
```

In fact, I'm a little confused why this isn't happening already. The assignment expression is the outermost `Group()` right? It should break on that before the its children (which presumably are the left and right sides)

Src/CSharpier.Tests/TestFiles/InitializerExpression/InitializerExpressions.cst
```diff
@@ -39,4 +39,13 @@ class ClassName
             longValue______________________________
         }
     };
+
+    void MethodName()
+    {
+        var x = new SomeClass
+        {
+            Property1 = true
+            // should indent
```
shocklateboy92, line 48: Like... who would put a comment _there_?! :banana:

Src/CSharpier.Tests/TestFiles/ParenthesizedLambdaExpression/ParenthesizedLambdaExpressions.cst
```diff
@@ -13,13 +13,10 @@ public class ClassName
             ) => someLongNameThatWillDoStuff
         );
 
-        var task = Task.Factory.StartNew(
-            async () =>
-            {
+        var task = Task.Factory.StartNew(async () => {
```
shocklateboy92, line 16: Oaf. Are you sure about this? It's the first time we've opened a curly brace before the line break...

Src/CSharpier.Tests/TestFiles/PropertyDeclaration/PropertyDeclarations.cst
```diff
@@ -54,7 +54,9 @@ public class ClassName
         set;
     }
 
-    public virtual ICollection<SomeObject> SomeLongNameThatForcesALineBreak { get; set; } =
+    public virtual ICollection<
+        SomeObject
```
shocklateboy92, line 58: We're breaking too eagerly here. `public virtual ICollection<SomeObject>` fits in one line.

Hypothetically, even if ti didn't, I think the type arguments should be the last group we break.
```cs
public virtual
    ICollection<SomeLooooooooooooooooooooooooooongType>
        SomeLongNameThatForcesALineBreak { get; set; } =
            HashSet<SomeLooooooooooooooooooooooooooongType>
```

Src/CSharpier.Tests/TestFiles/QueryExpression/QueryExpressions.cst
```diff
@@ -4,6 +4,13 @@ class ClassName
     {
         var query1 = from c in customers select c;
         query1 = from c in customers select c into d select d;
+        query1 =
+            from c in customers
+            join c1 in customers11111111111111111111111111111
```
shocklateboy92, line 9: Looks surprisingly nice :+1:

shocklateboy92 (commented):
Woah, this is one giant PR. I'm too sleepy to go through the changes in `SyntaxNodePrinters` atm, but I assume they just caused the formatting changes in the tests, which I did make comments on.

Src/CSharpier.Tests/TestFiles/FieldDeclaration/FieldDeclarations.cst
belav: There were a few edge cases that popped up due to the changes I was making. I created issues for some of them because this PR was already getting large.
Getting this to break after the = caused a lot of other cases to change. I think this one requires conditional groups (which don't exist yet), or possibly an ifBreak. 
https://github.com/belav/csharpier/issues/239

Src/CSharpier.Tests/TestFiles/InitializerExpression/InitializerExpressions.cst
belav: Exactly. But I think I ran into this case in one of the repos I test format.

Src/CSharpier.Tests/TestFiles/ParenthesizedLambdaExpression/ParenthesizedLambdaExpressions.cst
belav: I opened this to address it https://github.com/belav/csharpier/issues/238
But it turned out to be a two minute fix not counting the indentation.

Src/CSharpier.Tests/TestFiles/PropertyDeclaration/PropertyDeclarations.cst
belav: This is currently similar to how prettier breaks the following typescript
```
  public SomeName: SomeLongGenericClassThatBreaks<
    SomeRandomType,
    SomeOtherRandomType
  > = new SomeLongGenericClassThatBreaks(someValue, someOtherValue);
```
Although I'm not sure how much I like how that typescript breaks. I created https://github.com/belav/csharpier/issues/243

belav:
These changes caused all kinds of formatting to get real ugly when I was reviewing how the 0.9.5 RC would format the forked public repos I had. I pulled it out of master for now, and will probably spend the next week or two just cleaning up issues related to it.

Issue #241, opened 2021-05-30
MSBuild - can dotnet watch pass list of changed files to csharpier?


belav:
Right now when using the CSharpier.MSBuild package, when dotnet watch is used it reformats the entire project instead of just the files that were changed.

This can be used in CSharpier.MSBuild.targets to pass the list of files to csharpier, but @(Compile) includes all files even in dotnet watch run.
```xml
<PropertyGroup Condition="'$(DotNetWatchBuild)'=='true'">
    <!-- From the item list Compile, use the FullPath and separate them by a ' ' instead of the default ;-->
    <CSharpierArgs>$(CSharpierArgs) --check @(Compile->'%(FullPath)', ' ')</CSharpierArgs>
</PropertyGroup>
```

The readme for `dotnet watch`
https://github.com/dotnet/sdk/tree/main/src/BuiltInTools/dotnet-watch
The file where you can see DotNetWatchBuild being set
https://github.com/dotnet/sdk/blob/main/src/BuiltInTools/dotnet-watch/Internal/MsBuildFileSetFactory.cs

belav:
Requested this in dotnet watch - https://github.com/dotnet/sdk/issues/17980

PR #242, opened 2021-05-30
Only write files if they changed

### LINKED ISSUE #228: CSharpier.MSBuild does not work with dotnet watch run
There may be a couple problems here.
When I use CSharpier.MSBuild on one small project with dotnet watch run, it appears to get into an endless loop of formatting/triggering dotnet watch.
image
I believe this is because CSharpier writes out a file even if nothing was changed by formatting it.

When using CSharpier.MSBuild on CShapier there is a slightly different outcome which may be caused by the same problem.
image
The build errors in this come from that fact that somehow files are getting there contents changed to a single new line.

belav:
closes #228
dotnet watch run gets stuck in a loop because it detects a file write even if the contents of the file did not change with that write.
#241 was created so we can hopefully modify CSharpier.MSBuild in the future to only format files that were changed.


shocklateboy92 (approved):
LGTM

PR #264, opened 2021-06-05
refactoring + optimizing

belav:
Cleaned up some todos.
Refactored some small stuff.
Optimized a hot path in SyntaxNodeComparer
Optimized a hot path in IndentBuilder and removed dedent, because we don't support it yet.
Got rid of extra List<Doc> creations that weren't required.
Use single instance of some Docs that have no parameters.


belav:
This closes #277

PR #265, opened 2021-06-07
Better format interfaces and base classes

### LINKED ISSUE #189: Formatting of implemented classes/interfaces
```csharp
//Is
public class AClass
    : AReallyLongBuilder<BuilderType>,
        IAnotherReallyBuilder<AnotherType> { }

//ShouldBe
public class AClass
    : AReallyLongBuilder<BuilderType>, IAnotherReallyBuilder<AnotherType> { }

OR 

public class AClass
    : AReallyLongBuilder<BuilderType>, 
      IAnotherReallyBuilder<AnotherType> { }
```

divyanshshukla13:
Format the inherited interfaces and base classes according to https://github.com/belav/csharpier/issues/189#issuecomment-847140886. 

The new style is like this
```csharp
public class AClass : 
    AReallyLongBuilder<BuilderType>, 
    IAnotherReallyBuilder<AnotherType> { }
```

Closes #189

### DIFF
--- a/Src/CSharpier.Tests/TestFiles/ClassDeclaration/ClassDeclarations.cst
+++ b/Src/CSharpier.Tests/TestFiles/ClassDeclaration/ClassDeclarations.cst
@@ -6,13 +6,13 @@ class NoModifiers { }
 
 public class WithInterface : IInterface { }
 
-public class WithReallyLongNameInterface
-    : IReallyLongNameLetsMakeThisBreak___________________________ { }
+public class WithReallyLongNameInterface :
+    IReallyLongNameLetsMakeThisBreak___________________________ { }
 
-public class ThisIsSomeLongNameAndItShouldFormatWell1
-    : AnotherLongClassName,
-        AndYetAnotherLongClassName,
-        AndStillOneMore { }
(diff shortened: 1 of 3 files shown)


belav:
I updated the unit tests with this current formatting. It looks good for most cases except when a class has a base list + generic constraints, there is no real separation between them. It still improves most cases, so I think we could merge it as is and come back to address this case below later.

```c#
public class ThisIsSomeLongNameAndItShouldFormatWell2<T, T2, T3> :
    AnotherLongClassName<T>,
    AnotherClassName
    where T : new(), AnotherTypeConstraint
    where T2 : new()
    where T3 : new() { }
// I think it would look better like this, but I believe this would mean implementing the Align doc type
public class ThisIsSomeLongNameAndItShouldFormatWell2<T, T2, T3>
    : AnotherLongClassName<T>,
      AnotherClassName
    where T : new(), AnotherTypeConstraint
    where T2 : new()
    where T3 : new() { }
```

I also ran this against [aspnetcore](https://github.com/belav/aspnetcore/commit/5a73c0d2ae43fc9ac997c7f4049cdf3cdcc081c4) and found a couple other edge cases we will want to handle eventually. 
https://github.com/belav/csharpier/issues/266

PR #268, opened 2021-06-07
Fix indentation of lambdas during declaration

### LINKED ISSUE #236: Field with lambda
I'm not sure how this should format, but I think this could be improved.
```c#
    private static readonly Func<
        SyntaxTrivia,
        bool
    > s_isVisualBasicCommentTrivia =
        (syntaxTrivia) => syntaxTrivia.IsKind(
            VisualBasic.SyntaxKind.CommentTrivia
        );
```

### LINKED ISSUE #246: Extra Indentation with Lambda
0.9.5 introduced this probably as part of https://github.com/belav/csharpier/pull/240
```c#
Action find = () => EntryPointDiscoverer.FindStaticEntryMethod(
        typeof(IEnumerable<>).Assembly
    );
// should be
Action find = () => EntryPointDiscoverer.FindStaticEntryMethod(
    typeof(IEnumerable<>).Assembly
);
```
#238 may be related

divyanshshukla13:
Closes #236 
Closes #246

### DIFF
--- a/Src/CSharpier.Tests/TestFiles/FieldDeclaration/FieldDeclarations.cst
+++ b/Src/CSharpier.Tests/TestFiles/FieldDeclaration/FieldDeclarations.cst
@@ -51,6 +51,10 @@ public class ClassName
         SomeLongTypeName,
         SomeOtherLongTypeName_______________________________
     >();
+
+    private static readonly Func<SyntaxTrivia, bool> s_isVisualBasicCommentTrivia = (
+        syntaxTrivia
+    ) => syntaxTrivia.IsKind(VisualBasic.SyntaxKind.CommentTrivia);
 }
 
 struct S
--- a/Src/CSharpier.Tests/TestFiles/VariableDeclaration/VariableDeclarations.cst
+++ b/Src/CSharpier.Tests/TestFiles/VariableDeclaration/VariableDeclarations.cst
@@ -21,5 +21,9 @@ class ClassName
 
         var trailingMulti = 1; /* ensure no extra line after this */
         var trailingMultiAfter = 2;
+
+        Func<SyntaxTrivia, bool> s_isVisualBasicCommentTrivia = (
+            syntaxTrivia
+        ) => syntaxTrivia.IsKind(VisualBasic.SyntaxKind.CommentTrivia);
     }
 }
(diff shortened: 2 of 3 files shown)


shocklateboy92 (changes requested):
Could you please update add a test that covers this change?
It should be fairly straight-forward - you just need to create a `.cst` file in `Src/CSharpier.Tests/TestFiles/VariableDeclaration/` and reference it in `_VariableDeclarationTests.cs`.

belav:
> It should be fairly straight-forward - you just need to create a `.cst` file in `Src/CSharpier.Tests/TestFiles/VariableDeclaration/` and reference it in `_VariableDeclarationTests.cs`.
The standard for tests is a single file within each directory, unless there are special cases that warrant being in their own file. And this bug shows up with a FieldDeclaration, which would mean updating `/TestFiles/FieldDeclaration/FieldDeclarations.cst`, although it may also present itself when declaring a variable as well.

Src/CSharpier.Tests/TestFiles/FieldDeclaration/FieldDeclarations.cst
```diff
@@ -51,6 +51,10 @@ public class ClassName
         SomeLongTypeName,
         SomeOtherLongTypeName_______________________________
     >();
+
+    private static readonly Func<SyntaxTrivia, bool> s_isVisualBasicCommentTrivia = (
```
shocklateboy92, line 55: Do people really use Hungarian notation (e.g. `s_`) in C#? :open_mouth:

shocklateboy92 (approved):
Thanks for your contribution to CSharpier!
Feel free to drop by on our [Discord channel](https://discord.gg/HfAKGEZQcX) sometime

Src/CSharpier.Tests/TestFiles/FieldDeclaration/FieldDeclarations.cst
divyanshshukla13: Not really. I just picked up this example from the linked issue.

belav: I've definitely run into it, which is why it ended up in the linked issue.

PR #284, opened 2021-06-11
Better print generic methods and constructors

### LINKED ISSUE #94: Generic method is a little ugly
The where on this seems out of place. This could have to do with SpaceBrace, maybe if we just NewLineBrace where there is a where constraint it will be good.
or the where should stay on the same line as the CloseParen, but the brace should still newline
```c#
        public static IQueryable<TEntity> ApplyTracking<TEntity>(
            this IQueryable<TEntity> query,
            bool tracking
        )
            where TEntity : class {
            return tracking ? query : query.AsNoTracking();
        }
```

divyanshshukla13:
If there is only one generic constraint, we try to keep it on the same line. If more one, they're broken up.
As for the constructor initialiser, it's kept on the same line as the ending brace of the arguments.

Closes #94

### DIFF
--- a/Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
+++ b/Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
@@ -36,9 +36,8 @@ public class Initializers : BasicClass
         string reallyLooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongParameter
     ) : base(false) { }
 
-    public ReallyLoooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooogMethodName() : base(
-        false
-    ) { }
+    public ReallyLoooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooogMethodName()
+        : base(false) { }
 }
 
 public class Exactly80
--- a/Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
+++ b/Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
@@ -43,6 +43,14 @@ public class Initializers : BasicClass
         : base(
             reallyLooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongParameterIMeanIt
         ) { }
+
+    public Initializers(
+        string value
+    ) /*Comment*/
+      : base(value) { }
+
+    public ReallyLoooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooogMethodName()
+        /*Comment*/: base(false) { }
 }
 
 public class Exactly80
--- a/Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
+++ b/Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
@@ -22,11 +22,20 @@ public class BasicClass
 
 public class Initializers : BasicClass
 {
-    public Initializers()
-        : this(true) { }
+    public Initializers() : this(true) { }
 
-    public Initializers(string value)
-        : base(value) { }
+    public Initializers(string value) : base(value) { }
+
+    public Initializers(
+        string parameter1,
+        string parameter2,
+        string parameter3,
+        string parameter4
+    ) : base(false)
+    {
+        _parameter1 = parameter1;
+        _parameter2 = parameter2;
+    }
 }
 
 public class Exactly80
--- a/Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
+++ b/Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
@@ -38,6 +38,11 @@ public class Initializers : BasicClass
 
     public ReallyLoooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooogMethodName()
         : base(false) { }
+
+    public ReallyLoooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooogMethodName()
+        : base(
+            reallyLooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongParameterIMeanIt
+        ) { }
 }
 
 public class Exactly80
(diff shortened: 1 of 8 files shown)


Src/CSharpier/SyntaxPrinter/ConstraintClauses.cs
```diff
@@ -16,19 +16,40 @@ public static Doc Print(IEnumerable<TypeParameterConstraintClauseSyntax> constra
             {
                 return Doc.Null;
             }
+            else if (constraintClausesList.Count == 1)
```
shocklateboy92, line 19: So, this whole function can be replaced by C# 8/9's new pattern matching syntax :rocket: 
```
return constraintClauses.ToList() switch
{
    { Count: 0 } => Doc.Null,
    { Count: 1 } => /* stuff */,
    var list when list[0].Parent is MethodDeclarationSyntax => /* stuff */,
    var list => Doc.Indent(
        Doc.HardLine,
        Doc.Join(Doc.HardLine, list.Select(TypeParameterConstraintClause.Print))
    )                                               
};
```

shocklateboy92 (approved):
The implementation and tests look good.
Feel free to update to the new pattern matching syntax if you think it's cool :smile:

Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
```diff
@@ -22,11 +22,23 @@ public class BasicClass
...
+        reallyLooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongParameterIMeanIt
+    ) { }
+
+    public Initializers(
+        string reallyLooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongParameter
+    ) : base(false) { }
+
+    public ReallyLoooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooogMethodName() : base(
+        false
+    ) { }
```
belav, line 41: This one should probably change. Here's a real world example.
```
// now formats as
        public DiagnosticLocation(int line, int column) : this(
            $"{DiagnosticProject.DefaultFilePathPrefix}.cs",
            line,
            column
        ) { }
// used to format as this, which I think is preferable.
        public DiagnosticLocation(int line, int column)
            : this($"{DiagnosticProject.DefaultFilePathPrefix}.cs", line, column) { }
```

belav: When they are both long, I think it looks good
```
        public ServerInteropTest(
            BrowserFixture browserFixture,
            ToggleExecutionModeServerFixture<Program> serverFixture,
            ITestOutputHelper output
        ) : base(
            browserFixture,
            serverFixture.WithServerExecution().WithAdditionalArguments(GetAdditionalArguments()),
            output
        ) { }
```
Another example with a long base call which wouldn't fit on a single line. It isn't obvious all the parameters are going to a base call.
```
            public TestRemoteRenderer(IServiceProvider serviceProvider, IClientProxy client) : base(
                serviceProvider,
                NullLoggerFactory.Instance,
                new CircuitOptions(),
                new CircuitClientProxy(client, "connection"),
                NullLogger.Instance,
                null
            ) { }
// I'm thinking it should break before : base to make it more clear
            public TestRemoteRenderer(IServiceProvider serviceProvider, IClientProxy client)
                : base(
                    serviceProvider,
                    NullLoggerFactory.Instance,
                    new CircuitOptions(),
                    new CircuitClientProxy(client, "connection"),
                    NullLogger.Instance,
                    null
                ) { }
```

belav (changes requested):
I think the short constructor long base/this call should get handled, but in general this all looks good.

There are a couple other edge cases I ran into that can probably just go into new issues.

There is no indent on the comment but used to be. Probably not super important.
```
        public UsesRendererTypesInDeclarations()
        /*MMInvocation*/: base(null, null) { }
// this is probably the same issue
        public ServeCommand(CommandLineApplication parent)
        // We pass arbitrary arguments through to the ASP.NET Core configuration
        : base(throwOnUnexpectedArg: false)
```

This looks a little weird, but also looked weird before your changes and I'm not really sure how we'd want to format it.
```
// currently
        private T InstantiateComponent<T>() where T : IComponent =>
            (T)_renderer.InstantiateComponent<T>();
// used to be
        private T InstantiateComponent<T>()
            where T : IComponent => (T)_renderer.InstantiateComponent<T>();
```
A bigger version of the same thing
```
// currently
        private static IList<T> GetComponents<T>(RenderTreeBuilder builder) where T : IComponent =>
            builder.GetFrames()
                .AsEnumerable()
                .Where(x => x.FrameType == RenderTreeFrameType.Component)
                .Select(x => (T)x.Component)
                .ToList();
// used to be
        private static IList<T> GetComponents<T>(RenderTreeBuilder builder)
            where T : IComponent =>
            builder.GetFrames()
                .AsEnumerable()
                .Where(x => x.FrameType == RenderTreeFrameType.Component)
                .Select(x => (T)x.Component)
                .ToList();

belav:
> This looks a little weird, but also looked weird before your changes and I'm not really sure how we'd want to format it.
> 
> ```
> // currently
>         private T InstantiateComponent<T>() where T : IComponent =>
>             (T)_renderer.InstantiateComponent<T>();
> // used to be
>         private T InstantiateComponent<T>()
>             where T : IComponent => (T)_renderer.InstantiateComponent<T>();
> ```
> 
> A bigger version of the same thing
> 
> ```
> // currently
>         private static IList<T> GetComponents<T>(RenderTreeBuilder builder) where T : IComponent =>
>             builder.GetFrames()
>                 .AsEnumerable()
>                 .Where(x => x.FrameType == RenderTreeFrameType.Component)
>                 .Select(x => (T)x.Component)
>                 .ToList();
> // used to be
>         private static IList<T> GetComponents<T>(RenderTreeBuilder builder)
>             where T : IComponent =>
>             builder.GetFrames()
>                 .AsEnumerable()
>                 .Where(x => x.FrameType == RenderTreeFrameType.Component)
>                 .Select(x => (T)x.Component)
>                 .ToList();
> ```

After looking at this again, maybe it doesn't need to change. I'm not sure what else we'd do with it, and I think the new way is better than the old. It might just look odd because I'm not used to seeing generic expression bodied methods.

Src/CSharpier.Tests/TestFiles/ConstructorDeclaration/ConstructorDeclarations.cst
divyanshshukla13: Done!

divyanshshukla13:
> After looking at this again, maybe it doesn't need to change

Yeah, I wouldn't be so sure of either one. I've seen both styles in `csharpier-repos`. I like the one in this PR just because it preserves a line.

belav (approved):

PR #287, opened 2021-06-12
Support writing to stdout and taking input from stdin

### LINKED ISSUE #282: Add an option to write the formatted file to stdout
It would be great to be able to create a vscode extension for this formatter that is compatible with vscode's format-on-save.

It would be best if it were possible to leave the actual editing of the file contents to vscode, and to just print out the new file contents to stdout.

This doesn't need to be for multiple files in this use case.

This is how prettier works (except when using `--write`), so it makes sense to work in the same way.

Additionally it would help to support accepting a file from stdin rather than a path, so that it  can be made to work with virtual filesystems.

belav:
closes #282


Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -99,20 +93,52 @@ CancellationToken cancellationToken
...
+
+                    await commandLineFormatter.FormatFiles(cancellationToken);
+                }
             }
 
             result.ElapsedMilliseconds = stopwatch.ElapsedMilliseconds;
-            ResultPrinter.PrintResults(result, console, commandLineOptions);
+            if (!commandLineOptions.ShouldWriteStandardOut)
+            {
+                ResultPrinter.PrintResults(result, console, commandLineOptions);
```
shocklateboy92, line 132: Might still be worth writing this out to StdErr.

Src/CSharpier/Program.cs
```diff
@@ -28,14 +30,36 @@ static async Task<int> Main(string[] args)
             bool check,
             bool fast,
             bool skipWrite,
+            bool writeStdout,
             CancellationToken cancellationToken
         ) {
+            string? standardInFileContents = null;
+            if (Console.IsInputRedirected)
+            {
+                var input = new StringBuilder();
```
shocklateboy92, line 39: What's with all this getup? Wouldn't it be easier to use `StreamReader.ReadToEnd()`?

belav: Yeah that works too. I blame stumbling across a bad answer on stackoverflow.

Src/CSharpier/CommandLineFormatter.cs
belav: If I'm reading up on this correctly, I think that would mean all of the results/error information could be written to stderr and that wouldn't interfere with the formatted file that is written to stdout.
Which would definitely be nice because when I was working on the change below CSharpier wasn't giving me any information and not formatting the text I was trying to pipe to it.

PR #306, opened 2021-06-14
Always break + indent with optional braces

### LINKED ISSUE #245: Format else with no braces consistently with if else and if.
```c#
        if (true) DoSomething();
        else if (false) DoSomething();
        else
            DoSomething;
// should be
        if (true) DoSomething();
        else if (false) DoSomething();
        else DoSomething;

### LINKED ISSUE #255: Nested statements without braces
Currently statements that have optional braces will format onto a single line if they fit
```c#
if (someCondition) return 12;
```
If statements are nested, and still fit onto a single line they still don't break.
```c#
for (var i = 0; i < count; i++) if (comparer.Equals(this[i], item)) return i;
// and also
if (behaviors != null) foreach (var behavior in behaviors) AddBehavior(behavior);
```
Should statements without braces always break + indent?
Should they break + indent if they contain another statement that could use braces but doesn't?

### LINKED ISSUE #303: Always break statements without braces.
Statements that have optional braces should always break. I believe they are 

- if
- else
- while
- foreach
- for

```
if (behaviors != null) foreach (var behavior in behaviors) AddBehavior(behavior);

// should be
if (behaviors != null)
    foreach (var behavior in behaviors)
        AddBehavior(behavior);
```

belav:
We haven't heard any compelling arguments for not always breaking. And from looking at public repos it appears that always breaking is more common.
closes #303
closes #245
closes #255

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/OptionalBraces.cs
+++ b/Src/CSharpier/SyntaxPrinter/OptionalBraces.cs
@@ -1,0 +1,22 @@
+using System.Collections.Generic;
+using System.Linq;
+using CSharpier.DocTypes;
+using CSharpier.SyntaxPrinter.SyntaxNodePrinters;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace CSharpier.SyntaxPrinter
+{
+    public static class OptionalBraces
+    {
+        public static Doc Print(StatementSyntax node, string? groupId = null)
+        {
+            return node
+                is BlockSyntax blockSyntax
+                ? groupId == null
+                        ? Block.Print(blockSyntax)
+                        : Block.PrintWithConditionalSpace(blockSyntax, groupId)
+                : Doc.Indent(Doc.HardLine, Node.Print(node));
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


Src/CSharpier/SyntaxPrinter/OptionalBraces.cs
```diff
@@ -0,0 +1,22 @@
...
+using CSharpier.DocTypes;
+using CSharpier.SyntaxPrinter.SyntaxNodePrinters;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace CSharpier.SyntaxPrinter
+{
+    public static class OptionalBraces
+    {
+        public static Doc Print(StatementSyntax node, string? groupId = null)
```
shocklateboy92, line 12: I may be drinking the pattern matching cool-aid too much, but do you think this is more readable?
```cs
public static Doc Print(StatementSyntax node, string? groupId = null) =>
    (node, groupId) switch
    {
        (BlockSyntax blockSyntax, null) => Block.Print(blockSyntax),
        (BlockSyntax blockSyntax, var id) => Block.PrintWithConditionalSpace(
            blockSyntax,
            id
        ),
        (var otherSyntax, _) => Doc.Indent(Doc.HardLine, Node.Print(otherSyntax))
    };
```

Though, I do admit my ability to follow nested ternary operator blocks is somewhat subpar.

shocklateboy92 (approved):

Src/CSharpier/SyntaxPrinter/OptionalBraces.cs
belav: That is less readable to me, but it is also new to me.
I was also not super happy with the nested ternary, but it seemed "good enough"

PR #310, opened 2021-06-14
Format Implicit Array Initializer consistently

### LINKED ISSUE #297: Implicit Array Initializer does not format braces consistently with other statements.
```
        var options = new[] {
            new Option("-s", "an option"),
            new Option("--very-long", "an option")
        };

        var something = new SomeObject
        {
            Value = "kljasdlfkjaslkdfasdf",
            Value2 = "lkasdlfkjaslkjdfaslkdjf"
        };
```
The array initializer should be
```
        var options = new[]
        {
            new Option("-s", "an option"),
            new Option("--very-long", "an option")
        };

### PR DESCRIPTION
**Pull request by belav**
closes #297

### DIFF
--- a/Src/CSharpier.Playground/Startup.cs
+++ b/Src/CSharpier.Playground/Startup.cs
@@ -45,9 +45,9 @@ namespace CSharpier.Playground
                 app.UseExceptionHandler("/Error");
                 // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
                 app.UseHsts();
+                app.UseHttpsRedirection();
             }
 
-            app.UseHttpsRedirection();
             app.UseStaticFiles();
             app.UseSpaStaticFiles();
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Src/CSharpier.Playground/Startup.cs [hunk 1]
```diff
@@ -45,9 +45,9 @@ ILoggerFactory loggerFactory
                 app.UseExceptionHandler("/Error");
                 // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
                 app.UseHsts();
+                app.UseHttpsRedirection();
```
shocklateboy92, line 48: You probably already know this, but you should really get in the habit of doing this kind of thing in a separate PR :small_airplane:

shocklateboy92 (approved):

Src/CSharpier.Playground/Startup.cs
belav: I know, especially now that there are more contributers

PR #312, opened 2021-06-14
Break long attributes consistently

### LINKED ISSUE #257: Attribute should newline before close paren
When a long attribute breaks, it currently doesn't break before the closing paren. Doing so would make it more consistent with Invocations and constructors.
```c#
[
    ExportCodeFixProvider(
        LanguageNames.CSharp,
        LanguageNames.VisualBasic,
        Name = PredefinedCodeFixProviderNames.ApplyNamingStyle),
    Shared
]
public class ClassName { }
// should be
[
    ExportCodeFixProvider(
        LanguageNames.CSharp,
        LanguageNames.VisualBasic,
        Name = PredefinedCodeFixProviderNames.ApplyNamingStyle
    ),
    Shared
]
public class ClassName { }

### PR DESCRIPTION
**Pull request by belav**
closes #257

### DIFF
--- a/Src/CSharpier.Tests/TestFiles/ObnoxiousEdgeCases/ObnoxiousEdgeCases.cst
+++ b/Src/CSharpier.Tests/TestFiles/ObnoxiousEdgeCases/ObnoxiousEdgeCases.cst
@@ -3,7 +3,8 @@
 class ClassName
 {
     public static object? FormatValue<[DynamicallyAccessedMembers(
-            DynamicallyAccessedMemberTypes.All)]
+            DynamicallyAccessedMemberTypes.All
+        )]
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Src/CSharpier.Tests/TestFiles/ObnoxiousEdgeCases/ObnoxiousEdgeCases.cst [hunk 1]
```diff
@@ -3,7 +3,8 @@
 class ClassName
 {
     public static object? FormatValue<[DynamicallyAccessedMembers(
```
shocklateboy92, line 5: What does syntax/code actually mean? :open_mouth:

shocklateboy92 (approved):

PR #313, opened 2021-06-14
Format record braces consistently with class

### LINKED ISSUE #233: Record - does not format consistently with a class.
Records should format the same as classes. This is how a record currently formats.
```c#
    public class AddResultParameter
    {
        public Guid ActionRequestId { get; set; }
        public string Name { get; set; } = string.Empty;
        public string Value { get; set; } = string.Empty;
    }
// vs
    public record AddResultParameter {
        public Guid ActionRequestId { get; set; }
        public string Name { get; set; } = string.Empty;
        public string Value { get; set; } = string.Empty;
    }
```

belav:
closes #233

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/BaseTypeDeclaration.cs
+++ b/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/BaseTypeDeclaration.cs
@@ -104,8 +104,9 @@ namespace CSharpier.SyntaxPrinter.SyntaxNodePrinters
             {
                 DocUtilities.RemoveInitialDoubleHardLine(members);
 
+                var useSpaceBraceWithParameterList = groupId != null && parameterList != null;
                 docs.Add(
-                    groupId != null && parameterList != null
+                    useSpaceBraceWithParameterList
                         ? Doc.IfBreak(" ", Doc.Line, groupId)
                         : Doc.HardLine,
                     Token.Print(node.OpenBraceToken),
--- a/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/BaseTypeDeclaration.cs
+++ b/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/BaseTypeDeclaration.cs
@@ -105,7 +105,9 @@ namespace CSharpier.SyntaxPrinter.SyntaxNodePrinters
                 DocUtilities.RemoveInitialDoubleHardLine(members);
 
                 docs.Add(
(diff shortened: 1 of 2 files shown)


shocklateboy92 (approved):

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/BaseTypeDeclaration.cs
```diff
@@ -105,7 +105,9 @@ public static Doc Print(BaseTypeDeclarationSyntax node)
                 DocUtilities.RemoveInitialDoubleHardLine(members);
 
                 docs.Add(
-                    groupId != null ? Doc.IfBreak(" ", Doc.Line, groupId) : Doc.HardLine,
+                    groupId != null && parameterList != null
```
shocklateboy92, line 108: Relatively minor comment, but it took me a while to understand you're doing this because we want to maintain space-brace when there is a parameter list.
Also the fact that you're using an `IfBreak` here because you want the opposite effect of `Line` - i.e. space if the group does break, line otherwise.

Maybe adding a comment here might make that easier to parse.

shocklateboy92 (approved):
For full disclosure, this might have been too early in the morning for me.
Add the comment I suggested if you think people will have the same amount of trouble at a reasonable time of the day :coffee:

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/BaseTypeDeclaration.cs
belav: I added a variable to help explain it. I kind of prefer that to adding comments. I think it forces someone to pay attention if they start changing code, where comments tend to get forgotten and out of date sometimes.

I think `Doc.IfBreak` is hard to parse in general, even though I've used it quite a bit at this point.
Theoretically I think something like this could be implemented, but I'm debating if it is worth the extra effort. Also it may be less obvious how to use, but more obvious what it does when you read it later.
```c#
Doc.IfGroupBreaks(" ").Else(Doc.Line)
// which could be extended to include the groupId somehow
Doc.IfGroup(groupId).Breaks(" ").Else(Doc.Line)
```

shocklateboy92: I don't think the method chain would increase readability enough to be worth the cost of implementing it.
Thinking about it some more, using named parameters would probably be enough (assuming the parameters clear names).

belav: But what if it is fun to implement??! I threw it together this morning to see what it might look like.

PR #315, opened 2021-06-14
251 - Move the second soft line wrap outside of the indentation so it…

ryanavogel:
… uses the same indentation as the opening paren.


belav (approved):
Looks solid!

PR #319, opened 2021-06-18
When printing a token, only concat the list of docs if it has >= 2 it…

ryanavogel:
…ems. This will make looking through the node graph easier while debugging.


belav (approved):
Excellent idea!
You could possibly use a switch expression + pattern matching for this, but it is good as is.
Https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#switch-expressions

shocklateboy92 (approved):
Love the commit message :1st_place_medal: :100:

PR #320, opened 2021-06-18
Better format binary expressions

divyanshshukla13:
We do a couple of things in this PR
* Put nested binary expressions onto a newline and indent
* Fix `is` operator
* Fix the condition in `do-while`
* Extract the logic for printing the rhs of an `=` operator so that it can be used by both `VariableDeclaration` and `AssignmentExpression`

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/Assignment.cs
+++ b/Src/CSharpier/SyntaxPrinter/Assignment.cs
@@ -1,0 +1,69 @@
+using System;
+using CSharpier.DocTypes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace CSharpier.SyntaxPrinter
+{
+    public static class Assignment
+    {
+        public static Doc PrintRhs(ExpressionSyntax rhs)
+        {
+            var formatMode = rhs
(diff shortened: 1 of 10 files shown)


Src/CSharpier/SyntaxPrinter/Assignment.cs
```diff
@@ -0,0 +1,69 @@
+using System;
+using CSharpier.DocTypes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace CSharpier.SyntaxPrinter
+{
+    public static class Assignment
+    {
+        public static Doc PrintRhs(ExpressionSyntax rhs)
```
belav, line 9: I'm wondering if this should be named
```
RightHandSide.Print()
```
And I'd probably go with `ExpressionSyntax node` to keep it consistent with other classes. I had to think about what `rhs` was when looking at this.

belav:
It's generally a good idea to keep a PR focused on a single thing. I say this even though I often do unrelated things in the same PR, but I'm trying to break that habit.

divyanshshukla13:
Closing this one since I've split up this PR

Issue #37, opened 2021-03-15
BinaryExpression Grouping


belav:
Take this example
```c#
var longNameForcesLineBreakYeahLongName =
    someValue == someOtherValue && x == 7;
```
The && is the root BinaryExpression, which is a LogicalAnd
The left and right of it are EqualsExpression.
CSharpier should group the LogicalAnd, but if it groups every binary Expression than the following doesn't format correctly.
```c#
if (
    longStatementName
    && longerStatementName
    && evenLongerStatementName
    && superLongStatementName
) { }
```
Which is basically a tree of BinaryExpressions where the Left is LogicalAnd BinaryExpression and the Right is IdentifierName.
This may require flattening out BinaryExpressions and conditionally grouping them based on what operation they are performing.

See prettier/src/language-js/print/binaryish.js for all the logic that goes into making this work for js

belav:
This is also related
```c#
    return value == null
    || string.IsNullOrEmpty(value.Trim());
// should be
    return value == null
        || string.IsNullOrEmpty(value.Trim());
```

belav:
This was partially addressed in https://github.com/belav/csharpier/pull/84 but requires a lot more work

belav:
this looks weird
```c#
        if (
            trivia.Kind() == SyntaxKind.SingleLineCommentTrivia
            || trivia.Kind()
            == SyntaxKind.SingleLineDocumentationCommentTrivia1111111
        ) { }
```

belav:
See also
```c#
        protected override bool IsBinaryLike(ExpressionSyntax node) =>
            node is BinaryExpressionSyntax
            || node
                is IsPatternExpressionSyntax isPattern
            && isPattern.Pattern is ConstantPatternSyntax;
```

divyanshshukla13:
Related
```csharp
while (
  // stop if we have reached the end of the token being read
  initialDepth - 1 <
    reader.Depth - (JsonTokenUtils.IsEndToken(reader.TokenType) ? 1 : 0) &&
  writeChildren &&
  reader.Read()
);
```

https://github.com/belav/Newtonsoft.Json/blob/397e45649325788b5ea299be3ca5303b86eeaf9b/Src/Newtonsoft.Json/JsonWriter.cs#L773

respel:
See also
```csharp
if (
    query
    && (currentChar == '='
    || currentChar == '<'
    || currentChar == '!'
    || currentChar == '>'
    || currentChar == '|'
[...]

PR #321, opened 2021-06-19
Better format condition of do-while

divyanshshukla13:
This change mimics the `while` statement.

Updates #37


belav:
Looks great! I added a specific issue for this so that it makes it into the change log
closes #326

belav (approved):

Issue #237, opened 2021-05-29
SwitchExpression formatting.


belav:
```c#
// currently is
            return document switch
            {
                NullDoc => indent + "Doc.Null",
                StringDoc { IsDirective: true } directive => $"{indent}Doc.Directive({directive.Value.Replace("\"", "\\\"")})",
                StringDoc stringDoc => indent + "\"" + stringDoc.Value.Replace("\"", "\\\"") + "\"",
                HardLine hardLine => indent
                + "Doc.HardLine"
                + (hardLine.Squash ? "IfNoPreviousLine" : string.Empty)
                + (hardLine.SkipBreakIfFirstInGroup ? "SkipBreakIfFirstInGroup" : string.Empty),
                LiteralLine => indent + "Doc.LiteralLine",
                Group group => @$"{indent}Doc.Group{(@group.GroupId != null ? "WithId" : string.Empty)}(
{(@group.GroupId != null ? $"{nextIndent}\"{@group.GroupId}\",{newLine}" : string.Empty)}{PrintIndentedDocTree(@group.Contents)}{newLine}{indent})",
                Concat concat => Result(concat),
                _ => throw new Exception("Can't handle " + document)   
            };
// could be
            return document switch
            {
                NullDoc
                    => indent + "Doc.Null",
                StringDoc { IsDirective: true } directive 
                    => $"{indent}Doc.Directive({directive.Value.Replace("\"", "\\\"")})",
                StringDoc stringDoc
                    => indent + "\"" + stringDoc.Value.Replace("\"", "\\\"") + "\"",
                HardLine hardLine
                    => indent
                       + "Doc.HardLine"
                       + (hardLine.Squash ? "IfNoPreviousLine" : string.Empty)
                       + (hardLine.SkipBreakIfFirstInGroup ? "SkipBreakIfFirstInGroup" : string.Empty),
                LiteralLine
                    => indent + "Doc.LiteralLine",
                Group group
                    => @$"{indent}Doc.Group{(@group.GroupId != null ? "WithId" : string.Empty)}(
{(@group.GroupId != null ? $"{nextIndent}\"{@group.GroupId}\",{newLine}" : string.Empty)}{PrintIndentedDocTree(@group.Contents)}{newLine}{indent})",
                Concat concat
                    => Result(concat),
                _   
                    => throw new Exception("Can't handle " + document)   
            };
// or maybe
            return document switch
            {
                NullDoc                                   => indent + "Doc.Null",
[...]

PR #324, opened 2021-06-19
Use common logic for printing rhs of assignment expressions

divyanshshukla13:
Updates #37

### DIFF
--- a/Src/CSharpier.Tests/TestFiles/InitializerExpression/InitializerExpressions.cst
+++ b/Src/CSharpier.Tests/TestFiles/InitializerExpression/InitializerExpressions.cst
@@ -62,8 +62,7 @@ class ClassName
         var collectionInitializerExpressions = new SomeObject
         {
             ThisIsACollection = { one, two },
-            ThisIsAlsoACollection =
-            {
+            ThisIsAlsoACollection = {
                 longerValueThatMakesItBreak______________________,
                 longerValueThatMakesItBreak______________________
             },
--- a/Src/CSharpier.Tests/TestFiles/InitializerExpression/InitializerExpressions.cst
+++ b/Src/CSharpier.Tests/TestFiles/InitializerExpression/InitializerExpressions.cst
@@ -62,7 +62,8 @@ class ClassName
         var collectionInitializerExpressions = new SomeObject
         {
             ThisIsACollection = { one, two },
-            ThisIsAlsoACollection = {
+            ThisIsAlsoACollection =
+            {
                 longerValueThatMakesItBreak______________________,
                 longerValueThatMakesItBreak______________________
             },
--- a/Src/CSharpier/SyntaxPrinter/RightHandSide.cs
+++ b/Src/CSharpier/SyntaxPrinter/RightHandSide.cs
@@ -1,0 +1,69 @@
+using System;
+using CSharpier.DocTypes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace CSharpier.SyntaxPrinter
+{
+    public static class RightHandSide
+    {
+        public static Doc Print(ExpressionSyntax node)
+        {
+            var formatMode = node
+                is AnonymousObjectCreationExpressionSyntax
+                    or AnonymousMethodExpressionSyntax
+                    or InitializerExpressionSyntax
+                    or InvocationExpressionSyntax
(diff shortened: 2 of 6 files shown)


Src/CSharpier.Tests/TestFiles/InitializerExpression/InitializerExpressions.cst
```diff
@@ -62,8 +62,7 @@ class ClassName
         var collectionInitializerExpressions = new SomeObject
         {
             ThisIsACollection = { one, two },
-            ThisIsAlsoACollection =
-            {
+            ThisIsAlsoACollection = {
```
belav, line 65: I think we want to keep the old formatting on this, but your change does get this consistent with `arrayInitializerWithoutSize` above it. It could also be a new issue.

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/AssignmentExpression.cs
```diff
@@ -12,12 +12,7 @@ public static Doc Print(AssignmentExpressionSyntax node)
                 Node.Print(node.Left),
                 " ",
                 Token.Print(node.OperatorToken),
-                node.Right is QueryExpressionSyntax
-                    ? Doc.Indent(Doc.Line, Node.Print(node.Right))
-                    : Doc.Group(
-                            node.Right is InitializerExpressionSyntax ? Doc.Null : " ",
-                            Node.Print(node.Right)
-                        )
+                RightHandSide.Print(node.Right)
```
belav, line 15: 👍 for reusing the code from VariableDeclaration in here

belav (commented):
Looks good, just the potential change to put the brace on a new line for collection initializers that break.

Src/CSharpier/SyntaxPrinter/RightHandSide.cs
```diff
@@ -0,0 +1,69 @@
...
+                    or AnonymousMethodExpressionSyntax
+                    or InitializerExpressionSyntax
+                    or InvocationExpressionSyntax
+                    or ConditionalExpressionSyntax
+                    or ObjectCreationExpressionSyntax
+                    or SwitchExpressionSyntax
+                    or LambdaExpressionSyntax
+                    or AwaitExpressionSyntax
+                    or WithExpressionSyntax
+                ? FormatMode.NoIndent
```
shocklateboy92, line 22: This looks like a nested ternary if on type matches. Usually, a [switch statement](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#multiple-inputs) looks better.

Not sure how good it will look with a dozen `or`s. But if it doesn't look good you can probably split them into multiple arms of the switch.

Src/CSharpier/SyntaxPrinter/RightHandSide.cs
```diff
@@ -0,0 +1,69 @@
...
+                        or ArrayCreationExpressionSyntax
+                        or ImplicitArrayCreationExpressionSyntax
+                    ? Doc.IndentIfBreak(Doc.Group(Node.Print(node)), groupId)
+                    : Doc.Group(IndentIfNeeded(node, formatMode))
+            );
+        }
+
+        private static Doc IndentIfNeeded(ExpressionSyntax initializerValue, FormatMode formatMode)
+        {
+            if (formatMode is FormatMode.NoIndent)
```
shocklateboy92, line 51: I think this whole function could be replaced by an expression bodied method with a single switch statement. Up to you if you think this code looks better though.

shocklateboy92 (commented):
More pattern matching magic can make this code more magical... :mage_man: :star2:

Src/CSharpier/SyntaxPrinter/RightHandSide.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace CSharpier.SyntaxPrinter
+{
+    public static class RightHandSide
+    {
+        public static Doc Print(ExpressionSyntax node)
+        {
+            var groupId = Guid.NewGuid().ToString();
+            return node switch
```
belav, line 13: This gives us another good test case for #237.

belav (approved):
👍

PR #329, opened 2021-06-20
Print checked the same way as an invocationExpression

### LINKED ISSUE #270: Format checked like a invocation expression with an argument list
```c#
var retVal = new byte[
    checked(1 /* KDF alg */
    + 1 /* chaining mode */
    + sizeof(uint) /* sym alg key size */
    + sizeof(uint) /* sym alg block size */
    + sizeof(uint) /* hmac alg key size */
    + sizeof(uint) /* hmac alg digest size */
    + _symmetricAlgorithmBlockSizeInBytes /* ciphertext of encrypted empty string */
    + _hmacAlgorithmDigestLengthInBytes /* digest of HMACed empty string */
    )
];
// should maybe be this? or should we break after the checked( ?
var retVal = new byte[
    checked(1 /* KDF alg */
        + 1 /* chaining mode */
        + sizeof(uint) /* sym alg key size */
        + sizeof(uint) /* sym alg block size */
        + sizeof(uint) /* hmac alg key size */
        + sizeof(uint) /* hmac alg digest size */
        + _symmetricAlgorithmBlockSizeInBytes /* ciphertext of encrypted empty string */
        + _hmacAlgorithmDigestLengthInBytes /* digest of HMACed empty string */
    )
];

### PR DESCRIPTION
**Pull request by belav**
closes #270

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/CheckedExpression.cs
+++ b/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/CheckedExpression.cs
@@ -1,5 +1,4 @@
 using CSharpier.DocTypes;
-using CSharpier.SyntaxPrinter;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace CSharpier.SyntaxPrinter.SyntaxNodePrinters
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/CheckedExpression.cs [hunk 1]
```diff
@@ -10,9 +9,12 @@ public static Doc Print(CheckedExpressionSyntax node)
         {
             return Doc.Concat(
                 Token.Print(node.Keyword),
-                Token.Print(node.OpenParenToken),
-                Node.Print(node.Expression),
-                Token.Print(node.CloseParenToken)
+                Doc.Group(
+                    Token.Print(node.OpenParenToken),
```
shocklateboy92, line 13: Should we maybe have a helper method for invocation expression like things?

shocklateboy92 (approved):

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/CheckedExpression.cs
belav: I had a similar thought, but it was different enough from how parameters are printed for Invocations that I didn't think it made sense to try to combine them. I just looked around and there are a lot of SyntaxNodePrinters that print the same style for the parens, but they all seem different enough that it would be hard to share the logic. : /

PR #333, opened 2021-06-20
Break and Tab Parameters with Attributes

### LINKED ISSUE #204: Attributes on parameters
Currently when a method has parameters with attributes it looks like
```c#
    public void MethodName(
        int someParameter,
        [LongAttributeName(SomeFlag.SomeValue)]
        AnotherObject anotherObject,
        string anotherValue
    ) {
        return;
    }
```
Which makes it a little hard to tell where the parameters are.
We could do something like below

```c#
    public void MethodName(
        int someParameter,
        [ShortAttribute] AnotherObject anotherObject,
        string anotherValue
    ) {
        return;
    }

    public void MethodName(
        int someParameter,
        [LongAttributeName(SomeFlag.SomeValue, "MoreStuffToMakeThisLong")]
            AnotherObject anotherObject,
        string anotherValue
    ) {
        return;
    }
```

### LINKED ISSUE #334: New overload for `Doc.GroupWithId()`
https://github.com/belav/csharpier/blob/7ed2638213b9320b05141b33b51b8ada2fcb62d5/Src/CSharpier/DocTypes/Doc.cs#L84
should probably have another overload that takes a `List<Doc>`
so that we don't have instances like this:
https://github.com/belav/csharpier/blob/5583107d91fd54f4b07a3b791f4b8b1cd9a28855/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Parameter.cs#L37

where we convert a `List` into an `Array` and back into a `List` in the `Group()` constructor.

brikabrak:
Added breaking and tab logic to parameter when attribute is involved, while grouping the parameter itself as that allowed for a custom break. Removed unnecessary `using` statement.

Ignored `.vscode` for people who think that it is an IDE.

Fixes #204.
Fixes #334.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@
 # Jetbrains doesn't say to ignore these, but it seems like they shouldn't be committed
 **/.idea/**/watcherTasks.xml
 
+.vscode
+
 node_modules
 bin
 obj
--- a/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Parameter.cs
+++ b/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Parameter.cs
@@ -34,7 +34,7 @@ namespace CSharpier.SyntaxPrinter.SyntaxNodePrinters
             if (hasAttribute)
             {
                 docs.Add(Doc.Concat(paramDocs));
-                return Doc.Concat(Doc.GroupWithId(groupId, docs.ToArray()));
+                return Doc.GroupWithId(groupId, docs.ToArray());
             }
             else
             {
--- a/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Parameter.cs
+++ b/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Parameter.cs
@@ -19,28 +19,18 @@ namespace CSharpier.SyntaxPrinter.SyntaxNodePrinters
                 Modifiers.Print(node.Modifiers)
             };
 
-            var paramDocs = new List<Doc>();
             if (node.Type != null)
             {
-                paramDocs.Add(Node.Print(node.Type), " ");
+                docs.Add(Node.Print(node.Type), " ");
             }
 
-            paramDocs.Add(Token.Print(node.Identifier));
+            docs.Add(Token.Print(node.Identifier));
             if (node.Default != null)
             {
-                paramDocs.Add(EqualsValueClause.Print(node.Default));
+                docs.Add(EqualsValueClause.Print(node.Default));
             }
 
-            if (hasAttribute)
-            {
-                docs.Add(Doc.Concat(paramDocs));
-                return Doc.GroupWithId(groupId, docs.ToArray());
-            }
-            else
-            {
-                docs.AddRange(paramDocs);
-                return Doc.Concat(docs);
-            }
+            return hasAttribute ? Doc.GroupWithId(groupId, docs) : Doc.Concat(docs);
         }
     }
 }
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -26,6 +26,8 @@
 # Jetbrains doesn't say to ignore these, but it seems like they shouldn't be committed
 **/.idea/**/watcherTasks.xml
 
+.vscode
```
shocklateboy92, line 29: I don't know if it makes sense to ignore the whole directory.
There are a couple of things that are useful to share between people in the team (e.g. build tasks, extensions, format-on-save etc.)

Why don't you just commit your project specific settings?

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Parameter.cs
```diff
@@ -10,24 +10,37 @@ public static class Parameter
...
             {
-                docs.Add(EqualsValueClause.Print(node.Default));
+                paramDocs.Add(EqualsValueClause.Print(node.Default));
             }
 
-            return Doc.Concat(docs);
+            if (hasAttribute)
+            {
+                docs.Add(Doc.Concat(paramDocs));
+                return Doc.Concat(Doc.GroupWithId(groupId, docs.ToArray()));
```
shocklateboy92, line 37: Why does this group need to be in a `Concat()`?

brikabrak: More than likely a holdover from how I read the docs. Will change and see if tests still pass.

.gitignore
brikabrak: I typically ignore the root `.vscode` folder until someone has those reusable config items mentioned. In my case I'm using VsCodium which needs to leverage an environment path-specific debugger. Folks don't need my dev environment items in their codebase.

If a useful item pops up from another dev, they could add a caveat line to allow it.

belav (changes requested):
I just had a couple minor things, otherwise it looks good!

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Parameter.cs
```diff
@@ -10,24 +10,37 @@ public static class Parameter
...
             {
-                docs.Add(EqualsValueClause.Print(node.Default));
+                paramDocs.Add(EqualsValueClause.Print(node.Default));
             }
 
-            return Doc.Concat(docs);
+            if (hasAttribute)
+            {
+                docs.Add(Doc.Concat(paramDocs));
+                return Doc.GroupWithId(groupId, docs.ToArray());
```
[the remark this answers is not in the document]
belav: Just realized @shocklateboy92 had a similar idea and created an issue for it https://github.com/belav/csharpier/issues/334
So this can just get updated when someone picks up that issue.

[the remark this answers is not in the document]
brikabrak: I could solve that issue as part of this, not a big issue to do so.

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Parameter.cs
```diff
@@ -10,24 +10,37 @@ public static class Parameter
...
+            var groupId = hasAttribute ? Guid.NewGuid().ToString() : string.Empty;
             var docs = new List<Doc>
             {
                 AttributeLists.Print(node, node.AttributeLists),
-                node.AttributeLists.Any() ? Doc.Line : Doc.Null,
+                hasAttribute ? Doc.IndentIfBreak(Doc.Line, groupId) : Doc.Null,
                 Modifiers.Print(node.Modifiers)
             };
+
+            var paramDocs = new List<Doc>();
```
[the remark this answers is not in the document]
brikabrak: At a certain point it was holding onto a separate grouping logic, but that's no longer the case so yes it can be simplified now!

belav (approved):
👍 ❤️

PR #340, opened 2021-06-26
Formatting with 0.9.7

belav:


Src/CSharpier/DocPrinter/DocFitter.cs
```diff
@@ -94,8 +94,7 @@ void Push(Doc doc, PrintMode printMode, Indent indent)
 
                             // when determining if something fits, use the last option from a conditionalGroup, which should be the most expanded one
                             var groupContents = groupMode == PrintMode.Break
-                            && group
-                                is ConditionalGroup conditionalGroup
+                            && group is ConditionalGroup conditionalGroup
```
shocklateboy92, line 97: This should be indented. Is this tracked by a bug?

Src/CSharpier/SyntaxNodeComparer.cs
```diff
@@ -147,10 +148,8 @@ TextSpan newParentSpan
                 }
 
                 if (
-                    originalList[x]
-                        is SyntaxNode originalNode
-                    && formattedList[x]
-                        is SyntaxNode formattedNode
+                    originalList[x] is SyntaxNode originalNode
+                    && formattedList[x] is SyntaxNode formattedNode
```
shocklateboy92, line 152: Did we have a regression somewhere? I swear we used to indent binary operators

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/InitializerExpression.cs
```diff
@@ -37,8 +37,7 @@ private static Doc Print(InitializerExpressionSyntax node, string? groupId)
                 is not (ObjectCreationExpressionSyntax
                     or ArrayCreationExpressionSyntax
                     or ImplicitArrayCreationExpressionSyntax)
-            && node.Kind()
-                is not SyntaxKind.WithInitializerExpression ? Doc.Group(result) : result;
+            && node.Kind() is not SyntaxKind.WithInitializerExpression ? Doc.Group(result) : result;
```
shocklateboy92, line 40: What's going on here? Is this below the line length or do we have a bug?

Also, I think we should definitely break/indent the ternary operator if the condition needed a break.

PR #342, opened 2021-06-27
Fix issue with validation failing for mismatched line endings in verbatim strings

### LINKED ISSUE #244: Verbatim string with mismatched line endings triggers "failed syntax tree validation"
Use "auto" line endings.
Format the following file
```c#
public class ClassName
{
    public string Value = @"EndThisLineWith\r\n
EndThisLineWith\n
";
}
CSharpier will trigger the "failed syntax tree vaildation" message because it is comparing the line endings in the verbatim string.

belav:
closes #244


shocklateboy92 (approved):
LGTM

Issue #286, opened 2021-06-11
Automate formatting a large repo to test CSharpier changes


belav:
I've been doing the following manually and it helps track down edge cases that are missed by our current tests before we get to the release candidate stage. Automating this process would make it easier.

When a PR is created.

1. Create a new branch `pre-[nameOfCSharpierBranch]` off of master on https://github.com/belav/aspnetcore
2. Format the new aspnetcore branch with the current CSharpier master code and commit it.
3. Format the new aspnetcore branch with the CSharpier code from the PR and commit it.
4. Automatically add a message the CSharpier PR linking to the new commit.

3 and 4 could be repeated with each new commit to the PR.
This process could also be done using a script locally, to avoid having to create the PR first.
This process also could use a different repo, or be configurable if running locally.

belav:
I started working on this, but there are a number of issues to figure out.
This is the current action I had, but I haven't tested it.
```
name: Validate PR
on:
  pull_request:
    branches: [ master ]
jobs:
  format_repositories:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
        with:
          path: csharpier
      - run: echo "##[set-output name=branch;]$(echo ${GITHUB_HEAD_REF #refs/heads/})"
        id: extract_branch
        shell: bash
      - uses: actions/checkout@v2
        with:
          repository: belav/aspnetcore
          path: aspnetcore
      - uses: actions/checkout@v2
        with:
          path: csharpier-master
          ref: master
      - run: dotnet build csharpier-master/Src/CSharpier/CSharpier.csproj -c release
      - run: dotnet csharpier-master/Src/CSharpier/bin/release/net5.0/dotnet-csharpier.dll aspnetcore
      - run: |
          git checkout -b pre-${{ steps.extract_branch.outputs.branch}}
          git add -A
          git commit -m "Before changes"
          git push --set-upstream origin pre-${{ steps.extract_branch.outputs.branch}}
      - run: dotnet build csharpier/Src/CSharpier/CSharpier.csproj -c release
      - run: dotnet csharpier/Src/CSharpier/bin/release/net5.0/dotnet-csharpier.dll aspnetcore
      - run: |
          git checkout -b post-${{ steps.extract_branch.outputs.branch}}
          git add -A
          git commit -m "After changes"
          git push --set-upstream origin post-${{ steps.extract_branch.outputs.branch}}
```
The issues I know of

1. It would try to create a new pre-branch on every commit to the csharpier PR
[...]

PR #351, opened 2021-07-11
Scripting way to test formatting repo

belav:
This relates to #286, but doesn't yet close it. Ideally this would run automatically as PRs are created, and comment on those PRs.


shocklateboy92 (approved):
I think I sort of see what you're doing here.
But I'm not an expert on powershell, so what looks good to me might not mean much.

PR #359, opened 2021-07-17
Remove most configuration options

### LINKED ISSUE #358: Remove all configuration options except for print width.
CSharpier is supposed to be an opinionated code formatter, but it supplies options for tabs vs spaces, tab width, and line endings.

These options should be removed to keep CSharpier opinionated.

belav:
If csharpier is opinionated, it shouldn't offer options beyond print width.
This hides those options, but doesn't yet clean up the code related to them.
closes #358


shocklateboy92 (approved):
LGTM

PR #368, opened 2021-07-19
Getting ForEachVariable formatting the same as a regular ForEach

### LINKED ISSUE #300: ForEachVariableStatement not breaking/indenting consistently with regular ForEachStatement
I'm not exactly sure what this should look like, but it definitely isn't this.
```c#
foreach ((
    (int l2, int l3),
    (int total, string[] stats)
) in (from l2 in Enumerable.Range(1, 7)
from l3 in Enumerable.Range(1, 7)
select (l2, l3)).AsParallel()
    .Select(l => (l, res: GenerateTable(l.l2, l.l3, cutOff)))
    .OrderBy(v => v.res.Total))
{
    Console.WriteLine($"Stats for {l2}:{l3}");
    foreach (var line in stats) Console.WriteLine(line);
}

### PR DESCRIPTION
**Pull request by belav**
closes #300

### CODE REVIEW

File: Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/CommonForEachStatement.cs [hunk 1]
```diff
@@ -6,16 +6,30 @@
...
+        public static Doc Print(CommonForEachStatementSyntax node)
         {
             var groupId = Guid.NewGuid().ToString();
 
             var leadingTrivia = node.AwaitKeyword.Kind() != SyntaxKind.None
                 ? Token.PrintLeadingTrivia(node.AwaitKeyword)
                 : Token.PrintLeadingTrivia(node.ForEachKeyword);
 
+            Doc variable = Doc.Null;
+            if (node is ForEachStatementSyntax forEachStatementSyntax)
```
shocklateboy92, line 20: So, you don't have to drink the kool-aid if you don't want to this time, but how would this look if the whole thing were an inline switch expression?

shocklateboy92 (approved):
More kool-aid if you want it

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/CommonForEachStatement.cs
belav: better!

Issue #345, opened 2021-07-11
Conditional breaking without indentation


belav:
The indentation of the `&&` is probably already another ticket. This is more about the Ternary breaking.
```c#
Doc statementSeparator = node.Parent is AccessorDeclarationSyntax
&& node.Statements.Count <= 1 ? Doc.Line : Doc.HardLine;

// should be something like

Doc statementSeparator = 
    node.Parent is AccessorDeclarationSyntax
        && node.Statements.Count <= 1 
    ? Doc.Line
    : Doc.HardLine;

**Comment by belav** [comment]
See also
```c#
        public AccessTokenNotAvailableException(
            NavigationManager navigation,
            AccessTokenResult tokenResult,
            IEnumerable<string> scopes
        ) : base(
            message: "Unable to provision an access token for the requested scopes: " + scopes
            != null ? $"'{string.Join(", ", scopes ?? Array.Empty<string>())}'" : "(default scopes)"
        ) { }
```

belav:
Add
```c#
            fixed (
                byte* pbHeapAllocatedPasswordBuffer = (
                    cbPasswordBuffer
                    > Constants.MAX_STACKALLOC_BYTES
                ) ? new byte[cbPasswordBuffer] : null
            ) { }

Issue #272, opened 2021-06-07
Obnoxious Edge Cases


belav:
There are a number of obnoxious edge cases that we should address. They were added as tests so that when formatting changes that affects them, we can see how they evolve and hopefully address them somewhat with related bugs.

At some later date we should try to get them cleaned up. Hopefully that will just happen.

shocklateboy92:
Adding a link to [the file](https://github.com/belav/csharpier/blob/a4ba7787f00cb63e90bd821770860df82262ab37/Src/CSharpier.Tests/TestFiles/ObnoxiousEdgeCases/ObnoxiousEdgeCases.cst), for easy reference.

PR #370, opened 2021-07-24
Cleaning up formatting for nested TypeArguments

### LINKED ISSUE #295: Nested Generics in Variable Declaration
I'm not exactly sure what to do with this, but it doesn't look the best right now.
```
class ClassName
{
    void MethodName()
    {
        PropertyMapActions.Add(
            pm =>
            {
                Expression<Func<
                        TSource,
                        TDestination,
                        TMember,
                        TMember,
                        ResolutionContext,
                        bool
                    >> expr = (src, dest, srcMember, destMember, ctxt) => condition(
                    src,
                    dest,
                    srcMember,
                    destMember,
                    ctxt
                );
            }
        );
    }
}

### PR DESCRIPTION
**Pull request by belav**
closes #295

### DIFF
--- a/Src/CSharpier.Tests/TestFiles/ObnoxiousEdgeCases/ObnoxiousEdgeCases.cst
+++ b/Src/CSharpier.Tests/TestFiles/ObnoxiousEdgeCases/ObnoxiousEdgeCases.cst
@@ -2,15 +2,13 @@
 // try to clean these up if you are modifying formatting that affects them.
 class ClassName
 {
-    public readonly StackObjectPool<Dictionary<
-            object,
-            KeyedItemInfo
-        >> KeyedItemInfoDictionaryPool_____;
+    public readonly StackObjectPool<
+        Dictionary<object, KeyedItemInfo>
+    > KeyedItemInfoDictionaryPool_____;
 
-    internal StackObjectPool<Dictionary<
-            object,
-            KeyedItemInfo
-        >> KeyedItemInfoDictionaryPool { get; } =
+    internal StackObjectPool<
+        Dictionary<object, KeyedItemInfo>
+    > KeyedItemInfoDictionaryPool { get; } =
         new StackObjectPool<Dictionary<object, KeyedItemInfo>>(
             maxPreservedItems: 10,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/TypeArgumentList.cs [hunk 1]
```diff
@@ -9,13 +9,16 @@ public static class TypeArgumentList
     {
         public static Doc Print(TypeArgumentListSyntax node)
         {
+            Doc separator = node.Arguments.Count > 1
+            || node.Arguments.Any(o => o is GenericNameSyntax) ? Doc.SoftLine : Doc.Null;
```
shocklateboy92, line 13: Odd. Why didn't this indent?

Src/CSharpier.Tests/TestFiles/ObnoxiousEdgeCases/ObnoxiousEdgeCases.cst
```diff
@@ -2,15 +2,13 @@
 // try to clean these up if you are modifying formatting that affects them.
```
shocklateboy92, line 2: Is this comment still necessary now that you've cleaned these up, or do you have more cleanup you intend to do?

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/TypeArgumentList.cs
belav: I just checked this in 0.9.8 and it looks better now.
There is also #345 for improving some of the edge cases.

Src/CSharpier.Tests/TestFiles/ObnoxiousEdgeCases/ObnoxiousEdgeCases.cst
belav: Just glancing at them, there are probably still some improvements to be made. I mostly have the comment here so that people don't think this is really how we want things to look. If I ever get around to #272 this file can go away completely and they'll turn into proper test cases

PR #379, opened 2021-07-27
Format with 0.9.8

belav:


shocklateboy92:
What happened here?

belav:
I couldn't get nuget to download 0.9.8 locally, created this PR to see if github could see the new version, then figured out that the --check github action wasn't failing builds if code wasn't formatted. I don't recall why I closed this PR and created the new one though...

PR #380, opened 2021-07-27
Format with 0.9.8

belav:


shocklateboy92 (commented):
All in all, csharpier is shaping up nicely.
I think this might be the version I will inflict upon my coworkers.

PR #384, opened 2021-07-29
Prefer breaking before `when` clause, instead of inside

### LINKED ISSUE #382: Line breaks in "when" clause in SwitchExpression
I came across some strange formatting of a long pattern statement with a when clause. I think CSharpier should prefer putting a line break before the `when` expression over breaking up the `when` condition.

Current behaviour:
```
someValue switch
{
    "SomeLongishValue______________________________________________" when count
        > 0
      => true
    _ => false
}
```
Expected behaviour:
```
someValue switch
{
    "SomeLongishValue______________________________________________"
        when count > 0
      => true
    _ => false
}
```

I did notice an improvement from 0.9.7 to 0.9.8, the previous formatting was:
```
someValue switch
{
    "SomeLongishValue______________________________________________" when count
    > 0 => true
    _ => false
}
```

pmccloghrylaing:
Closes #382

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/WhenClause.cs
+++ b/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/WhenClause.cs
@@ -1,4 +1,3 @@
-using System;
 using CSharpier.DocTypes;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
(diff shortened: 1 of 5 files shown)


Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/WhenClause.cs
```diff
@@ -7,9 +8,14 @@ public static class WhenClause
     {
         public static Doc Print(WhenClauseSyntax node)
         {
-            return Doc.Concat(
-                Token.PrintWithSuffix(node.WhenKeyword, " "),
-                Node.Print(node.Condition)
+            var groupId = Guid.NewGuid().ToString();
+            return Doc.GroupWithId(
```
shocklateboy92, line 12: Is there any particular reason you're making this a `GroupWithId` instead of just a `Group`? 
It doesn't look like you're doing anything with the id.

pmccloghrylaing: Ah - I had thought I needed to use `IndentIfBreak`. I'll change it to `Group`.

shocklateboy92 (approved):

shocklateboy92:
Thanks! 🫂😊

PR #386, opened 2021-08-02
Respect empty lines between case statements and switch expression patterns

### LINKED ISSUE #383: Respect new lines between case statements
The current behaviour removes any new lines between case statements, so it converts:
```csharp
switch (value)
{

    case 0:
        break;


    case 1:
        break;

}
```
to:
```csharp
switch (value)
{
    case 0:
        break;
    case 1:
        break;
}
```

I think it should maintain an extra space, just like between statements in a block:
```csharp
switch (value)
{
    case 0:
        break;

    case 1:
        break;
}
```

pmccloghrylaing:
Respect empty lines between case statements and switch expression patterns

Closes #383


shocklateboy92 (approved):
LGTM. Thanks!

PR #388, opened 2021-08-03
Require directoryOrFile argument when not piping stdin

### LINKED ISSUE #381: Require the directoryOrFile argument when not piping into to stdin
The --check option is not longer reporting missing files when used inside of a github action.

This was most likely caused by the changes for #282
Going back to 0.9.4, the current github action will fail a build for unformatted files.
As of 0.9.8 it definitely does not.
#282 came out in 0.9.6

After resolving this bug, also add a step to the ValidatePR that tests --check to ensure we don't break this again.

belav:
This also includes creating a number of tests that run with a github workflow to ensure we don't introduce bugs into the CLI with any later changes.
closes #381


.github/workflows/validatePR.yml
```diff
@@ -9,20 +9,19 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: dotnet test Src/CSharpier.Tests/CSharpier.Tests.csproj -c release
-  test_stdin:
-    runs-on: ubuntu-latest
```
shocklateboy92, line 13: Are you sure this is worth burning through our free build minutes at double rate?
If you prefer powershell over bash, you can probably upgrade to [powershell core 6](https://aka.ms/pscore6) which runs in both Linux and Windows.

If you're just unfamiliar with bash/Linux, we can have a discord session together :fox_face:

belav: From what I saw, we get unlimited minutes because it's a public repo. But it was super easy to switch over to powershell core and there is no real need to run it on windows.

I should probably learn bash some day, but I feel like there is too much else to do!

belav:
I wanted to use TestCli for recreating/testing #392, so I included that in this PR. 
closes #392

Scripts/CreateTestingPR.ps1
```diff
@@ -34,7 +34,7 @@ if ($firstRun)
     & git reset --hard
     & git checkout -b $preBranch
 
-    dotnet $csharpierProject\Src\CSharpier\bin\Release\net5.0\dotnet-csharpier.dll
+    dotnet $csharpierProject\Src\CSharpier\bin\Release\net5.0\dotnet-csharpier.dll .
```
shocklateboy92, line 37: probably worth making a `$csharpier` variable like you do in `TestCli.ps`.

Scripts/CreateReviewCodePRs.ps1
```diff
@@ -46,7 +46,7 @@ foreach($folder in Get-ChildItem $csharpierRepos) {
 
 Push-Location $csharpierRepos
 
-dotnet $csharpierProject\Src\CSharpier\bin\Release\net5.0\dotnet-csharpier.dll
+dotnet $csharpierProject\Src\CSharpier\bin\Release\net5.0\dotnet-csharpier.dll .
```
shocklateboy92, line 49: The 2 comments I made about `CreateTestingPr.ps1` apply to this file too.

Scripts/CreateTestingPR.ps1
shocklateboy92: Also, github won't let me make comments on lines that are not part of the diff, but it's probably worth putting `Release` into a variable and referencing it here and in the build command.

You'll probably never change build configuration, but I'm a stickler for not repeating code :smirk:

shocklateboy92 (approved):
Okay, actually looked through the PS code this time.
Looks pretty good - made some minor comments that might be worth doing.

Also, is there an easy way of importing other scripts in powershell? You have a lot of repeated variables (especially involving paths) in your scripts. If you can extract them to some common file and import them, it's probably worth doing.

PR #389, opened 2021-08-04
Fixing formatting of align in serializer

belav:


shocklateboy92 (approved):
On a (semi) related note, have you considered making doc types records instead of classes?
Then you would get _most_ of the serialization for free.

belav:
> On a (semi) related note, have you considered making doc types records instead of classes?
> Then you would get _most_ of the serialization for free.

I hadn't considered it but I just did some playing around now and I'm not sure records would provide too much benefit.

The `DocSerialier` serializes things in a way that provides you working c# code to reproduce the Doc. Maybe it should be named something else because that isn't really how you'd typically serialize anything.
```c#
Doc.Concat(
    " ",
    Doc.Null,
    Doc.Null,
    "get",
    ";"
)
```


With a basic version of record Docs, I get this before any customization
```c#
Concat { contents = System.Collections.Generic.List`1[Doc] }
```

PR #402, opened 2021-08-16
Adding new error for unsupported file type. Changing how errors and warnings print.

### LINKED ISSUE #398: Better error handling when given a csproj or sln file
When I attempted to use this, I misread the command line instructions and passed in a `.sln` file or a `.csproj` file because this is how the [dotnet formatter](https://github.com/dotnet/format) works.  This resulted in this output:

```
+dotnet csharpier Foo.sln
/Foo.sln                                                                                                                 - failed to compile
total time:                                                                          243ms
Total files:                                                                           1
Failed syntax tree validation:                                                         0
Threw exceptions while formatting:                                                     0
files that threw exceptions while validating syntax tree:                              0
```

Which led me to believe that the sln file was not compiling, and not that I was supplying an invalid file.  I understand now how it is meant to work, but wanted to just suggest that this error message be improved.   When not given a `.cs` file or a directory, can it print out clearer instructions?

EDIT:  Or, alternatively, actually process csproj / sln files and format all the C# scripts inside them

belav:
closes #398

### DIFF
--- a/Src/CSharpier/CommandLineFormatter.cs
+++ b/Src/CSharpier/CommandLineFormatter.cs
@@ -169,6 +169,12 @@ namespace CSharpier
                 return;
             }
 
+            if (!filePath.EndsWithIgnoreCase(".cs") && !filePath.EndsWithIgnoreCase(".cst"))
+            {
+                WriteError(filePath, "Is an unsupported file type.");
+                return;
+            }
+
             cancellationToken.ThrowIfCancellationRequested();
 
             var (encoding, fileContents, unableToDetectEncoding) = await FileReader.ReadFile(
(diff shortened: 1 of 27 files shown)


Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -349,6 +353,35 @@ private bool ShouldIgnoreFile(string filePath)
                 || this.IgnoreFile.IsIgnored(filePath);
         }
 
+        private void WriteError(string filePath, string value)
+        {
+            this.WriteMessage(filePath, value, "Error", ConsoleColor.DarkRed);
+        }
+
+        private void WriteWarning(string filePath, string value)
```
shocklateboy92, line 361: You know, have you considered using an `ILogger`? 
https://docs.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line

The more complex this gets, the more it feels like you're re-implementing the wheel. :smirk:

shocklateboy92 (approved):
You should definitely think about the comment I made, but I'm going to approve/merge this since it doesn't make sense to change as part of this PR.

Issue #404, opened 2021-08-17
Improve if directive formatting


belav:
Some things that don't work yet.

1. nested directives
2. if the final else needs a specific combination, that won't be in the set
3. we don't actually look at not's correctly. Meaning `#IF !ONE && TWO` is parsed as ONE TWO, instead of just TWO
4. related, for something like `#IF ONE || TWO` we only need one of the two symbols
5. we may want to allow specifying symbolSets by file
6. #121

belav:
https://github.com/belav/aspnetcore/blob/main/src/ProjectTemplates/Web.ProjectTemplates/content/BlazorServerWeb-CSharp/Startup.cs has some good examples

PR #405, opened 2021-08-17
Support for basic if/elif directives

### LINKED ISSUE #15: Code in IfDirective can't currently be formatted
Given the following IfDirective, roslyn does not parse the contents because DEBUG is not defined
```c#
#if DEBUG
        Console.WriteLine("Rosyln does not parse this because DEBUG is not defined");
#endif
```
The contents are just a block of text.
Can I determine some way to parse those contents and format them? Currently they are just written out as is.

belav:
closes #15

This handles the basics of formatting code in if directives and supplies a configuration option for more complex cases.
#404 will improve upon this initial work


belav:
I tested this across the whole of https://github.com/belav/csharpier-repos and came across

- An unrelated issue with switch/pattern matching that just happened to be in an `#if` #408
- An edge case that feels low priority #410
- Instances of files that won't compile because they have bad code inside of an `#if`

There may be some other edge cases similar to #410, I haven't reviewed the actual formatting changes this makes except on the aspnetcore repo. But I think this can be deemed good enough for the first pass.

Src/CSharpier.Tests/ConfigurationFileOptionsTests.cs
```diff
@@ -36,22 +37,43 @@ public void Should_Return_Default_Options_With_No_File()
...
+                @"
+printWidth: 10
+preprocessorSymbolSets: 
+  - 
+    - 1
+    - 2
+  - 
+    - 3
+"
+            );
```
belav, line 70: this feels ugly. Should it use comma delimited strings instead? 
```
preprocessorSymbolSets:
    - DEBUG,INCLUDE_X
    - INCLUDE_Y
```
vs
```
preprocessorSymbolSets:
    - 
        - DEBUG
        - INCLUDE_X
    - 
        - INCLUDE_Y
```

PR #426, opened 2021-08-28
Making use of ILogger

### LINKED ISSUE #406: Implement proper logging.
The console logging that used to be simple, has gotten to a point it should probably switch to a proper logger instead of using Console.WriteLine and setting console colors.

See https://github.com/dotnet/format/tree/main/src/Logging for some ideas on how to get going.

belav:
closes #406


Src/CSharpier/CommandLineFormatter.cs
```diff
@@ -22,6 +22,7 @@ public class CommandLineFormatter
         protected readonly IFileSystem FileSystem;
         protected readonly IConsole Console;
         protected readonly IgnoreFile IgnoreFile;
+        protected readonly ILogger Logger;
```
shocklateboy92, line 25: You should take `ILogger<CommandLineFormatter>`.
That way, you can set an option for it to print the scope/source of the message should you chose to later.

Src/CSharpier/ConsoleLogger.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace CSharpier
+{
+    public class ConsoleLogger : ILogger
```
shocklateboy92, line 6: So, console and file log output providers already exist. If you use `ILoggerFactory` (or better yet, [dependency injection](https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection)) to instantiate the loggers, you'll get all this for free.

Since this is not an asp.net web app, you'll need to instantiate them differently, like so:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-5.0#non-host-console-app-1

However, if you chose to use .net's dependency injection framework in this project, it will automatically register all the loggers:
https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#framework-provided-services

shocklateboy92 (commented):
There are more stuff I think you should consider utilizing from asp/asp.net. Might require a bit of thought, but I think they will get you more stuff for significantly less code.

Src/CSharpier/CommandLineFormatter.cs
belav: Unless an IOC container gets passed around I don't see how that would be possible

Src/CSharpier/ConsoleLogger.cs
belav: The standard .net console logger doesn't format things the same as this custom one. The two main things are that it includes the log level on everything, and it includes the location in code where the log is coming from.
```
info: CSharpier.Program[0]
      neiarst
fail: CSharpier.Program[0]
      The .csharpierignore file at C:\Projects\csharpier-repos\.csharpierignore could not be parsed due to the following line:
      \Src\Uploads\*.cs
```
I also couldn't get the logger to log anything unless I logged something immediately after setting up the IOC container.

Using IOC just to get a logger seems a little silly. If you look at dotnet format (an older version because I don't want to spend any time getting the .net 6 solution working)
This is all the code they use to get a single instance of a logger which is then passed around.
```c#
        // method in Program.cs
        private static ILogger<Program> SetupLogging(IConsole console, LogLevel logLevel)
        {
            var serviceCollection = new ServiceCollection();
            serviceCollection.AddSingleton(new LoggerFactory().AddSimpleConsole(console, logLevel));
            serviceCollection.AddLogging();

            var serviceProvider = serviceCollection.BuildServiceProvider();
            var logger = serviceProvider.GetService<ILogger<Program>>();

            return logger;
        }
        
    internal static class SimpleConsoleLoggerFactoryExtensions
    {
        public static ILoggerFactory AddSimpleConsole(this ILoggerFactory factory, IConsole console, LogLevel logLevel)
        {
            factory.AddProvider(new SimpleConsoleLoggerProvider(console, logLevel));
            return factory;
        }
    }

    internal class SimpleConsoleLoggerProvider : ILoggerProvider
    {
        private readonly IConsole _console;
        private readonly LogLevel _logLevel;

        public SimpleConsoleLoggerProvider(IConsole console, LogLevel logLevel)
        {
            _console = console;
            _logLevel = logLevel;
        }

        public ILogger CreateLogger(string name)
        {
            return new SimpleConsoleLogger(_console, _logLevel);
        }

        public void Dispose()
        {
        }
    }
```
Unless I am missing something, that is a whole lot of code that could just be replaced by
```c#
        private static ILogger<Program> SetupLogging(IConsole console, LogLevel logLevel)
        {
            return new SimpleConsoleLogger(_console, _logLevel);
        }

shocklateboy92: Okay, I see your point. It doesn't make sense to use an IOC container.
Maybe as this grows in complexity that might change, but for now let's go ahead with this approach.

belav: Yeah with almost all the code being static methods that just pass around SyntaxNodes and return Doc, an IOC container doesn't add much. I kept thinking that eventually we would run into something that would require us to move away from static methods but it hasn't happened yet.

PR #427, opened 2021-08-29
More uses for source generators

belav:
This ditches the rest of the "generator" unit tests and converts them to source generators.
It makes use of scriban for the generators that were easy to convert to it. The last two would take a lot of effort to convert.


shocklateboy92 (approved):
LGTM.

PR #437, opened 2021-09-18
Adding doc on integrating with dotnet format

belav:
Also ran dotnet format without formatting rules

### DIFF
--- a/Src/CSharpier.Tests/DocPrinterTests.cs
+++ b/Src/CSharpier.Tests/DocPrinterTests.cs
@@ -7,7 +7,7 @@ namespace CSharpier.Tests
 {
     public class DocPrinterTests
     {
-        private static string NewLine = System.Environment.NewLine;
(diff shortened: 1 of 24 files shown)


Src/CSharpier.Tests/DocPrinterTests.cs
```diff
@@ -7,7 +7,7 @@ namespace CSharpier.Tests
 {
     public class DocPrinterTests
     {
-        private static string NewLine = System.Environment.NewLine;
+        private static readonly string NewLine = System.Environment.NewLine;
```
shocklateboy92, line 10: instead of `static readonly`, why can't you just do `const`, since they are strings?

belav: `Environment.NewLine` isn't a constant value, it would vary between windows and linux

PR #445, opened 2021-09-27
Cleaning up the doc tree

belav:
This PR cleans up the doc that we generate, which helps when trying to understand how code is being formatted. It probably also has some minor performance gains.


Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/Argument.cs
```diff
@@ -14,7 +14,11 @@ public static Doc Print(ArgumentSyntax node)
                 docs.Add(NameColon.Print(node.NameColon));
             }
 
-            docs.Add(Token.PrintWithSuffix(node.RefKindKeyword, " "));
+            if (node.RefKindKeyword.RawKind != 0)
```
shocklateboy92, line 17: Is there a constant to compare to?
Or maybe a comment explaining the significance of `0` in `RawKind` would be helpful.

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/BaseMethodDeclaration.cs
```diff
@@ -74,14 +74,14 @@ public static Doc Print(CSharpSyntaxNode node)
 
             var docs = new List<Doc> { ExtraNewLines.Print(node) };
 
-            if (attributeLists.HasValue)
+            if (attributeLists.HasValue && attributeLists.Value.Any())
```
shocklateboy92, line 77: You could do both these checks in a single pattern match, if you think it looks better:
```
attributeLists is { HasValue: true, Value: { Count: > 0 } }
```

Src/CSharpier/SyntaxPrinter/ArgumentListLikeSyntax.cs
```diff
@@ -11,19 +13,23 @@ public static class ArgumentListLike
...
-                      )
-                    : Doc.Null,
-                Token.Print(closeParenToken)
-            );
+        ) {
+            var docs = new List<Doc> { Token.Print(openParenToken) };
+
+            if (arguments.Any())
+            {
+                docs.Add(
```
shocklateboy92, line 21: Why is this change needed?
I'm not saying I prefer the old code over this, but I would like to understand why you're doing this.

Is there some penalty for having `Doc.Null`? If yes, wouldn't it be easier to filter them out in `Doc.Concat()`, causing them to be removed everywhere?

shocklateboy92 (approved):

PR #456, opened 2021-10-16
Misc updates to the playground

belav:
switch back to CRA
run playground with single command
get playground rendering better at smaller screen sizes
add csharpier version to playground


shocklateboy92:
Switching back to CRA is definitely good, but how do you feel about running `dotnet watch run` and `npm run start` separately?

I realized that none of the stuff I've done for csharpier has involved the playground, so I don't have a super strong opinion on it.
However, it would be nice to do `dotnet build` on the solution root without msbuild trying to invoke npm internally and failing because `npm install` hasn't been run. :shrug:

PR #472, opened 2021-11-01
Split CSharpier so it can be a library and a tool

### LINKED ISSUE #470: Using csharpier with a raw SourceText as source
I am currently writing some Source Generators and I am emitting the source from raw SyntaxTrees through the SyntaxFactory API, would there be any way that I can throw the SourceText at csharpier and get a SourceText/string back? 

I personally would love to have this feature as it would allow for the Source Generator consumer to have way more readable code as the one that NormalizeWhitespace emits on its own.

belav:
There are use cases for wanting to programmaticly format c# code, the main one is with source generators.
This PR splits apart CSharpier into CSharpier.CLI (still published as a dotnet tool named CSharpier) and CSharpier (now published as a library named CSharpier.Core).
Almost everything in CSharpier is now internal.
closes #470


.github/workflows/nuget_publish.yml
```diff
@@ -13,21 +13,21 @@ jobs:
     runs-on: ubuntu-latest
     name: publish nuget
     needs: test
+    env:
+      VERSION_FILE_PATH: CSharpier.Build.props
+      TAG_FORMAT: "*"
+      NUGET_KEY: ${{secrets.NUGET_API_KEY}}
     steps:
       - uses: actions/checkout@v2
       - name: publish csharpier on version change
```
shocklateboy92, line 22: maybe clarify to publish `cshapier core` on version change, as that's what it's now called.
Or maybe put one in brackets, to make easier for someone debugging who hasn't got the mental mapping of `CSharpier.csproj` to `CSharpier.Core`

CSharpier.sln
```diff
@@ -64,6 +65,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CSharpier.Tests.Generators"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CSharpier.Generators", "Src\CSharpier.Generators\CSharpier.Generators.csproj", "{D846F494-57FC-4315-A2A8-357343175D2E}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CSharpier.CLI", "src\CSharpier.CLI\CSharpier.CLI.csproj", "{F2BF106F-BB0A-44AC-8249-91BC53FFC83A}"
```
shocklateboy92, line 68: So, the .net style guide on capitalization says to treat [acronyms longer than 2 letters as regular words](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers).

According to that, this project should be `CSharpier.Cli`.
That is, if you so happen to chose that style guide for this project :wink:

Scripts/BuildPackages.ps1
```diff
@@ -0,0 +1,10 @@
+$repositoryRoot = Join-Path $PSScriptRoot ".."
+
+$projects = @("CSharpier.CLI", "CSharpier", "CSharpier.MSBuild")
```
shocklateboy92, line 3: So, the reason you're doing this instead of `dotnet pack` on the solution is because `CSharpier.MsBuild` can't be in the `.sln`, right?

If so, a comment explaining that might help newcomers.

Scripts/Helpers.ps1
```diff
@@ -1,6 +1,6 @@
 $repositoryRoot = Join-Path $PSScriptRoot ".."
-$csProjectPath = Join-Path $repositoryRoot "Src/CSharpier/CSharpier.csproj"
-$csharpierDllPath = Join-Path $repositoryRoot "Src/CSharpier/bin/release/net5.0/dotnet-csharpier.dll"
+$csProjectPath = Join-Path $repositoryRoot "Src/CSharpier.CLI/CSharpier.CLI.csproj"
```
shocklateboy92, line 2: Remember to update the directory name and this script if you take my suggestion re. `CSharpier.Cli`.

Src/CSharpier.CLI/CSharpier.CLI.csproj
```diff
@@ -0,0 +1,33 @@
...
+        <PackageReference Include="System.IO.Abstractions" Version="13.2.29" />
+        <PackageReference Include="System.IO.Abstractions.TestingHelpers" Version="13.2.29" />
+        <PackageReference Include="System.Text.Encoding.CodePages" Version="5.0.0" />
+        <PackageReference Include="UTF.Unknown" Version="2.3.0" />
+        <PackageReference Include="YamlDotNet" Version="11.1.1" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
+            <_Parameter1>CSharpier.Tests</_Parameter1>
```
shocklateboy92, line 25: Woah... we don't do the `AssemblyInfo.cs` anymore? :open_mouth:

Src/CSharpier.CLI/CommandLineFormatter.cs
```diff
@@ -9,9 +9,9 @@
 using CSharpier.Utilities;
 using Microsoft.Extensions.Logging;
 
-namespace CSharpier
+namespace CSharpier.CLI
```
shocklateboy92, line 12: Okay, seeing the namespace makes me feel a little more strongly about the casing.

Src/CSharpier/CodeFormatter.cs
```diff
@@ -14,12 +14,33 @@ namespace CSharpier
 {
     public class CodeFormatter
     {
-        public CSharpierResult Format(string code, PrinterOptions printerOptions)
+        public static string Format(string code, CodeFormatterOptions? options = null)
```
shocklateboy92, line 17: Why is this necessary? With C#9 having async main, there's really no reason to do `.Result` anymore.

shocklateboy92 (commented):
I'll be real, this review would have been a little nicer if making all the classes `internal` was in a separate PR :zebra:

Src/CSharpier.CLI/CommandLineFormatter.cs
belav: Yeah, I think Cli does look better after making the change

Src/CSharpier.CLI/CSharpier.CLI.csproj
belav: It is autogenerated now, although I did find a cleaner way to do this, because it is pretty gross.

Src/CSharpier/CodeFormatter.cs
belav: If you mean why have both `Format` and `FormatAsync` it seems like a nuget library should supply sync and async versions of a method because there are cases where someone may want to format code but they aren't in an async method. From what I know source generators aren't async.

belav:
> I'll be real, this review would have been a little nicer if making all the classes `internal` was in a separate PR 🦓

I had the same thought when I did my self review. I probably I could have created a first PR to master just with the internal changes, but that would mean not being able to do this work until the other PR was done (I think?).
Or if the change to internal was its own commit, then you could review that commit by itself and just scan through it.

Src/CSharpier/CodeFormatter.cs
shocklateboy92: That is what I meant. I figured if we're just calling `.Result` anyway, we may as well leave that up to the caller to deal with.
That being said, I don't feel too strongly about it so If you think it's a worthwhile addition, let's go ahead with it.

shocklateboy92:
> > I'll be real, this review would have been a little nicer if making all the classes `internal` was in a separate PR zebra
> 
> I had the same thought when I did my self review. I probably I could have created a first PR to master just with the internal changes, but that would mean not being able to do this work until the other PR was done (I think?). Or if the change to internal was its own commit, then you could review that commit by itself and just scan through it.

You can make a PR with the target branch not being master. That way, you can essentially base 1 PR on the other.
Though, in this particular case you probably could have had them be completely independent PRs because the changes weren't really dependent on each other :racing_car:

shocklateboy92 (approved):

PR #475, opened 2021-11-10
Adding support for c# 10 and targeting .net 6

### LINKED ISSUE #448: Support C# 10
The following new syntax needs to be supported and properly formatted.
1. Records struct/class
```c#
public readonly record struct Point(double X, double Y, double Z);
public readonly record class Point(double X, double Y, double Z);
```
2. File-scoped namespace declaration
```c#
namespace MyNamespace;
```
3. Extended property patterns
```c#
{ Prop1.Prop2: pattern }
// vs having to be
{ Prop1: { Prop2: pattern } }
```
4. Global using directives
```c#
global using System;
```

The following may just work without any new formatting

1. [Allow AsyncMethodBuilder attribute on methods](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-10#allow-asyncmethodbuilder-attribute-on-methods)

The following work without any new formatting
1. Struct - [Parameterless constructors and field initializers](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#parameterless-constructors-and-field-initializers)
2. A left-hand operand of the [with expression](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/with-expression) can be of any structure type.
3. [Constant interpolated strings](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-10#constant-interpolated-strings)
4. [Record types can seal ToString](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-10#assignment-and-declaration-in-same-deconstruction)
5. [Assignment and declaration in same deconstruction](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-10#assignment-and-declaration-in-same-deconstruction)

belav:
closes #448


Src/CSharpier.Playground/CSharpier.Playground.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
     <PropertyGroup>
-        <TargetFramework>net5.0</TargetFramework>
+        <TargetFramework>net6.0</TargetFramework>
```
belav, line 4: I still have to update the dockerfile for the playground to work with .net6

PR #477, opened 2021-11-11
Ensure files end with a single new line

### LINKED ISSUE #464: File with multiple newlines at the end keeps them
```c#
public class ClassName { }




```
should be
```c#
public class ClassName { }

```

belav:
closes #464

### DIFF
--- a/Src/CSharpier/DocPrinter/DocPrinter.cs
+++ b/Src/CSharpier/DocPrinter/DocPrinter.cs
@@ -42,10 +42,7 @@ namespace CSharpier.DocPrinter
                 ProcessNextCommand();
             }
 
-            if (Output.Length == 0 || Output[^1] != '\n')
-            {
-                Output.Append(EndOfLine);
-            }
+            EnsureOutputEndsWithSingleNewLine();
 
             var result = Output.ToString();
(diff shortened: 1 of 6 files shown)


Src/CSharpier/DocPrinter/DocPrinter.cs
```diff
@@ -56,6 +53,21 @@ public string Print()
             return result;
         }
 
+        private void EnsureOutputEndsWithSingleNewLine()
+        {
+            var trimmed = 0;
+            for (; trimmed < Output.Length; trimmed++)
+            {
+                if (Output[^(trimmed + 1)] != '\r' && Output[^(trimmed + 1)] != '\n')
```
shocklateboy92, line 61: Does this work for Unix line endings that don't have a carriage return?

shocklateboy92 (changes requested):

Src/CSharpier/DocPrinter/DocPrinter.cs
belav: It does, I have some unit tests for it in DocPrinterTests. I did have a copy paste error and wasn't actually testing just "\n" though.

PR #482, opened 2021-11-15
Fixing MSbuild project issue

belav:


Src/CSharpier.MsBuild/CSharpier.MsBuild.csproj
```diff
@@ -3,7 +3,7 @@
     <PropertyGroup>
         <TargetFrameworks>net5.0;netstandard2.0</TargetFrameworks>
         <PackageId>CSharpier.MsBuild</PackageId>
-        <CSharpierOutputDir>../CSharpier.CLI/bin/$(Configuration)/net5.0</CSharpierOutputDir>
+        <CSharpierOutputDir>../CSharpier.Cli/bin/$(Configuration)/net5.0</CSharpierOutputDir>
```
shocklateboy92, line 6: on an unrelated note: we should probably stop hardcoding the `net5.0`, since we're about to upgrade to net 6 for c# 10.

shocklateboy92 (approved):

Src/CSharpier.MsBuild/CSharpier.MsBuild.csproj
belav: Good call, I created https://github.com/belav/csharpier/issues/483 for this.

PR #516, opened 2021-12-20
Parallel tests

belav:


shocklateboy92:
I thought this was the default. Though, I might be confusing NUnit with XUnit.

belav:
> I thought this was the default. Though, I might be confusing NUnit with XUnit.

It isn't the default in NUnit. Our tests at work fail miserably when run in parallel. I was playing around with trying to get them working in parallel and figured why make csharpier do it. No such thing as too fast.

PR #525, opened 2021-12-21
Keep lines above comments in enums

### LINKED ISSUE #524: Allow empty lines before comments in enums
```c#
    public enum SubCode
    {
        Success = 0,

        // allow the line above
        GeneralFailure = 20,

### PR DESCRIPTION
**Pull request by belav**
closes #524

### CODE REVIEW

File: Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/EnumMemberDeclaration.cs [hunk 1]
```diff
@@ -6,6 +6,17 @@ public static Doc Print(EnumMemberDeclarationSyntax node)
     {
         var docs = new List<Doc>
         {
+            node.GetLeadingTrivia()
+            .Any(
```
shocklateboy92, line 10: On an unrelated note, you reckon there should be an indent here? Since it's the start of a method chain?

PR #530, opened 2021-12-22
Getting the rest of the tests to be fully parallel

belav:
This is one way I thought of to get the last two tests to run fully in parallel. The `fileSystem` field causes issues with the tests running in parallel.
I don't know that I love it, and I don't know if I love the name `TestContext`

Another idea that might be better
```c#
    [Test]
    public void Format_Writes_Failed_To_Compile()
    {
        var fileSystem = new FakeFileSystem(); // wrapper around MockFileSystem
        fileSystem.WhenAFileExists("Invalid.cs", "asdfasfasdf");

        var result = this.Format(fileSystem);

        result.ErrorLines
            .First()
            .Should()
            .Be("Error /Invalid.cs - Failed to compile so was not formatted.");
    }
```

Or another idea
```c#
    [Test]
    public void Format_Writes_Failed_To_Compile()
    {
        var fileSystem = new MockFileSystem();
        WhenAFileExists(fileSystem, "Invalid.cs", "asdfasfasdf");

        var result = this.Format(fileSystem);

        result.ErrorLines
            .First()
            .Should()
            .Be("Error /Invalid.cs - Failed to compile so was not formatted.");
    }
```

If this test had 6 fields to get rid of then maybe the `TestContext` idea makes more sense. Newing up 6 objects in each test would be obnoxious.


Src/CSharpier.Tests/CommandLineFormatterTests.cs
```diff
@@ -213,10 +216,11 @@ public void Format_Checks_Formatted_File()
     [TestCase("TestFile.g.i.cs")]
     public void Format_Skips_Generated_Files(string fileName)
     {
+        var context = new TestContext();
         var unformattedFilePath = fileName;
-        WhenAFileExists(unformattedFilePath, UnformattedClassContent);
+        context.WhenAFileExists(unformattedFilePath, UnformattedClassContent);
 
-        var result = this.Format();
+        var result = context.Format();
```
shocklateboy92, line 223: It feels weird to have a the `Format()` on a `context`. Like, it's not just a context anymore.
Maybe you can call it `CommandLineFormatTester`?

Or maybe you can have the `Format()` method be outside that class, and pass the context in as a parameter?

Src/CSharpier.Tests/CommandLineFormatterTests.cs
```diff
@@ -448,66 +465,71 @@ public void Empty_Config_Files_Should_Log_Warning(string configFileName)
...
+        public string WhenAFileExists(string path, string contents)
         {
-            directoryOrFilePaths = new[] { GetRootPath() };
+            path = this.fileSystem.Path.Combine(GetRootPath(), path).Replace('\\', '/');
+            this.fileSystem.AddFile(path, new MockFileData(contents));
+            return path;
         }
-        else
+
+        public string GetRootPath()
```
shocklateboy92, line 479: Maybe `CreateAndGetRootPath()`? It feels really weird for a method called `GetBlah()` to mutate state.

shocklateboy92 (dismissed):
Looks good overall, some comments about naming.

Src/CSharpier.Tests/CommandLineFormatterTests.cs
```diff
@@ -448,66 +465,71 @@ public void Empty_Config_Files_Should_Log_Warning(string configFileName)
...
+                        WriteStdout = writeStdout || standardInFileContents != null,
+                        StandardInFileContents = standardInFileContents,
+                    },
+                    this.fileSystem,
+                    fakeConsole,
+                    testLogger,
+                    CancellationToken.None
+                ).Result;
+
+            fakeConsole.Close();
```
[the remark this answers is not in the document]
belav: It isn't, I cleaned up how it is used a bit.

Src/CSharpier.Tests/CommandLineFormatterTests.cs
belav: I like passing context to a format method. It solves some of the problems with the other potential solutions.

Src/CSharpier.Tests/CommandLineFormatterTests.cs
belav: I switched this around so it no longer mutates state.

PR #535, opened 2022-01-01
Implementing rules for when lines are forced before members

### LINKED ISSUE #285: Require Lines Between Methods?
Csharpier likes getting rid of duplicate lines, but doesn't seem to have an opinion on single line breaks between methods. Both of these examples are unaltered by running csharpier, for example:

```cs
public string Foo() 
{
  return "foo";
}

public string Bar() 
{
  return "bar";
}
```
```cs
public string Foo() 
{
  return "foo";
}
public string Bar() 
{
  return "bar";
}
```
Personally the first one is more appealing to me, but either way, an opinion would be useful.

Are there any plans to implement something like this? I assume there a whole mess of edge cases I'm not considering here.

belav:
closes #285


belav:
This is basically done, there are just a lot of edge cases related to preprocessor directives to figure out. I also wanted to see if anyone had any feedback on the proposed rules for when blank lines are forced.

shocklateboy92 (commented):
Oaf. This is long and complicated. I'm just going to trust that you got it right :stuck_out_tongue_closed_eyes:

belav:
@shocklateboy92 it was actually pretty simple until I started testing on real world code and ran into all the edge cases with preprocessor statements. I am pretty confident the code works. But there is some of it that I am sure could probably be simplified.

PR #540, opened 2022-01-03
Formatting with 13

belav:


Src/CSharpier/DocPrinter/DocFitter.cs
```diff
@@ -23,7 +23,7 @@ void Push(Doc doc, PrintMode printMode, Indent indent)
 
         var output = new StringBuilder();
 
-        for (var x = 0; x < remainingCommands.Count || newCommands.Count > 0;)
+        for (var x = 0; x < remainingCommands.Count || newCommands.Count > 0; )
```
shocklateboy92, line 26: Is this a bug?

shocklateboy92 (approved):

Src/CSharpier/DocPrinter/DocFitter.cs
belav: No, I made the change to get csharpier to abide by the stylecop rule for always having a space after a semicolon. Prettier formats for's without iterators the same way.

PR #544, opened 2022-01-11
Implementing husky.net + documenting it.

### LINKED ISSUE #537: Document running csharpier with husky.net
### Discussed in https://github.com/belav/csharpier/discussions/536

<div type='discussions-op-text'>

<sup>Originally posted by **alirezanet** January  2, 2022</sup>
Hi, 
I think prettier is popular because of other tools like husky and lint-staged that can automate the prettier formatting process. We have a better alternative to those for dotnet ( [Husky.Net](https://github.com/alirezanet/Husky.Net) ). I suggest that you add instructions for csharpier using Huksy.net to the documentations.</div>

belav:
Also cleaning up the solution a bit with notarget projects.
closes #537

### DIFF
--- a/Scripts/Scripts.csproj
+++ b/Scripts/Scripts.csproj
@@ -1,0 +1,1 @@
+<Project Sdk="Microsoft.Build.NoTargets/3.1.0" />
--- a/Src/CSharpier/CSharpier.csproj
+++ b/Src/CSharpier/CSharpier.csproj
@@ -28,4 +28,8 @@
         <ProjectReference Include="..\CSharpier.Generators\CSharpier.Generators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
     </ItemGroup>
 
+    <Target Name="husky" BeforeTargets="Restore;CollectPackageReferences" Condition="'$(HUSKY)' != 0">
+        <Exec Command="dotnet tool restore"  StandardOutputImportance="Low" StandardErrorImportance="High"/>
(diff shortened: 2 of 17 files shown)


Src/CSharpier/CSharpier.csproj
```diff
@@ -28,4 +28,8 @@
         <ProjectReference Include="..\CSharpier.Generators\CSharpier.Generators.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
     </ItemGroup>
 
+    <Target Name="husky" BeforeTargets="Restore;CollectPackageReferences" Condition="'$(HUSKY)' != 0">
```
shocklateboy92, line 31: Didn't you say you were going to make these opt-in for this repo?

Scripts/Scripts.csproj
```diff
@@ -0,0 +1 @@
+<Project Sdk="Microsoft.Build.NoTargets/3.1.0" />
```
shocklateboy92, line 1: Isn't it a bit overkill to have all these different projects?
It may not be the case anymore but visual studio's performance used to degrade quite drastically based on the amount of projects loaded.

Src/CSharpier/CSharpier.csproj
belav: I also updated the husky json file because it was not obvious the first part of it was a comment.

Scripts/Scripts.csproj
belav: I work with another team who uses projects this way. It seems like a decent alternative to managing solution folders. I am pretty sure either VS2022, or 2019 claimed to fix the performance issues with large numbers of projects.

PR #550, opened 2022-01-18
Fixing issue with unicode characters when piping standard in.

### LINKED ISSUE #545: Encoding During Format On Save
Firstly, thank you so much for this project.

I am part of an international team and when formatting utf8 files with unicode chars the chars get replaced with "?".
In our .editorconfig  we have charset=utf-8
Then in our cs file we have this line:

```csharp

private const char test = '。';

```

After saving the file it becomes:

```csharp

private const char test = '?';

```

For now I can use '\u' but sometimes we have Japanese text in files so we do not want those to break.

Sorry I left out a lot of needed info...

Using VS2022, global install + extension.
Using command "dotnet csharpier ." doesn't cause this issue.
"Reformat with CSHarpier on Save" = true causes the issue on save.

Thanks.

belav:
Closes #545 eventually, seeing if the CLI test does the same thing on github


belav:
Switching over to reading/writing with utf8 in the csharpier.cli may be a breaking change. I can probably figure that out by testing the existing extensions with the new csharpier. If it is a breaking change I may need to add a `-use-utf8` argument. Or since csharpier is still beta, it could be a documented breaking change.

PR #560, opened 2022-01-20
If file has no preprocessor symbols, only format it once

### LINKED ISSUE #555: File with no preprocessor symbols formats twice.
See CodeFormatter line 122 or so, if that returns an empty array, we should not format again.
Or it should not return anything if there are no symbol sets.

belav:
closes #555

### DIFF
--- a/Src/CSharpier.Cli.Tests/CliTests.cs
+++ b/Src/CSharpier.Cli.Tests/CliTests.cs
@@ -38,7 +38,7 @@ public class CliTests
         var formattedContent = "public class ClassName { }" + lineEnding;
         var unformattedContent = $"public class ClassName {{{lineEnding}{lineEnding}}}";
 
-        await WriteFileAsync("BasicFile.cs", unformattedContent);
+        await this.WriteFileAsync("BasicFile.cs", unformattedContent);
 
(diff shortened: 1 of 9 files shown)


Src/CSharpier.Cli.Tests/CliTests.cs
```diff
@@ -38,7 +38,7 @@ public async Task Should_Format_Basic_File(string lineEnding)
         var formattedContent = "public class ClassName { }" + lineEnding;
         var unformattedContent = $"public class ClassName {{{lineEnding}{lineEnding}}}";
 
-        await WriteFileAsync("BasicFile.cs", unformattedContent);
+        await this.WriteFileAsync("BasicFile.cs", unformattedContent);
```
shocklateboy92, line 41: If you want to always prepend `this.`, that's fine, but could you change all of them in a single PR?
Littering the changes all over the place makes it hard to review.

shocklateboy92 (approved):

Src/CSharpier.Cli.Tests/CliTests.cs
belav: Yeah sorry, I went ahead and merged it - https://github.com/belav/csharpier/pull/562

Issue #549, opened 2022-01-18
Rider plugin stack trace: AssertionError: Wrong line separators


commonquail:
I sometimes get this when I "open recent" a solution.

Plugin 1.0.3, CSharpier 0.11.1.

```
java.lang.AssertionError: Wrong line separators: '\r\nWelcome ...' at offset 0
	at com.intellij.openapi.util.text.StringUtil.assertValidSeparators(StringUtil.java:2475)
	at com.intellij.openapi.editor.impl.DocumentImpl.assertValidSeparators(DocumentImpl.java:714)
	at com.intellij.openapi.editor.impl.DocumentImpl.replaceString(DocumentImpl.java:608)
	at com.intellij.openapi.editor.impl.DocumentImpl.replaceString(DocumentImpl.java:599)
	at com.intellij.csharpier.FormattingService.lambda$format$0(FormattingService.java:55)
	at com.intellij.openapi.command.WriteCommandAction.lambda$runWriteCommandAction$4(WriteCommandAction.java:361)
	at com.intellij.openapi.command.WriteCommandAction$BuilderImpl.lambda$doRunWriteCommandAction$1(WriteCommandAction.java:150)
	at com.intellij.openapi.application.impl.ApplicationImpl.runWriteAction(ApplicationImpl.java:947)
	at com.intellij.openapi.command.WriteCommandAction$BuilderImpl.lambda$doRunWriteCommandAction$2(WriteCommandAction.java:148)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:219)
	at com.intellij.openapi.command.impl.CoreCommandProcessor.executeCommand(CoreCommandProcessor.java:184)
	at com.intellij.openapi.command.WriteCommandAction$BuilderImpl.doRunWriteCommandAction(WriteCommandAction.java:157)
	at com.intellij.openapi.command.WriteCommandAction$BuilderImpl.run(WriteCommandAction.java:124)
	at com.intellij.openapi.command.WriteCommandAction.runWriteCommandAction(WriteCommandAction.java:361)
	at com.intellij.openapi.command.WriteCommandAction.runWriteCommandAction(WriteCommandAction.java:349)
	at com.intellij.csharpier.FormattingService.format(FormattingService.java:54)
	at com.intellij.csharpier.ReformatWithCSharpierOnSave.beforeDocumentSaving(ReformatWithCSharpierOnSave.java:35)
[...]

Issue #531, opened 2021-12-26
Extensions do not detect when csharpier is installed or upgraded and in some cases prevent upgrading.


belav:
This probably applies to VSCode, VS and Rider.
The extensions only determine the version of csharpier when they start. If someone does not have csharpier installed, then installs it. They need to restart the IDE.
If 0.12.0+ is being used, then the extension keeps csharpier running. Which prevents someone from updating it without closing the IDE.

-  Modify the keep csharpier running code
    - If someone is not active in a document, the csharpier process should be killed
    - When someone becomes active in a document, determine the csharpier version if it is 12+ warm up the process

This may be affected by the logic for using the solution version of csharpier instead of a global version. #493

belav:
#563 does a better job of ensuring the IDE does not keep the process running. But we still want to either determine when csharpier was updated and reload the processes in the IDE, or give the user the ability to reload the processes.
VSCode has the timeout implemented, but it should come back out.
Another way to detect when the version changes is to watch the tool manifest file and the global install location.
Or to recheck we are using the versions on occasion. Without killing the running versions.

belav:
#569 covers this

PR #566, opened 2022-01-23
Updates to the rider plugin

### LINKED ISSUE #493: Extensions do not take into account solution version of csharpier
Currently the vscode extension runs `dotnet cshapier` outside of the context of where the file lives. This could result in it not formatting code the way it should

1. Install version 0.11.1 globally
2. Install version 0.12.0 in a solution
3. Formatting a file from that solution in VSCode should use the 0.12.0 version of csharpier.

When we initialize the vscode extension, we do it outside of any kind of file/folder context.
We need some way to figure out what version of csharpier a file should be formatted with, and load up a new csharpier process if the version does not match.

### LINKED ISSUE #533: Extensions - there are times when dotnet csharpier --version returns more than the version
I had an instance of testing csharpier with VS2022 where the output contained a "welcome to .net 6" sort of message and ended with the version. This caused the extension to not figure out the csharpier version.
Working around it would just require reading the last line of the output.
This may also be affected by #531

### LINKED ISSUE #563: Extensions - for long running csharpier, use a custom tools path
Currently when the extensions keep csharpier running it prevents a user from uninstalling/updating csharpier. They get the generic error 
```bash
Failed to uninstall tool package 'csharpier': Access to the path 'C:\Users\bela\.dotnet\tools\.store\csharpier\0.13.0' is denied.
```
The VSCode extension has code in place that shuts down the csharpier process and restarts it when it is needed. But this sometimes leads to a slow format of a file. It also causes a lot of unnecessary warming of csharpier by formatting files behind the scenes.

A better approach would be to install csharpier in a custom location - https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools#install-a-global-tool-in-a-custom-location

- After determining which version of csharpier should be used, check if that version of csharpier is installed in the custom location. Install it if needed.
- Always run csharpier from the custom location
- Try to determine when someone upgrades csharpier, or provide them an action to reload csharpier. This will handle the case of someone updating csharpier while the IDE is running and using the newer version.

If this approach works, then #557, #558 are not required.

belav:
These fixes were all touching the same code, so it made sense to do them all at once.

Use DOTNET_NOLOGO - fixes #533 and #549
Use a custom tools path to allow csharpier to be updated/installed while running - fixes #563 and #531
Use the solution version of csharpier - fixes #493


belav:
I still need to double check all of the test cases.

belav:
#567 was was going to conflict with this unless I did them together.

Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessPipeMultipleFiles.java
```diff
@@ -21,8 +21,9 @@
     public CSharpierProcessPipeMultipleFiles(String csharpierPath, boolean useUtf8) {
         this.csharpierPath = csharpierPath;
         try {
-            this.process = new ProcessBuilder("dotnet", csharpierPath, "--pipe-multiple-files")
-                    .start();
+            ProcessBuilder processBuilder = new ProcessBuilder(csharpierPath, "--pipe-multiple-files");
```
shocklateboy92, line 24: I think you can use `var` as of Java 10.

Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CustomPathInstaller.java
```diff
@@ -0,0 +1,37 @@
...
+            return;
+        }
+
+        String[] command = {"dotnet", "tool", "install", "csharpier", "--version", version, "--tool-path", directoryForVersion};
+        ProcessHelper.ExecuteCommand(command, null, null);
+    }
+
+    private String getDirectoryForVersion(String version) {
+        String userHome = System.getProperty("user.home");
+        Path path = Path.of(userHome, ".csharpier", version);
```
shocklateboy92, line 28: Can we put it in `~/.cache/csharpier` instead of contributing to the littering of people's home directories?
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

shocklateboy92 (approved):
Took a quick skim through it. I might time to go through it in more detail, but that will be a week from now, if at all.

Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CustomPathInstaller.java
commonquail: +1, though on Windows you'd probably want a `System.getenv("LOCALAPPDATA")` prefix.

Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessPipeMultipleFiles.java
belav: You sir, are a life saver! I will convert everything to vars in another PR. Weird that intellij wasn't suggesting this to me.

Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CustomPathInstaller.java
belav: Both good ideas, done.

belav:
I put up a beta of this at https://github.com/belav/csharpier/releases

belav:
I believe I ran into this too and fixed it right after I created the release. The plugin was trying to find csharpier for every open window when it initialized, which included things like a welcome message.
I have a new beta that resolves it, and cleans up the thread handling #572 
https://github.com/belav/csharpier/releases/tag/rider-1.2.0-beta2

PR #572, opened 2022-01-25
Better way of dealing of threads 

### LINKED ISSUE #568: Rider - better way to deal with threads for reading out/error with long running process.
See the comment in CSharpierProcessPipeMultipleFiles.

belav:
I believe this is a better way of dealing with the threads that read the output from the long running csharpier process.

closes #568


shocklateboy92 (approved):
LGTM. I don't know enough about the Java standard libraries to give meaningful critique of thread handling though.

belav:
I dug through some other plugins + documentation, and this seemed to be the proper way to deal with them. I have been using the beta with this thread handling for a couple days and it hasn't caused any issues.

PR #582, opened 2022-01-30
Better formatting of multiline comments when indentation is changing

### LINKED ISSUE #580: Multiline comments are not properly indented.
I was hoping to use csharpier on a code base that currently uses 2 space indents. One thing I noticed right away is that the indentation of multiline comments is not handled correctly. Only the first line in indented and the rest of the lines remain where they were. This applies to both xml doc comments and C-style comments.

Actual:

image

Should be:

image

Actual:

image

Should be:

image

As a workaround, I used `dotnet format` first, then `dotnet csharpier`.

belav:
closes #580


Src/CSharpier/SyntaxPrinter/Token.cs
```diff
@@ -135,26 +135,38 @@ public static Doc PrintLeadingTriviaWithNewLines(SyntaxTriviaList leadingTrivia)
             {
                 printNewLines = true;
             }
-            if (IsSingleLineComment(kind))
+
+            void AddLeadingCommend(CommentType commentType)
```
shocklateboy92, line 139: I think you meant `AddLeadingComment`

PR #584, opened 2022-02-01
Implement csharpier-ignore for some node types

### LINKED ISSUE #581: // csharpier-ignore
There are some cases where initializers have super-long lists of arguments. It would be helpful to have the equivalent of [// prettier-ignore](https://prettier.io/docs/en/ignore.html#javascript) to handle these cases.

belav:
closes #581


shocklateboy92 (approved):
You know, when I glanced over the issue I thought you were just going to make `// csharpier-ignore` just ignore the rest of the file.

This is way cooler, but remember file scope is acceptable behavior if you find yourself losing too much hair trying to maintain this in future :smile:

PR #585, opened 2022-02-01
Remove extra lines before closing brace

### LINKED ISSUE #575: Extra blank lines should be removed at the end of a method
```c#
public class ClassName
{
    public void MethodName()
    {
        var x = 1;





    }
}

### PR DESCRIPTION
**Pull request by belav**
closes #575

### CODE REVIEW

File: Src/CSharpier/SyntaxPrinter/Token.cs [hunk 1]
```diff
@@ -113,27 +114,34 @@ public static Doc PrintLeadingTriviaWithNewLines(SyntaxTriviaList leadingTrivia)
     private static Doc PrivatePrintLeadingTrivia(
         SyntaxTriviaList leadingTrivia,
         bool includeInitialNewLines = false,
-        bool skipLastHardline = false
+        bool skipLastHardline = false,
+        bool useLastWasLine = false
```
shocklateboy92, line 118: This is kind of a terrible name. What does it use `lastWasLine` for?
To skip consecutive lines? If so, name it as such.

Src/CSharpier/SyntaxPrinter/Token.cs
```diff
@@ -113,27 +114,34 @@ public static Doc PrintLeadingTriviaWithNewLines(SyntaxTriviaList leadingTrivia)
...
     )
     {
         var docs = new List<Doc>();
 
         // we don't print any new lines until we run into a comment or directive
         // the PrintExtraNewLines method takes care of printing the initial new lines for a given node
         var printNewLines = includeInitialNewLines;
-
+        var lastWasLine = false;
```
shocklateboy92, line 126: maybe call this `lastTokenWasLine`?

Src/CSharpier/SyntaxPrinter/Token.cs
belav: I realized it was basically the equivalent of !printNewLines and figured out how to get rid of it completely. Even with a better name the code was still ugly.

shocklateboy92 (approved):

PR #586, opened 2022-02-02
Getting VSCode in sync for 1.2.0

### LINKED ISSUE #563: Extensions - for long running csharpier, use a custom tools path

belav:
Use DOTNET_NOLOGO #533
Use a custom tools path to allow csharpier to be updated/installed while running - fixes #563


Src/CSharpier.VSCode/package.json
```diff
@@ -28,7 +28,7 @@
         "Formatters"
     ],
     "activationEvents": [
-        "onStartupFinished"
+        "onLanguage:csharp"
```
shocklateboy92, line 31: :+1:

Src/CSharpier.VSCode/src/CustomPathInstaller.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (!version) {
+            return;
+        }
+        const directoryForVersion = this.getDirectoryForVersion(version);
+        if (fs.existsSync(directoryForVersion)) {
+            this.logger.debug("File at " + directoryForVersion + " already exists");
+            return;
+        }
+
+        const command = "dotnet tool install csharpier --version " + version + " --tool-path";
```
shocklateboy92, line 23: Uhm... shouldn't you have the path appended to this command after `--tool-path`?

Src/CSharpier.VSCode/src/CSharpierProcessProvider.ts
```diff
@@ -218,9 +166,7 @@ export class CSharpierProcessProvider implements Disposable {
 
     private killRunningProcesses = () => {
         for (const key in this.csharpierProcessesByVersion) {
-            this.logger.debug(
-                "disposing of process for version " + (key === "" ? "null" : key),
-            );
+            this.logger.debug("disposing of process for version " + (key === "" ? "null" : key));
```
shocklateboy92, line 169: nit: you can shorten this to `key || "null"`.

Src/CSharpier.VSCode/src/CSharpierProcessPipeMultipleFiles.ts
```diff
@@ -20,9 +20,10 @@ export class CSharpierProcessPipeMultipleFiles implements ICSharpierProcess {
     }
 
     private spawnProcess = (csharpierPath: string, workingDirectory: string) => {
-        const csharpierProcess = spawn("dotnet", [csharpierPath, "--pipe-multiple-files"], {
+        const csharpierProcess = spawn(csharpierPath, ["--pipe-multiple-files"], {
```
shocklateboy92, line 23: Uh... you sure about this?
What does `csharpierPath` evaluate to? a `.exe`? a `.dll`? I don't think dotnet produces standalone executables for any platform other than windows.

Src/CSharpier.VSCode/src/CSharpierProcessSingleFile.ts
```diff
@@ -15,9 +15,10 @@ export class CSharpierProcessSingleFile implements ICSharpierProcess {
     formatFile(content: string, filePath: string): Promise<string> {
         const directory = path.parse(filePath).dir;
         return new Promise((resolve, reject) => {
-            const csharpier = spawn("dotnet", [this.csharpierPath, "--write-stdout"], {
+            const csharpier = spawn(this.csharpierPath, ["--write-stdout"], {
```
shocklateboy92, line 18: Same comment as the other place

Src/CSharpier.VSCode/src/CustomPathInstaller.ts
belav: er, yeah. This probably worked with my testing because I forgot to delete everything from the custom path.

Src/CSharpier.VSCode/src/CSharpierProcessPipeMultipleFiles.ts
belav: It evaluates to "[path]/dotnet-csharpier", which works on windows because including the .exe is optional. I tested it out on ubuntu with when I did the rider plugin and just doublechecked it now for vscode in ubunto.

shocklateboy92 (dismissed):
LGTM. 
But I'm not going to merge this because the version in the manifest file still ends in `-beta2`.
I'll leave you to remove it before merging. :cake:

PR #589, opened 2022-02-05
Better breaking for generic object creation expressions with no arguments or initializer

### LINKED ISSUE #578: Generics + ObjectCreationExpression should break consistently
```c#
namespace CSharpier.VisualStudio
{
    public class CSharpierProcessProvider
    {
        // should look like the one below
        private readonly Dictionary<string, string> csharpierVersionByDirectory = new Dictionary<
            string,
            string
        >();
        private readonly Dictionary<string, ICSharpierProcess> csharpierProcessesByVersion =
            new Dictionary<string, ICSharpierProcess>();

### PR DESCRIPTION
**Pull request by belav**
closes #578

### CODE REVIEW

**Comment by belav** [comment]
One potential downside to this
```c#
// before this change
            _projectFixersMap = new ConditionalWeakTable<
                IReadOnlyList<AnalyzerReference>,
                ImmutableDictionary<string, List<CodeFixProvider>>
            >();
// after this change
            _projectFixersMap =
                new ConditionalWeakTable<
                    IReadOnlyList<AnalyzerReference>,
                    ImmutableDictionary<string, List<CodeFixProvider>>
                >();
```

We may want to break after the `=` if the object creation can fit onto a single line, but I don't know how to accomplish that, and you could agree the "after" is more readable.

https://github.com/belav/csharpier-repos/pull/27/files shows what this does to real world code.

shocklateboy92 (approved):
LGTM. Definitely better to break on the `=` first, so I'm okay with the "downside".

PR #592, opened 2022-02-06
Fixing random failure

belav:
This fixes an occasional exception that was being thrown when trying to update the action's visibility in the right click menu


Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/ReformatWithCSharpierAction.java
```diff
@@ -27,15 +27,20 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
     @Override
     public void update(@NotNull AnActionEvent e) {
-        var virtualFile = e.getData(PlatformDataKeys.VIRTUAL_FILE).toString();
+        var virtualFile = e.getData(PlatformDataKeys.VIRTUAL_FILE);
+        if (virtualFile == null) {
+            e.getPresentation().setVisible(false);
+            return;
+        }
+        var virtualFileString = virtualFile.toString();
         var filePrefix = "file://";
```
shocklateboy92, line 36: Probably worth extracting this to a constant.
https://stackoverflow.com/a/9639022

shocklateboy92 (approved):

PR #593, opened 2022-02-06
Make sure null conditionals break like regular expressions

### LINKED ISSUE #561: Null conditional operator does not break consistently
The ?. calls should break the same.

```c#
class ClassName
{
    void MethodName()
    {
        this.CallMethod_____________________()
            .CallMethod_____________________()
            .CallMethod_____________________();

        this.CallMethod_____________________()?.CallMethod_____________________()?.CallMethod_____________________();
    }
}

### PR DESCRIPTION
**Pull request by belav**
closes #561

### DIFF
--- a/Src/CSharpier/DocTypes/Doc.cs
+++ b/Src/CSharpier/DocTypes/Doc.cs
@@ -2,6 +2,11 @@ namespace CSharpier.DocTypes;
 
 internal abstract class Doc
 {
+    public override string ToString()
+    {
+        return DocSerializer.Serialize(this);
+    }
+
     public static implicit operator Doc(string value)
     {
         return new StringDoc(value);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by belav** [comment]
I still need to look at the result of this change on csharpier-repos. Creating the testing PR is taking significantly longer than it should. I created #594 to look into it.

**Comment by belav** [comment]
I just noticed that the `!.` has the same issue
```c#
        this.CallMethod_____________________()!.CallMethod_____________________()!.CallMethod_____________________();
```
I will fix that tomorrow.

Src/CSharpier/DocTypes/Doc.cs
```diff
@@ -2,6 +2,11 @@ namespace CSharpier.DocTypes;
 
 internal abstract class Doc
 {
+    public override string ToString()
+    {
+        return DocSerializer.Serialize(this);
```
shocklateboy92, line 7: I'm surprised it took this long for that to happen :grin:

shocklateboy92 (approved):

Src/CSharpier/DocTypes/Doc.cs
belav: So nice for debugging, I don't know why I didn't look into how to do this like 12 months ago.

PR #597, opened 2022-02-07
Drastically speed up formatting when ignore or config file exist.

### LINKED ISSUE #594: .csharpierignore causes csharpier to be significantly slower
Formatting csharpier-repos now takes 60 minutes. 
#581 #580 or #575 are the most likely ones to have caused the problem.

belav:
closes #594

### DIFF
--- a/Src/CSharpier.Cli/CommandLineFormatter.cs
+++ b/Src/CSharpier.Cli/CommandLineFormatter.cs
@@ -17,6 +17,31 @@ internal static class CommandLineFormatter
     {
         try
         {
+            async Task<(IgnoreFile, PrinterOptions)> GetIgnoreFileAndPrinterOptions(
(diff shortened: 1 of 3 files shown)


shocklateboy92 (approved):
Okay, I not sure what the magic bit that caused the massive performance improvement is.
But I don't see any glaring code issues, so LGTM.

belav:
> Okay, I not sure what the magic bit that caused the massive performance improvement is. But I don't see any glaring code issues, so LGTM.

Creating the ignore file is somewhat slow. Before these changes I think it was creating the ignore file once per directory that contained .cs files. Or maybe even once per .cs file that was formatted..
Formatting csharpier-repos with an ignore file took an hour before these changes, vs I think ~7 minutes with these changes or no ignore file.

PR #601, opened 2022-02-08
Handles an edge case from #580

belav:


Src/CSharpier/DocPrinter/DocPrinter.cs
```diff
@@ -187,9 +187,10 @@ private void AppendComment(LeadingComment leadingComment, Indent indent)
             //  *  keeps the * in line
             //  */
             var firstLineIndentLength =
-                firstLine!.Replace("\t", "    ").Length - firstLine.TrimStart().Length;
+                firstLine!.Replace("\t", "    ").Length
```
shocklateboy92, line 190: This is probably worth extracting to a constant. It seems to come up a fair bit.

PR #602, opened 2022-02-08
New conditional access

belav:
This has a lot of edge cases, and is probably not worth doing.

https://github.com/belav/csharpier-repos/pull/31/files


commonquail:
[These operators](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-) are called `?.` respectively `?[]` so I think it's not unreasonable to consider them a single token.

The [null-forgiving operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving), on the other hand, is specifically _just_ `!` (without `.`).

PR #607, opened 2022-02-14
Add `DevelopmentDependency` property to `CSharpier.MsBuild.csproj`

martinothamar:
Hi, awesome project, thanks for your work!

This package only runs in `BeforeTargets="Build"` right? I just tested installing the MsBuild package like this (with PrivateAssets attr), in a global `Directory.Build.Props` file, and it seems to work:
```csproj
  <ItemGroup>
    <PackageReference Include="CSharpier.MsBuild" Version="0.15.0" PrivateAssets="all" /> 
  </ItemGroup>
```

Usecase being we have some packages that are going to be published, and we only need `csharpier` during build (doesnt need to be a dep of the package).

### DIFF
--- a/Src/CSharpier.MsBuild/CSharpier.MsBuild.csproj
+++ b/Src/CSharpier.MsBuild/CSharpier.MsBuild.csproj
@@ -8,6 +8,7 @@
         -->
         <TargetFramework>netstandard2.0</TargetFramework>
         <PackageId>CSharpier.MsBuild</PackageId>
+        <DevelopmentDependency>true</DevelopmentDependency>
         <CSharpierOutputDir>../CSharpier.Cli/bin/$(Configuration)/net6.0</CSharpierOutputDir>
     </PropertyGroup>
 


martinothamar:
Btw, this should also add the `PrivateAssets` attribute to the package ref when users install via nuget.org or CLI, currently it installs as a normal dependency
image

belav:
Thanks for reporting this.

I did find that when CSharpier.MsBuild is a DevelopmentDependency, it ends up in the csproj like this
```
      <PackageReference Include="CSharpier.MsBuild" Version="0.15.1">
        <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
        <PrivateAssets>all</PrivateAssets>
      </PackageReference>
```
[StyleCop.Analyzerz](https://www.nuget.org/packages/StyleCop.Analyzers/) does the same thing.

belav (approved):

shocklateboy92 (commented):
Woah... @belav beat me to a review 😮 Australia time-zone is having some disadvantages.
Thanks for your contribution @martinothamar !

PR #609, opened 2022-02-14
Building out a proper website for csharpier

### LINKED ISSUE #505: Turn CSharpier.com into a proper website.

belav:
This is the new csharpier website. The playground will be moving to a subdomain because getting that working with docusaurus isn't worth the effort.

closes #505


shocklateboy92:
You should probably have a link to this in the project summary so it shows up on the top right.
(I don't know if there's a fancy website field in project settings though)

PR #616, opened 2022-02-19
fix: macos rider

JSanchezIO:
For #610

On Mac, the `dotnet` binary isn't available to the `ProcessBuilder`. I added the binary to the `PATH`. I wanted to be mindful of ARM/x64 so the `PATH` exposes the ARM build first and the x64 secondly. However, the `PATH` variable isn't available to the `ProcessBuilder`, but it is for any spawned processes. With that in mind, I updated the`ProcessHelper` to wrap the `command` argument within the OS's default shell.

**Alternatives**
- Add a `.GetDotNetBinary()` method that returned `/usr/local/share/dotnet/dotnet` on Mac.
    - This worked fine, but it wouldn't support x64 binaries on the M1 Mac, so I decided against it.
- Get the binary in use from Rider 
    - I couldn't figure out how to get this to work, but if we could make it happen, I believe it'd be the best option.

### DIFF
--- a/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/ProcessHelper.java
+++ b/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/ProcessHelper.java
@@ -1,6 +1,8 @@
 package com.intellij.csharpier;
 
+import org.apache.commons.lang.SystemUtils;
 import java.io.*;
+import java.util.HashMap;
 import java.util.Map;
 
 public class ProcessHelper {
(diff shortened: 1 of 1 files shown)


Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/ProcessHelper.java
```diff
@@ -47,4 +57,32 @@ public static String ExecuteCommand(String[] command, Map<String, String> env, F
...
+    }
+
+    // Setting the PATH doesn't affect the ProcessBuilder's command, but it will apply to
+    // spawned processes, so we'll run the desired command in the OS's default shell.
+    private static String[] GetShellCommand(String[] command) {
+        if (SystemUtils.IS_OS_MAC) {
+            return new String[] {"/bin/zsh", "-c", String.join(" ", command)};
+        }
+
+        if (SystemUtils.IS_OS_WINDOWS) {
```
shocklateboy92, line 82: Seems a little overkill to launch a shell for all platforms, when the workaround is only needed for macos.
How about renaming this method to `GetShellCommandIfNeeded()` and make it just return the original command if it's not currently MacOS.

shocklateboy92 (changes requested):
Hey, Thanks for your contribution!
I don't have a macbook, so I'll trust you that this is the best workaround for the issue.

However, could you make the changes a little more scoped to macos?

Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/ProcessHelper.java
JSanchezIO: Sure thing. I was aiming for normalizing the way we run commands but totally understand your perspective. I'll make that change and push it up shortly.

shocklateboy92 (approved):
Cheers!

PR #619, opened 2022-02-21
Add support for netstandard 2.0 to CSharpier.Core

belav:
This shims in the required code so that CSharpier.Core can be built targeting netstandard2.0


Src/CSharpier/CSharpier.csproj
```diff
@@ -2,7 +2,7 @@
     <Import Project="../../CSharpier.Build.props" />
     <PropertyGroup>
         <PackageId>CSharpier.Core</PackageId>
-        <TargetFrameworks>net5.0;net6.0</TargetFrameworks>
+        <TargetFrameworks>net5.0;net6.0;netstandard2.0</TargetFrameworks>
```
shocklateboy92, line 5: Is there any particular reason this has to be `netstandard2.0`?
IIRC `netstandard2.1` includes the `Index` type and you won't have to do this shim.

PR #623, opened 2022-02-26
Consistently indent by 4 spaces (or by tabs)

### LINKED ISSUE #617: Consistently Indent By 4 Spaces
#346 doesn't work great with EditorConfig. For C#, one typically sets `indent_size = 4`. When using editorconfig-checker however, one has to set `indent_size = 2` to avoid throwing errors on nested ternary statements, for example. That then causes the editor to tab 2 spaces rather than 4, largely defeating the purpose of specifying an `indent_size` in the first place. It seems simpler to expect folks who want less indentation not to put an inordinate amount of code in a single statement than to use 4-space indentation 99% of the time and 2-space indentation 1% of the time. I'm positive there are other tooling scenarios that are affected by the 1%.

belav:
closes #617


Src/CSharpier/SyntaxPrinter/ConstraintClauses.cs
```diff
@@ -32,11 +16,6 @@ public static Doc Print(IEnumerable<TypeParameterConstraintClauseSyntax> constra
             constraintClausesList.Select(TypeParameterConstraintClause.Print)
         );
 
-        return Doc.Group(
-            Doc.Indent(groupId != null ? Doc.IfBreak(" ", prefix, groupId) : prefix),
-            groupId != null
-              ? Doc.IfBreak(Doc.Align(2, body), Doc.Indent(body), groupId)
```
shocklateboy92, line 38: Heh, funny we never extracted the `2` to a constant. I was kind of hoping this change (on the code side) would be a simple `HalfIndent = DefaultIndent` :smirk:

Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/BinaryExpression.cs
```diff
@@ -24,8 +24,11 @@ or SwitchExpressionSyntax
                     or SwitchStatementSyntax
                     or WhereClauseSyntax
                     or WhileStatementSyntax
-            || node.Parent is ConditionalExpressionSyntax
-                && node.Parent.Parent is not ArgumentSyntax;
+            || (
+                node.Parent is ConditionalExpressionSyntax conditionalExpressionSyntax
```
shocklateboy92, line 28: I think you could have done some fancy pattern matching shenanigans like:
```
node.Parent is ConditionalExpressionSyntax
{ 
    WhenTrue: not node,
    WhenFalse: not node
}
```

Though, I don't know if it will allow `node` as the target or if it needs to be a literal value.

shocklateboy92 (approved):

PR #634, opened 2022-03-18
fix: ignore file detection when directory contains period

JSanchezIO:
I was using CSharpier in a project whose directory name contained a period, e.g., `/home/jsanchezio/Documents/Repos/My.Project/`, and despite having a `.csharpierignore` file, it would still report files excluded by the ignore as unformatted. I was running the command as `dotnet csharpier --check .` After some investigation, it turns out `fileSystem.Path.GetExtension(directoryOrFile)` would return the path after the period, e.g., `.Project`. This change alters the command to instead check if the `directoryOrFile` is a directory that exists and reacts accordingly.

After I discovered the issue, I renamed my project's folder name and all worked well. However, in CI environments there is no guarantee that a `.` won't be in the path, so I figured it was best to handle it within CSharpier.

### DIFF
--- a/Src/CSharpier.Cli/CommandLineFormatter.cs
+++ b/Src/CSharpier.Cli/CommandLineFormatter.cs
@@ -21,10 +21,11 @@ internal static class CommandLineFormatter
                 string directoryOrFile
             )
             {
-                var baseDirectoryPath =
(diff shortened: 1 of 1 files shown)


shocklateboy92 (approved):
LGTM. Thanks!

PR #640, opened 2022-04-04
Dealing with an edge case for conditional access

### LINKED ISSUE #603: Conditional access edge cases
These may be related to the way #596 edge cases are handled

```c#
// these do not format consistently if you remove the ?

            output.Measurements.Add(
                new BenchmarkMeasurement
                {
                    Timestamp = DateTime.UtcNow,
                    Name = "blazorwasm/commit",
                    Value = typeof(Program).Assembly
                        .GetCustomAttributes<AssemblyMetadataAttribute>()
                        .FirstOrDefault(f => f.Key == "CommitHash")?.Value,
                }
            );

                    var runtimeVersion =
                        typeof(object).Assembly
                            .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
                            ?.InformationalVersion ?? "Unknown";

        return httpContext.RequestServices
                ?.GetService<IOptions<JsonOptions>>()
                ?.Value?.SerializerOptions ?? JsonOptions.DefaultSerializerOptions;

var ignoredSourceMembers = _sourceMemberConfigs?
                    .Values.Where(smc => smc.IsIgnored())
                    .Select(pm => pm.SourceMember.Name);

belav:
closes #603

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/InvocationExpressions.cst
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/InvocationExpressions.cst
@@ -234,5 +234,13 @@ class ClassName
                 longParameter_______________________________________
             )
             .CallMethod__________________();
+
+        someThing_______________________.Property
+            .CallMethod__________________()
+            .CallMethod__________________();
+
+        someThing_______________________?.Property
+            .CallMethod__________________()
+            .CallMethod__________________();
     }
 }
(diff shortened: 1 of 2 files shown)


shocklateboy92 (approved):
TBH I'm not sure I like breaking after the property access.
But it makes no sense for conditional access to be different than regular access, so...

belav:
> TBH I'm not sure I like breaking after the property access.

I have similar thoughts, it was copied from prettier and I'm not in love with it.

PR #642, opened 2022-04-04
Updating roslyn and .net6 to handle c# 11 node types.

belav:
The latest version of .net6 caused code to generate that required 4.1.0 of Microsoft.CodeAnalysis.
There is also a new warning about using `.Kind()`, so code was changed to use `.IsKind()` or `RawKind`
The printer was also modified to not error when it runs into unsupported node types, although we have to start parsing code with the preview version of c# if we really want to format c# 11 code.


Src/CSharpier/SyntaxTriviaExtensions.cs
```diff
@@ -4,10 +4,9 @@ public static class SyntaxTriviaExtensions
 {
     public static bool IsComment(this SyntaxTrivia syntaxTrivia)
     {
-        return syntaxTrivia.Kind()
-            is SyntaxKind.SingleLineCommentTrivia
-                or SyntaxKind.MultiLineCommentTrivia
-                or SyntaxKind.SingleLineDocumentationCommentTrivia
-                or SyntaxKind.MultiLineDocumentationCommentTrivia;
+        return syntaxTrivia.IsKind(SyntaxKind.SingleLineCommentTrivia)
```
shocklateboy92, line 7: Looks like the vast majority of uses were along the lines of `Kind() == SyntaxKind.Something`.
Replacing these with `IsKind()` perfectly fine.

However, these 2 cases where we replace a single pattern with multiple method calls doesn't make sense to do "because performance". Looking at the code, it looks like doing `IsKind()` lets them skip a check to make sure it's a C# syntax token:
https://github.com/dotnet/roslyn/blob/dbc30912d98f02aa6ee532e89217d7f884a8fc31/src/Compilers/CSharp/Portable/CSharpExtensions.cs#L185

Since this project only deals with C# code, how do you feel about writing our own extension method that's like `Kind()` but skips the `IsCSharpKind()` check?

Src/CSharpier/SyntaxPrinter/PreprocessorSymbols.cs
```diff
@@ -37,9 +37,10 @@ public static void AddSymbolSet(SyntaxTrivia trivia)
             return;
         }
 
-        var kind = trivia.Kind();
-
-        if (kind is SyntaxKind.IfDirectiveTrivia or SyntaxKind.ElifDirectiveTrivia)
+        if (
+            trivia.IsKind(SyntaxKind.IfDirectiveTrivia)
```
shocklateboy92, line 41: The other comment I made applies here too.

shocklateboy92 (dismissed):
Okay, after thinking some more I managed to nail down exactly what part of this change I don't like.

I have some suggestions on how to make it more palatable. Though, feel free to ignore them if you feel it's overkill :smile:

Src/CSharpier/SyntaxTriviaExtensions.cs
belav: I added RawSyntaxKind and then spent too much time thinking about when to use that vs using IsKind, so settled on just changing everything to use RawSyntaxKind.

PR #648, opened 2022-04-14
Fixing issues with formatting unsaved documents in vscode

### LINKED ISSUE #647: VS Code Extension - Unsaved document issues
I'm having some issues when trying to format a document that has not been written to a file (an unsaved document tab in VS Code). When triggering the *Format Document (Shift + Alt + F)* command, VS Code indicates that the formatting is in progress. However, the formatting never finishes and VS Code keeps showing the busy indication indefinitely or until I switch to another document. 

After starting the formatting, I can observe:
- The document does not get formatted
- VS Code Memory usage increases by about 10 MB per second (no end in sight)
- VS Code uses about 12% CPU
- When restarting VS Code and switching to the unsaved document I tried to format, the Memory leak and CPU usage shows the same pattern again (without triggering another format, see [1] for a possible cause) 
- Closing the document does not stop the Memory leak or the CPU usage
- Disabling the CSharpier extension (and performing the reload that VS Code suggests) does not stop the Memory leak or the CPU usage
- When the CSharpier extension is not installed or disabled when starting VS Code, non of the above behavior occurs (or in other words: everything works as expected)

[1]: After some further experimentation I noticed that as soon as I switch to an unsaved document in language mode *C#* (can be an empty document), the memory leak and CPU usage occurs. So this issue is not only triggered by formatting, but also by switching to an unsaved document in language mode *C#*. However, opening a new document (no language specified) and pasting C# code or manually changing the language mode to *C#* does not trigger this behavior (but selecting any other document and switching back to this new one does).

----

CSharpier Info:
Extension Version: v1.2.1
CSharpier is globally installed with version: 0.16.0

VS Code Info:
Version: 1.66.2 (user setup)
Commit: dfd34e8260c270da74b5c2d86d61aee4b6d56977
Date: 2022-04-11T07:46:01.075Z
Electron: 17.2.0
Chromium: 98.0.4758.109
Node.js: 16.13.0
V8: 9.8.177.11-electron.0
OS: Windows_NT x64 10.0.19044

belav:
closes #647


Src/CSharpier.VSCode/src/CSharpierProcessProvider.ts
```diff
@@ -67,15 +67,35 @@ export class CSharpierProcessProvider implements Disposable {
 
         if (!version || !this.csharpierProcessesByVersion[version]) {
             // this shouldn't really happen, but just in case
+            this.logger.debug("returning NullCSharpierProcess");
             return new NullCSharpierProcess();
```
shocklateboy92, line 71: If the only reason `NullCSharpierProcess` exists is to make the type checker happy, you can just throw an exception instead.

Src/CSharpier.VSCode/src/CSharpierProcessProvider.ts
```diff
@@ -67,15 +67,35 @@ export class CSharpierProcessProvider implements Disposable {
...
+            this.logger.debug("Unsaved document has no directory, falling back to __dirname: " + directory);
+        }
+
+        return directory;
+    };
+
     private getCSharpierVersion = (directoryThatContainsFile: string): string => {
         let currentDirectory = directoryThatContainsFile;
-        while (true) {
+        let x = 0;
```
shocklateboy92, line 97: `x` is kind a terrible name

Src/CSharpier.VSCode/src/CSharpierProcessProvider.ts
belav: `NullCSharpierProcess` is supposed to be for preventing errors from bubbling up to the user over and over. Although I just tested it out, and it seems like vscode eats any exceptions and doesn't bubble them up to the user at all.

Src/CSharpier.VSCode/src/CSharpierProcessProvider.ts
belav: came up with one that is less bad

Src/CSharpier.VSCode/src/CSharpierProcessProvider.ts
```diff
@@ -67,15 +67,35 @@ export class CSharpierProcessProvider implements Disposable {
...
+            this.logger.debug("Unsaved document has no directory, falling back to __dirname: " + directory);
+        }
+
+        return directory;
+    };
+
     private getCSharpierVersion = (directoryThatContainsFile: string): string => {
         let currentDirectory = directoryThatContainsFile;
-        while (true) {
+        let loopNumber = 0;
```
shocklateboy92, line 97: Okay, that's not a _better_ name :stuck_out_tongue: 
How about `parentNumber`?

Or better yet, how about stopping at the filesystem root instead of an arbitrary count?
https://stackoverflow.com/a/70115897

shocklateboy92 (dismissed):

shocklateboy92 (approved):

PR #649, opened 2022-04-18
Re-adding the option to have tabs

### LINKED ISSUE #645: Option for indentation
First, thanks for this project! 
I'm a big fan of Prettier and really happy to see a similar formatting tool for C#.

I understand the opiniated Prettier philosophy of keeping options to a strict minimum that is now frozen, but I feel like indent style (tabs vs spaces) and width are amongst the few things that should be configurable. 
They are (unreasonably) very polarizing and forcing spaces might hinder adoption of Csharpier.

I'm sure you're aware that Prettier does let you configure them:
https://prettier.io/docs/en/options.html#tab-width
Also observe the default Prettier tab-width is `2`, csharpier uses `4`.

The best part is that it seems you already have full support for those option, they just need to be exposed?
https://github.com/belav/csharpier/blob/master/Src/CSharpier.Cli/ConfigurationFileOptions.cs#L44-L48

belav:
closes #645


shocklateboy92 (approved):

shocklateboy92:
Oaf, okay. I came across this post and had to concede:
https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/

PR #651, opened 2022-04-25
VS wanted all of these changes and now doesn't build the non-buildable projects

belav:


shocklateboy92 (dismissed):
Warnings as errors is a really important comment.
But I understand that enabling it requires fixing existing warnings, which is a (potentially) very big change. So I'm approving this as is, but you should really get on that ☺

PR #655, opened 2022-05-02
Error output loses characters, so let's just ditch it.

### LINKED ISSUE #588: Rider - reading from error output randomly loses characters.
When the rider plugin for pipe-multiple-files gets error output, it randomly loses characters.

Adding `logger.debug(nextCharacter)` - all of the characters are being read.
Adding `logger.debug(stringBuilder.toString())` inside of CreateReadingThread - not all of the characters seem to be appended.
Creating a StringBuilder, string, or StringBuffer inside CreateReadingThread does not seem to fix it.
CSharpier does write '\u0003' to stdout, not stderr, so the output thread is the one that finishes.
Adding a thread.sleep into CreateReadingThread after receiving EOF does not help
Trying to write the EOF on stderr resulted in the plugin hanging.

Maybe the solution is to switch to protobuf?
Why do other IDEs not appear to have the same problem?

belav:
closes #588

### DIFF
--- a/Src/CSharpier.Rider/CHANGELOG.md
+++ b/Src/CSharpier.Rider/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # csharpier-rider Changelog
 
+## [1.2.5]
+- Error output loses characters
+
 ## [1.2.4]
 - Support for 2022.1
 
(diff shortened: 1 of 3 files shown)


shocklateboy92 (approved):
Ooh. No more dealing with threads. Nice!

PR #656, opened 2022-05-02
Error reading code has potential issues, ditching it.

belav:
There isn't an obvious issue like rider, but if the error chunk is ever big enough to come back as two pieces, this would definitely cause issues.

### DIFF
--- a/Src/CSharpier.VSCode/CHANGELOG.md
+++ b/Src/CSharpier.VSCode/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [1.2.3]
+- Proper fix for unsaved documents not formatting.
+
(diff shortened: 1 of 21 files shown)


Src/CSharpier.VSCode/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## [1.2.4]
+- Error output loses characters
```
shocklateboy92, line 2: In hindsight, this isn't a release note: it doesn't explain what's changed.
Maybe something like:
```
Remove error output display. It used to lose characters.
```

shocklateboy92 (approved):

PR #662, opened 2022-05-09
Handling some edge cases with #if directives

### LINKED ISSUE #659: Format only the selected code
Hello team, here with a feature request. 

In most cases, we don't want to format the whole class but certain complex parts of the code. With this in mind, it would be super useful if formatting would be possible only on the selected part of the code.

Please look into the possibility of this implementation.

Thanks in advance! 🤞

belav:
closes #659


belav:
I hate most of this code but couldn't come up with a better way to do things. There are also a couple edge cases that I didn't handle but I doubt anyone in the real world will run into them. I think most usages of conditional compilation are pretty basic.

This maybe could have moved the printing of the leading trivia into `MembersWithForcedLines` but that would require changes to a lot of areas to make sure they didn't also print the leading trivia. And the logic inside of Token probably still would be pretty gross.

I do have a PR where I am playing around with an idea for getting rid of the global ThreadStatics by passing a `FormattingContext` through all the print methods.

PR #670, opened 2022-05-21
Ditching some ugly thread statics

belav:


Src/CSharpier/SyntaxPrinter/FormattingContext.cs
```diff
@@ -1,3 +1,8 @@
 namespace CSharpier.SyntaxPrinter;
 
-public class FormattingContext { }
```
shocklateboy92, line 3: Note: you should consider making this a record, since it's probably never going to have methods on it.

shocklateboy92 (approved):
I didn't read through #663, but I'm going to assume this class will be available everywhere now.

Also, we only do 1 formatting job per thread, right? This class will never need to be thread-safe/re-entrant?

belav:
> I didn't read through #663, but I'm going to assume this class will be available everywhere now.

Basically yeah. It gets passed through almost all of the print methods.

> Also, we only do 1 formatting job per thread, right? This class will never need to be thread-safe/re-entrant?

That's correct

PR #671, opened 2022-05-21
Fixing bug with empty csproj causing csharpier to fail

### LINKED ISSUE #665: Empty (or malformed) .csproj file will cause csharpier to fail.
1. In a directory, add an empty Project.csproj file
2. Run csharpier in that directory

CSharpier fails while trying to load the csproj file. It should either ignore these types of failures, or just log them. But it should not fail formatting because of them.

belav:
closes #665

### DIFF
--- a/Src/CSharpier.Cli/HasMismatchedCliAndMsBuildVersions.cs
+++ b/Src/CSharpier.Cli/HasMismatchedCliAndMsBuildVersions.cs
@@ -21,7 +21,20 @@ public static class HasMismatchedCliAndMsBuildVersions
         {
             // this could potentially use the Microsoft.CodeAnalysis.Project class, but that was
             // proving difficult to use
-            var csProjXElement = XElement.Load(fileSystem.File.OpenRead(pathToCsProj));
+            XElement csProjXElement;
+            try
+            {
+                csProjXElement = XElement.Load(fileSystem.File.OpenRead(pathToCsProj));
+            }
+            catch (Exception ex)
+            {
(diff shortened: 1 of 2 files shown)


Src/CSharpier.Cli/HasMismatchedCliAndMsBuildVersions.cs
```diff
@@ -21,7 +21,20 @@ public static bool Check(string directory, IFileSystem fileSystem, ILogger logge
         {
             // this could potentially use the Microsoft.CodeAnalysis.Project class, but that was
             // proving difficult to use
-            var csProjXElement = XElement.Load(fileSystem.File.OpenRead(pathToCsProj));
+            XElement csProjXElement;
```
shocklateboy92, line 24: Odd. How come this doesn't need to be an `XElement?`?

shocklateboy92 (approved):

Src/CSharpier.Cli/HasMismatchedCliAndMsBuildVersions.cs
belav: I missed the `?` next to XElement and was very confused by your question at first...

After playing around it seems to be because if you do something like this
```c#
XElement csProjXElement;
var x = csProjXElement.XPathSelectElements(xpath).FirstOrDefault();
```
Then you get the error
```
Local variable 'csProjXElement' might not be initialized before accessing
```

So it's possible to have non-nullable element that isn't immediately initialized.

Issue #672, opened 2022-05-22
Improve printing for arguments


belav:
There are improvements that could be made to printing arguments, namely with lambdas as parameters. #669 stuck to just improving some simple cases (single arguments or parenthesized lambdas with no arguments). Prettier has more complicated logic that we should look into.

A simple lambda with too long of an argument name + a block doesn't break when it should.
Simple lambdas with no block.
Parenthesized lambdas with no block.
Parenthesized lambdas with arguments.

belav:
```c#
            builder.Entity<IdentityUserToken<string>>(b =>
            {
                b.HasKey(
                    l =>
                        new
                        {
                            l.UserId,
                            l.LoginProvider,
                            l.Name
                        }
                );
                b.ToTable("AspNetUserTokens");
            });

**Comment by belav** [comment]
this may be more because of breaking with the `new { }` stuff that came from 
```c#
builder.Entity<EntityWithThreeProperties>(e =>
                {
                    e.HasIndex(
                        t =>
                            new
                            {
                                t.X,
                                t.Y,
                                t.Z
                            },
                        "IX_unspecified"
                    );
                    e.HasIndex(
                            t =>
                                new
                                {
                                    t.X,
                                    t.Y,
                                    t.Z
                                },
                            "IX_empty"
                        )
                        .IsDescending();
                    e.HasIndex(
                            t =>
                                new
                                {
                                    t.X,
                                    t.Y,
                                    t.Z
                                },
                            "IX_all_ascending"
                        )
                        .IsDescending(false, false, false);
                    e.HasIndex(
                            t =>
                                new
                                {
                                    t.X,
                                    t.Y,
                                    t.Z
                                },
                            "IX_all_descending"
                        )
                        .IsDescending(true, true, true);
                    e.HasIndex(
                            t =>
                                new
                                {
                                    t.X,
                                    t.Y,
                                    t.Z
                                },
                            "IX_mixed"
                        )
                        .IsDescending(false, true, false);
                });

**Comment by belav** [comment]
```
class ClassName
{
    void MethodName()
    {
        var searchRequest = new SearchDescriptor<ElasticsearchProduct>().WithAggregations(
            o =>
                o.WithFilter(
                    key,
                    f =>
                        f.WithFilter(fd => filter)
                            .WithAggregations(
                                ad =>
                                    ad.WithTerm(
                                        key,
                                        cd =>
                                            cd.WithField(facetField).WithSize(MaximumCategoryFacets)
                                    )
[...]

PR #673, opened 2022-05-22
Better lambda indent

### LINKED ISSUE #669: Excessive indent level with lambda as the only method call argument
```cs
Foo(() =>
{
    Bar();
    Baz();
});
```
is currently formatted to:
```cs
Foo(
    () =>
    {
        Bar();
        Baz();
    }
);
```
It is an extra unnecessary level of indentation.

belav:
This handles the simple cases, #672 will take care of some more complex cases.
closes #669

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/InvocationExpressions.cst
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/InvocationExpressions.cst
@@ -197,21 +197,17 @@ class ClassName
         // comment shouldn't do weird stuff
         SomeObject.CallMethod();
 
-        CallMethod(
-                lambdaWithBlock =>
-                {
-                    return;
-                }
-            )
+        CallMethod(lambdaWithBlock =>
+            {
+                return;
+            })
             .CallMethod()
             .CallMethod();
 
-        someObject.CallMethod(
-            lambdaWithBlock =>
-            {
-                return;
-            }
-        );
+        someObject.CallMethod(lambdaWithBlock =>
+        {
+            return;
+        });
 
         o.CallMethod(parameter_____________, parameter_____________)
             .CallMethod(parameter_____________, parameter_____________);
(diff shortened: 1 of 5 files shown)


Src/CSharpier.Tests/FormattingTests/TestFiles/InvocationExpressions.cst
```diff
@@ -197,21 +197,17 @@ class ClassName
...
             .CallMethod()
             .CallMethod();
 
-        someObject.CallMethod(
-            lambdaWithBlock =>
-            {
-                return;
-            }
-        );
+        someObject.CallMethod(lambdaWithBlock =>
```
shocklateboy92, line 207: Is this really "better"? IMHO the previous version was more readable for both these cases.

shocklateboy92 (approved):

Src/CSharpier.Tests/FormattingTests/TestFiles/InvocationExpressions.cst
belav: I think it is mostly better, with those examples being the ones that fall into "maybe not better" category.

These examples I think are definitely better. It saves vertical space and I don't think it is any less readable there.
https://github.com/belav/csharpier-repos/pull/41/commits/831f2012d4b99fa4f5cbdec68b6c2cb988a71fce#diff-bc2e3864cdf7b47541cca4aa46b6d378cb2c11cb0091dca0647fed8a565a7dbf

It indents less too, which helps in other cases like this
https://github.com/belav/csharpier-repos/pull/41/commits/831f2012d4b99fa4f5cbdec68b6c2cb988a71fce#diff-668431d269a33a846504a3895822785838e4d92a10e9a3ce025b37de65724ed2

fwiw, prettier formats it like this. I don't really like that they force `( )` on it
```js
someObject.CallMethod((lambdaWithBlock) => {
    return;
});

PR #677, opened 2022-05-28
Switching to vite, ditching styled components because it wasn't playing nice with vite

belav:


shocklateboy92 (commented):
Okay, I just realized I had draft comments from a half done review this whole time so I'm submitting it.

Also, my reviews will probably continue to be spotty for the next week or 3 because I'm currently driving through the western half of the country with a friend.

PR #680, opened 2022-06-02
use relative file path in CommandLineFormatter

dlech:
This changes the format of the file name to add a leading "." so that the path is a relative path instead of an absolute path.

Modern terminal programs allow control-clicking on paths to jump to a file, so if this is a relative path, we can jump to the file. If it is printed as an absolute path, this doesn't work.


belav:
I think this is a great idea but I did find a couple issues. 

I think I added a unit test to recreate the first, but I am having trouble running the unit tests right now. The source generators are causing me some pain after I started looking at supporting c# 11. I at least managed to build things so that I could test the cli.

1. Passing a subdirectory as a path, the relative path output isn't correct
```
PS C:\projects> dotnet csharpier csharpier  
Error /Src/CSharpier.Generators/NodePrinterGenerator.cs - Failed to compile so was not formatted.

PS C:\projects> C:\projects\csharpier\Src\CSharpier.Cli\bin\Debug\net6.0\dotnet-csharpier.exe csharpier
Error .Src/CSharpier.Generators/NodePrinterGenerator.cs - Failed to compile so was not formatted.
```
2. Passing a fully qualified path, csharpier currently outputs a path relative to the path that was input (although missing the .) - to get it clickable would require it to be relative to the current directory, or a fully qualified path. Maybe the fully qualified path makes more sense, because the path supplied was fully qualified.
```
PS C:\projects> dotnet csharpier c:\projects\csharpier
Error /Src/CSharpier.Generators/NodePrinterGenerator.cs - Failed to compile so was not formatted.

PS C:\projects> C:\projects\csharpier\Src\CSharpier.Cli\bin\Debug\net6.0\dotnet-csharpier.exe c:\projects\csharpier
Error .Src/CSharpier.Generators/NodePrinterGenerator.cs - Failed to compile so was not formatted.
```

belav:
I may have my terminology confused.
I intended csharpier to output paths relative to the path that was supplied to be formatted, but they were missing the `.`.

If csharpier is supplied a relative path, like `.` or `Subdirectory`, then it should output a relative path that is clickable.
If csharpier is supplied an absolute path, like `c:/projects/csharpier`, then it should probably output an absolute path that is clickable.

And it looks like if you supply a path as `./Subdirectory` then csharpier fails. I am going to create a bug for that.

belav (approved):

dlech:
Thanks for fixing this up. I didn't have time to get back to it yet.

belav:
You're welcome!

PR #683, opened 2022-06-06
Ditching problematic source generators

belav:


shocklateboy92:
So uh... does this mean you're no longer auto-generating this code? Maintaining it manually seems like it'll be kind of a pain. What issues did you run into?

PR #695, opened 2022-06-29
docs: Fix the CI example codes to be valid

eai04191:
- Fixed the format of the dotnet local tool command to be as shown in the .NET docs.
- Fixed typo for "github actions"

I'm not good at dotnet, but the current actions example didn't really work.
I fix the command to work by referring to the [.NET docs](https://docs.microsoft.com/en-us/dotnet/core/tools/local-tools-how-to-use).

### DIFF
--- a/Docs/ContinuousIntegration.md
+++ b/Docs/ContinuousIntegration.md
@@ -22,9 +22,9 @@ Normally when using a code formatter like CSharpier, you'll want to ensure that 
 2. Use your preferred CI/CD tool to run the following commands.
    ```bash
    dotnet tool restore
-   dotnet-csharpier --check
+   dotnet tool run dotnet-csharpier --check .
    ```   
-   An example of a github action to accomplish this
+   An example of a GitHub Actions to accomplish this
    ```yaml
    name: Validate PR
    on:
@@ -38,7 +38,7 @@ Normally when using a code formatter like CSharpier, you'll want to ensure that 
          - uses: actions/checkout@v2
          - run: |
            dotnet tool restore
-           dotnet csharpier --check
+           dotnet tool run dotnet-csharpier --check .
    
    ```
 
--- a/Docs/ContinuousIntegration.md
+++ b/Docs/ContinuousIntegration.md
@@ -24,7 +24,7 @@ Normally when using a code formatter like CSharpier, you'll want to ensure that 
    dotnet tool restore
    dotnet tool run dotnet-csharpier --check .
    ```   
-   An example of a GitHub Actions to accomplish this
+   An example Github Actions workflow to accomplish this
    ```yaml
    name: Validate PR
    on:


Docs/ContinuousIntegration.md
```diff
@@ -22,9 +22,9 @@ Normally when using a code formatter like CSharpier, you'll want to ensure that
 2. Use your preferred CI/CD tool to run the following commands.
    ```bash
    dotnet tool restore
-   dotnet-csharpier --check
+   dotnet tool run dotnet-csharpier --check .
    ```   
-   An example of a github action to accomplish this
+   An example of a GitHub Actions to accomplish this
```
belav, line 27: This sounds a little weird, and the original version wasn't quite right either.
After looking at the doc for Github Actions, maybe it should be something like
```
An example Github Actions workflow to accomplish this

File: Docs/ContinuousIntegration.md [hunk 2]
```diff
@@ -22,9 +22,9 @@ Normally when using a code formatter like CSharpier, you'll want to ensure that
 2. Use your preferred CI/CD tool to run the following commands.
    ```bash
    dotnet tool restore
-   dotnet-csharpier --check
+   dotnet tool run dotnet-csharpier --check .
```
belav, line 25: Both `dotnet csharpier` and `dotnet-csharpier` work locally for me and in the workflows I have set up. I'm running windows, and the workflows run on ubuntu.
Most places I'm looking don't mention `dotnet tool run` at all, besides the microsoft doc for creating a tool.
The docs for these don't mention `dotnet tool run`
https://www.nuget.org/packages/dotnet-ef/
https://www.nuget.org/packages/dotnet-serve/

Could this be a problem specific to your setup somehow?

The `.` definitely needs to be there though.

belav (changes requested):
Thanks for the submission, I just had a couple of comments.

Docs/ContinuousIntegration.md
eai04191: Here is an example I actually tried.
On ubuntu-latest, the `dotnet-csharpier` command does not work after restore. `dotnet tool run` works reliably.
https://github.com/eai04191/actions-testing/commits/csharpier

Docs/ContinuousIntegration.md
eai04191: Absolutely! I'm not a native English speaker, so I lacked a perspective on the naturalness of the text.

Docs/ContinuousIntegration.md
shocklateboy92: I see from your commit log you've tried `dotnet-csharpier` and it didn't work.
Could you please also try
```
dotnet csharpier --check .
```

eai04191: I'm late to fix but I fixed it like that so I think we can merge.

shocklateboy92 (approved):

belav (approved):

Issue #542, opened 2022-01-08
Extensions - they don't work properly with CSharpier.MSBuild


belav:
I need to test this, but I believe depending on the version of csharpier, you will either

- Get the file formatted by the global version, which could be a mismatched version
- The global csharpier will detect that the msbuild version doesn't match, and files will not format and you may not get any useful message about it.

Issue #692, opened 2022-06-20
Look into implementing caching to speed up formatting.


belav:
Prettier recently added caching to determine if formatting a file can be skipped. https://prettier.io/docs/en/cli.html#--cache

Look into what they are doing and determine if a similar feature can be added to csharpier.

shocklateboy92:
Is that really necessary? I feel like csharpier is already bloody fast. :stuck_out_tongue:

belav:
But what if we can make it super bloody fast?
It is ~22 seconds on our work main repo (13,000 files or so) without the syntax checker.
If caching can drop that down to like 5 seconds that would be awesome. I'd do a quick POC first to see if it makes sense to go through with it.

belav:
My initial POC using last modified date - the 13,000 file repo can be formatted in 3 seconds if caching is used and no files have changed. 10x improvement!

PR #698, opened 2022-07-04
Modify VSCode extension so it can detect version from CSharpier.MSBuild

belav:
#542


belav:
So this doesn't handle a number of things.

1. Central package management - there are a few ways to do that https://www.mytechramblings.com/posts/centrally-manage-nuget-versions/
2. packages.config - kinda legacy at this point
3. adding a packages to all csprojs via a Directory.Build.props file

I am wondering if it makes sense to modify the msbuild packages to instead add/update the dotnet-tools.json file. Or specify a version number some other way.

shocklateboy92:
Hmmm. There was previously no XML parsing at all in this plugin, right?
I can see how this approach can get a little cumbersome (especially when ported to all the other plugins).

I'm already not a super big fan of the fact that source files get modified on build (yes, I know I came with the MsBuild plugin idea in the first place - irony is not lost on me). So I'm reluctant to mess with their dotnet tools manifest on build without telling the user.

How about we write to a special temporary file in the `$(IntermediateOutputPath)` that we control? Hopefully users wouldn't even notice. 
I would suggest having csharpier do the actual writing and having MsBuild pass a special parameter letting it know to do that.

belav:
> Hmmm. There was previously no XML parsing at all in this plugin, right?
I can see how this approach can get a little cumbersome (especially when ported to all the other plugins).

The VS plugin will be pretty straightforward, dealing with XML isn't bad in c#. And CSharpier already does a check for what version of CSharpier.MsBuild someone is using to make sure there isn't a mismatch. I'm more worried about all the edge cases.

> I'm already not a super big fan of the fact that source files get modified on build

There are definitely people out there that like this approach though. For smaller projects it is a bit more straightforward to setup. And smaller projects are probably less likely to run into the edge cases that I called out so maybe I shouldn't worry so much about them. And a work around is that someone can add the manifest file even if they do use the MsBuild package.

> How about we write to a special temporary file in the $(IntermediateOutputPath) that we control?

That folder could get deleted now and then, but writing the data somewhere does make sense to me. For #692 to work, we'd need to store data somewhere to keep track of the last time a file was formatted. In that same place we store the caching data we can store data about which version of the MsBuild package is being used. When a restore is run, we can create that data. Maybe that data will end up in $(IntermediateOutputPath) or in the same place we download csharpier when we run plugins (AppData on windows, I forget where on linux).

That'll only work for people on a new version of CSharpier.MsBuild. So I like the idea of this xml parsing being around as a fallback for anyone on older versions of CSharpier.MsBuild but a new version of the plugin. It is a lot easier to upgrade the IDE plugin. Updating CSharpier comes with potential changes in formatting to a lot of files.

And maybe we just wait and see if anyone actually runs into the edge cases that I called out, and add this extra data writing at that time.

shocklateboy92 (approved):
If you want to do this, I have no strong objections. But you should maybe make csharpier start writing state to the intermediate build directory anyway, in case we decide to use that information later.

PR #702, opened 2022-07-11
VS - making options specific to the solution

### LINKED ISSUE #696: IDE plugin format on save as a per-project setting
Problem: I like csharpier and want to use it, so I install the plugin for my editor and turn the format on save setting.  Unfortunately not all of the projects I work on use it, and when I switch to one of those projects I often forget about the auto formatting until I go to commit and notice a lot of formatting changes.

I can think of a few possible solutions to this issue:

1. Replace the current (boolean on/off) IDE format on save setting with a dropdown that has options such as 'Never', 'Always', and 'When the project uses CSharpier'.  CSharpier usage in a project could be detected by checking the project's `dotnet tool list`, or looking for the presence of one of the config files.
2. Add a setting to the configuration such as `formatOnSave: boolean`.  The setting would be ignored by the tool itself and used by IDE plugins.
3. Most IDEs have a way to do project specific configuration that overrides the IDE global settings, so assuming that's expose through their plugin API it could be used to provide a project specific setting.

belav:
It was painful, but I figured out a way to save settings that are local to the solution.
Implementing `IVsPersistSolutionOpts` allows the extension to save options in the `.suo` file, but this was proving flaky.
closes #696


Src/CSharpier.VisualStudio/CSharpier.VisualStudioShared/CSharpierOptionsPage.cs
```diff
@@ -1,18 +1,27 @@
...
-    public class CSharpierOptionsPage : DialogPage
+    public class CSharpierOptionsPage : BaseOptionPage<CSharpierOptions> { }
+
+    public class CSharpierOptions : BaseOptionModel<CSharpierOptions>
     {
         [Category("CSharpier")]
         [DisplayName("Reformat with CSharpier on Save")]
-        [Description("Reformat with CSharpier on Save")]
+        [Description(
+            "Reformat with CSharpier on Save - this option is saved locally to the solution"
```
shocklateboy92, line 12: Do you plan to add another option that's not per-solution?

shocklateboy92 (dismissed):
Might be because I don't know much VS internals, but LGTM.

Issue #699, opened 2022-07-04
Ignore node_modules


RichiCoder1:
This may seem an odd ignore to add, but we're using `csharpier` in a mixed npm and dotnet project (using AWS's CDK), and as a result the `node_modules` folder contains `.cs` template code. Would be nice to have this be a default ignore!

TiraelSedai:
Any specific reason that explicitly adding `**/node_modules/*.cs` to `.csharpierignore` file won't do the trick?

RichiCoder1:
It did and that absolutely worked! Was more of requesting a default.

PR #703, opened 2022-07-15
Add node modules as default ignore

SubjectAlpha:
As requested by issue #699

### DIFF
--- a/Src/CSharpier.VSCode/package.json
+++ b/Src/CSharpier.VSCode/package.json
@@ -12,7 +12,7 @@
     },
     "license": "MIT",
     "bugs": {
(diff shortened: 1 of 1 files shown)


.csharpierignore
```diff
@@ -1 +1,2 @@
-Uploads/
\ No newline at end of file
+Uploads/
+**/node_modules/*.cs
```
belav, line 2: This is actually the ignore file for csharpier itself, the code change would probably end up going in https://github.com/belav/csharpier/blob/master/Src/CSharpier.Cli/CommandLineFormatter.cs

belav (changes requested):

PR #707, opened 2022-07-18
Adding a cache option to speed up formatting.

### LINKED ISSUE #692: Look into implementing caching to speed up formatting.

belav:
closes #692

### DIFF
--- a/Src/CSharpier.Cli/FormattingCache.cs
+++ b/Src/CSharpier.Cli/FormattingCache.cs
@@ -31,7 +31,7 @@ internal static class FormattingCacheFactory
         CancellationToken cancellationToken
     )
     {
-        if (commandLineOptions.NoCache)
+        if (commandLineOptions.NoCache || commandLineOptions.Check)
         {
             return NullCache;
         }
--- a/Src/CSharpier.Cli/FormattingCache.cs
+++ b/Src/CSharpier.Cli/FormattingCache.cs
@@ -3,6 +3,7 @@ using System.IO.Abstractions;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.Json;
+using CSharpier.Utilities;
 
 namespace CSharpier.Cli;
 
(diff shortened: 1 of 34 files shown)


Src/CSharpier.Cli/FormattingCache.cs
```diff
@@ -0,0 +1,112 @@
+using System.Collections.Concurrent;
+using System.IO.Abstractions;
+using Newtonsoft.Json;
```
shocklateboy92, line 3: I know this is just a POC, but you should avoid adding new dependencies on `Newtonsoft`.
Use `System.Text.Json` instead.

Src/CSharpier.Cli/FormattingCache.cs
```diff
@@ -0,0 +1,144 @@
...
+        private static string GetOptionsHash(PrinterOptions printerOptions)
+        {
+            var csharpierVersion = typeof(FormattingCache).Assembly.GetName().Version;
+            return Hash($"{csharpierVersion}_${JsonSerializer.Serialize(printerOptions)}");
+        }
+
+        private static string Hash(string input)
+        {
+            using var md5 = MD5.Create();
+            var result = md5.ComputeHash(Encoding.ASCII.GetBytes(input));
```
shocklateboy92, line 108: MD5 is kind of outdated and if you enable code analysis they will warn you when they see it.
Better of using SHA-1 or one of its variants from the get go.

shocklateboy92 (approved):
LGTM. Left a minor comment that accidentally wasn't part of this review

Src/CSharpier.Cli/FormattingCache.cs
belav: I assumed MD5 would be faster even if it is outdated, but it turns out SHA1 is actually a bit faster. Or using the new XxHash32 is even faster with a smaller cache size!

Working on a repo with 11k files, non-scientific benchmark (ran it a few times and did the average in my head), with everything already formatted
md5 3.9s 2mb file
sha1 3.8s 2.2mb file
xxh32 3.4s 1.5mb file

It does require using a preview version of System.IO.Hashing. There are no open issues related to xxh32 and the recent history is pretty quiet - https://github.com/dotnet/runtime/commits/main/src/libraries/System.IO.Hashing

belav:
Whoops @shocklateboy92, I just realized this wasn't created as a draft. It is now ready to review not counting the doc changes I will hopefully still get to today.

PR #711, opened 2022-07-30
Code changes for reading msbuild version in rider plugin

### LINKED ISSUE #542: Extensions - they don't work properly with CSharpier.MSBuild

belav:
closes #542

### DIFF
--- a/Src/CSharpier.Rider/CHANGELOG.md
+++ b/Src/CSharpier.Rider/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # csharpier-rider Changelog
 
+## [1.3.0]
+- Adding support for detecting version of CSharpier.MsBuild from csproj
+
 ## [1.2.7]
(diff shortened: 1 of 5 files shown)


shocklateboy92 (approved):
What happens if there are multiple `csproj` files in the solution that reference different versions?

It's okay if you don't want to handle that case. But you should probably call that out in the release notes (or better yet, print a warning when it encounters the situation).

belav:
> What happens if there are multiple `csproj` files in the solution that reference different versions?
> 
> It's okay if you don't want to handle that case. But you should probably call that out in the release notes (or better yet, print a warning when it encounters the situation).

Assuming a basic setup, this will handle that situation. For each file it needs to format, it looks for a csproj or tool manifest above it in the file system. I can document that somewhere.

PR #722, opened 2022-08-22
Rider - detect bad install + disable global warning

belav:


shocklateboy92:
LGTM. Kudos for removing the message about installing csharpier globally that I keep getting _all the time_ 🗡️

PR #726, opened 2022-08-23
docs: Added neovim plugin option

jjohnson338:
I recently merged a PR to [neoformat](https://github.com/sbdchd/neoformat) to support `csharpier` out of the box. This PR updates the editor plugin docs to reference it as an option for `neovim` users.

[Referenced neoformat PR](https://github.com/sbdchd/neoformat/pull/428)

### DIFF
--- a/Docs/Editors.md
+++ b/Docs/Editors.md
@@ -7,6 +7,7 @@ Running CSharpier on save is recommended. It will speed up your development time
 
 ### Visual Studio
 Use the [official 2022 extension](https://marketplace.visualstudio.com/items?itemName=csharpier.CSharpier)
+\
 Use the [official 2019 extension](https://marketplace.visualstudio.com/items?itemName=csharpier.CSharpier2019)
 ### Visual Studio Code
 Use the [official extension](https://marketplace.visualstudio.com/items?itemName=csharpier.csharpier-vscode)
--- a/Docs/Editors.md
+++ b/Docs/Editors.md
@@ -6,9 +6,11 @@ hide_table_of_contents: true
 Running CSharpier on save is recommended. It will speed up your development time.
 
 ### Visual Studio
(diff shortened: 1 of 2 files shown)


Docs/Editors.md
```diff
@@ -6,9 +6,11 @@ hide_table_of_contents: true
 Running CSharpier on save is recommended. It will speed up your development time.
 
 ### Visual Studio
-Use the [official 2022 extension](https://marketplace.visualstudio.com/items?itemName=csharpier.CSharpier)  
+Use the [official 2022 extension](https://marketplace.visualstudio.com/items?itemName=csharpier.CSharpier)
```
belav, line 9: Was this an accidental change? There may be a better way for me to have these on separate lines.

belav (changes requested):
Thanks for adding support for csharpier!

Do you mind also making the change to https://github.com/belav/csharpier/blob/master/Src/Website/docs/Editors.md, those are the "released" versions of the doc that the website uses.

Docs/Editors.md
jjohnson338: Sorry, that was a result of my automatic trim trailing whitespace I've scripted into my editor. I can add it back, if you'd like.

jjohnson338:
> Thanks for adding support for csharpier!
> 
> Do you mind also making the change to https://github.com/belav/csharpier/blob/master/Src/Website/docs/Editors.md, those are the "released" versions of the doc that the website uses.

No problem! I just pushed a commit doing that.

Docs/Editors.md
jjohnson338: I pushed a third commit to this PR breaking the lines in the editor documentation without the use of trailing whitespace. You can use a `\` on a line to trigger a line-break.

belav: Thanks, `\` seems like a much better option, I never liked the trailing whitespace solution but didn't bother to go looking.

belav (approved):

PR #741, opened 2022-10-18
Add line when await precedes query expression

### LINKED ISSUE #740: Await + LINQ query syntax breaks formatting
Given the following code:

```cs
var result = await from a in b
    from c in d
    from e in SomethingAsync(a, c)
    select e;
```

It gets formatted into:

```cs
var result = await from a in b
from c in d
from e in SomethingAsync(a, c)
select e;
```

instead of

```cs
var result = await 
    from a in b
    from c in d
    from e in SomethingAsync(a, c)
    select e;
```

domn1995:
Hi, @belav! I believe this PR fixes #740, but I am not sure whether my solution is optimal, nor whether it's complete. I would really appreciate your feedback and any direction if this is not adequate.

Thanks!

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/QueryExpressions.cst
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/QueryExpressions.cst
@@ -63,5 +63,11 @@ class ClassName
                 Country________________________ = g.Key,
                 CustCount____________________________ = g.Count()
             };
+
+        var asyncQuery = await
+            from c in customers
+            from u in users
+            from res in HandleAsync(c, u)
+            select res;
     }
 }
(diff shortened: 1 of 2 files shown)


shocklateboy92 (approved):
LGTM!

Assuming this doesn't break any tests, I don't expect @belav will have any objections.
Will give a few days just in case and merge.

belav:
The changes look good, I'm just waiting on the results of running `/Scripts/CreateTestingPR.ps1` to see if there are any edge cases that would need to be handled.

belav:
Running `/Scripts/CreateTestingPR.ps1` told me was that this case doesn't appear at all in https://github.com/belav/csharpier-repos. Perhaps none of the repos there use query syntax, or at least not with await.

PR #744, opened 2022-10-19
order modifiers

### LINKED ISSUE #725: sort modifiers
e.g. methods like 

````cs
public override async Task Method1() {...} 
async public override Task Method1() {...} 
```` 

are equivalent 

It'd be nice to have first visibility and last async, or whatever, but consistent.

This is similar to sort `using`s maybe and [that discussion](https://github.com/belav/csharpier/issues/661) might apply here too.

### PR DESCRIPTION
**Pull request by glmnet**
an attempt to fix #725

I believe now comes the "opinionated" part, I could implement this order default value (which I found after this quick implementation):

`public, private, protected, internal, file, static, extern, new, virtual, abstract, sealed, override, readonly, unsafe, required, volatile, async`

from:

 https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/Modifiers_Sort_Preprocessors.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/Modifiers_Sort_Preprocessors.test
@@ -1,0 +1,15 @@
+public class DontReorderThis
+{
+#if !SILVERLIGHT
+    [SecuritySafeCritical]
+#endif
+    internal
+#if !SILVERLIGHT
+    unsafe
+#endif
+    static String Method() { }
+
+#if MONO
+    public void NewSymbolSet() { }
+#endif
+}
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/Modifiers_Sort_Preprocessors.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/Modifiers_Sort_Preprocessors.test
@@ -1,7 +1,7 @@
 public class DontReorderThis
 {
 #if !SILVERLIGHT
-    [SecuritySafeCritical]
+    [Attribute]
 #endif
     internal
 #if !SILVERLIGHT
@@ -12,4 +12,13 @@ public class DontReorderThis
 #if MONO
     public void NewSymbolSet() { }
 #endif
+
+#if !SILVERLIGHT
+    [Attribute]
+#endif
+    private
+#if !SILVERLIGHT
+    static
+#endif
+    string Field;
 }
--- a/Src/CSharpier/SyntaxPrinter/Modifiers.cs
+++ b/Src/CSharpier/SyntaxPrinter/Modifiers.cs
@@ -22,12 +22,14 @@ internal static class Modifiers
             return Doc.Null;
         }
 
+        var sortedModifiers = modifiers.OrderBy(o => o.Text);
+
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

**Review by shocklateboy92 — CHANGES_REQUESTED** [review]
Hey, I took a quick look at the test files and yeah I'm not a super big fan of putting `async` before `public`.

Would very much prefer if you sorted them in the order you listed (as per style rule IDE00036).

**Comment by belav** [comment]
The other part of this is figuring out how to get `SyntaxNodeComparer` to not complain that modifiers were reordered. A lot of that code is generated by `CSharpier.FakeGenerators`. Maybe it needs to call a new method `CompareModifiers` instead of the existing `CompareLists`, unless `CompareLists` can determine that is is comparing modifiers.

**Comment by voxlz** [comment]
Is this forgotten? I just ran into this issue, and would love to see it implemented.

File: Src/CSharpier/SyntaxPrinter/Modifiers.cs [hunk 1]
```diff
@@ -2,34 +2,92 @@ namespace CSharpier.SyntaxPrinter;
...
+            "readonly",
+            "unsafe",
+            "required",
+            "volatile",
+            "async",
+        };
+
+        public int Compare(string? x, string? y)
+        {
+            int GetIndex(string? value)
```
shocklateboy92, line 31: why is this a nested method? 
I feel like it could be a private static method in that class.

Src/CSharpier/SyntaxPrinter/Modifiers.cs
```diff
@@ -2,34 +2,92 @@ namespace CSharpier.SyntaxPrinter;
...
+            {
+                var result = Array.IndexOf(DefaultOrdered, value);
```
shocklateboy92, line 33: So, this gets called for every comparison of _every_ declaration in the source code, right?
It might be worth optimizing to be constant time access by say, initializing a static dictionary of string to index.

Src/CSharpier/SyntaxPrinter/Modifiers.cs
```diff
@@ -2,34 +2,92 @@ namespace CSharpier.SyntaxPrinter;
...
         FormattingContext context
     )
+    {
+        return PrintWithSortedModifiers(
+            modifiers,
+            sortedModifiers =>
+                Doc.Group(
+                    Token.PrintWithoutLeadingTrivia(sortedModifiers.First(), context),
+                    " ",
+                    sortedModifiers.Count() > 1
```
shocklateboy92, line 63: Doesn't this give a "multiple enumeration of IEnumerable" error?
It might be better to make `PrintWithSortedModifers()` call `ToArray()` on `sortedModifers` and pass it as an `IReadOnlyList` instead of an `IEnumerable`.

Src/CSharpier.Tests/FormattingTests/TestFiles/cs/Modifiers_Sort_Preprocessors.test
```diff
@@ -0,0 +1,24 @@
+public class DontReorderThis
+{
+#if !SILVERLIGHT
+    [Attribute]
+#endif
+    internal
+#if !SILVERLIGHT
+    unsafe
+#endif
+    static String Method() { }
```
shocklateboy92, line 10: Damn, how did you even come up with this case :astonished:

shocklateboy92 (approved):

PR #746, opened 2022-10-24
Fix so that empty lines inside of ranged ignores are maintained.

### LINKED ISSUE #742: Csharpier removes empty lines in ignored blocks of code
Reproduction code:
```cs
public class ClassName 
{
    public void LongUglyMethod() 
    { 
    	// csharpier-ignore-start
        LongUglyMethodName(context, magic[1],
            "very-long-string-one",
            "very-long-string-two",
            "very-long-string-three"
        );

        LongUglyMethodName(context, magic[2],
            "very-long-string-one",
            "very-long-string-two",
            "very-long-string-three"
        );
        // csharpier-ignore-end
        
        // csharpier-ignore
        LongUglyMethodName(context, magic[1],
            "very-long-string-one",
            "very-long-string-two",
            "very-long-string-three"
        );

        // csharpier-ignore
        LongUglyMethodName(context, magic[2],
            "very-long-string-one",
            "very-long-string-two",
            "very-long-string-three"
        );
    }
}
```

Expected:
no changes

Actual:
image

What's funny, is that the only white line that is not covered by ignore IS preserved,

belav:
closes #742

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/CSharpierIgnore.cs
+++ b/Src/CSharpier/SyntaxPrinter/CSharpierIgnore.cs
@@ -38,7 +38,7 @@ internal static class CSharpierIgnore
 
             if (printUnformatted)
             {
-                unFormattedCode.Append(PrintWithoutFormatting(node));
+                unFormattedCode.Append(PrintWithoutFormatting(node, context));
             }
             else
             {
@@ -54,8 +54,8 @@ internal static class CSharpierIgnore
         return statements;
     }
 
-    public static string PrintWithoutFormatting(SyntaxNode syntaxNode)
+    public static string PrintWithoutFormatting(SyntaxNode syntaxNode, FormattingContext context)
     {
-        return syntaxNode.GetText().ToString();
+        return syntaxNode.GetText().ToString().TrimEnd() + context.LineEnding;
     }
 }
--- a/Src/CSharpier/SyntaxPrinter/CSharpierIgnore.cs
+++ b/Src/CSharpier/SyntaxPrinter/CSharpierIgnore.cs
@@ -1,23 +1,17 @@
 namespace CSharpier.SyntaxPrinter;
 
+using System.Text;
+
 internal static class CSharpierIgnore
 {
     public static bool IsNodeIgnored(SyntaxNode syntaxNode)
     {
-        if (
-            syntaxNode.Parent
-            is not (
-                BaseTypeDeclarationSyntax
-                or BlockSyntax
-                or CompilationUnitSyntax
-                or NamespaceDeclarationSyntax
-            )
-        )
-        {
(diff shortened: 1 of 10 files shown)


Src/CSharpier/SyntaxPrinter/CSharpierIgnore.cs
```diff
@@ -26,29 +20,42 @@ FormattingContext context
...
         }
 
         return statements;
     }
 
-    public static Doc PrintWithoutFormatting(SyntaxNode syntaxNode)
+    public static string PrintWithoutFormatting(SyntaxNode syntaxNode)
     {
-        return syntaxNode.GetText().ToString().Trim();
+        return syntaxNode.GetText().ToString();
```
shocklateboy92, line 59: So, I think this means we won't even trim trailing whitespace for `// csharpier-ignore` blocks.
Is this intentional?

Src/CSharpier/SyntaxPrinter/CSharpierIgnore.cs
```diff
@@ -1,23 +1,17 @@
...
-            return false;
-        }
-
-        return (Token.HasLeadingComment(syntaxNode, "// csharpier-ignore"));
+        return syntaxNode.Parent
+                is BaseTypeDeclarationSyntax
+                    or BlockSyntax
+                    or CompilationUnitSyntax
+                    or NamespaceDeclarationSyntax
+            && Token.HasLeadingComment(syntaxNode, "// csharpier-ignore");
```
shocklateboy92, line 14: Do we want this to also work for 
```
/* csharpier-ignore */
```

shocklateboy92 (approved):

Src/CSharpier/SyntaxPrinter/CSharpierIgnore.cs
belav: Maybe only if someone complains? I feel like `//` is pretty standard for c# single line comments

Src/CSharpier/SyntaxPrinter/CSharpierIgnore.cs
belav: I think it used to unintentionally trim trailing whitespace, which also caused the problems with removing lines when it shouldn't. Or maybe I did intentionally make it do that? I didn't have a test for it so I'm not sure.
Either way now it trims trailing properly(with tests!), and fixes end of lines. But I don't see a good way to test the end of line part. The test file could have CRLF or LF depending on how the repo gets cloned.

shocklateboy92 (approved):
LGTM. Maybe your original idea of not even trimming trailing whitespace for ignored lines was a good idea. If there's some case where a user needs to have trailing whitespace, there's currently no way of doing that, right?

PR #747, opened 2022-10-24
Fix warning

belav:


shocklateboy92:
Finally! That warning was showing up on every pull request!

Side note: Did you want to enable `<TreatWarningsAsErrors>`?

belav:
> Side note: Did you want to enable `<TreatWarningsAsErrors>`?

@shocklateboy92 I have that set in `CSharpier.Build.props` so it isn't clear to me why this warning wasn't failing the build. I know you can `NoWarn` things. Maybe there is also a list of warnings somewhere that dotnet doesn't treat as errors even with `TreatWarningAsErrors`. At work I feel like we also have a lot of warnings that don't fail our build even though we also use `TreatWarningAsErrors`. Maybe I'll find some time to dig into it more soon.

PR #760, opened 2022-11-21
fixing nullable warnings

belav:


Src/CSharpier.Tests/FormattingTests/BaseTest.cs
```diff
@@ -13,16 +10,14 @@ namespace CSharpier.Tests.FormattingTests;
...
 
     [OneTimeSetUp]
     public void Setup()
     {
-        this.rootDirectory = new DirectoryInfo(Directory.GetCurrentDirectory());
-        while (this.rootDirectory.Name != "CSharpier.Tests")
-        {
-            this.rootDirectory = this.rootDirectory.Parent;
-        }
+        this.rootDirectory = DirectoryFinder.FindParent("CSharpier.Tests");
```
shocklateboy92, line 20: Since this is a static method, why not just call it in the field initializer?
That way, there's no need to suppress `CS8618`.

PR #704, opened 2022-07-15
Add node_modules as a default ignore

SubjectAlpha:
Add the ability to easily add more default ignores if wanted. As requested by #699

PR #762, opened 2022-11-23
Fix for subfolders in node_modules

belav:
relates to #704

### DIFF
--- a/Src/CSharpier.Cli/IgnoreFile.cs
+++ b/Src/CSharpier.Cli/IgnoreFile.cs
@@ -7,7 +7,7 @@ public class IgnoreFile
 {
     protected Ignore.Ignore Ignore { get; }
     protected string IgnoreBaseDirectoryPath { get; }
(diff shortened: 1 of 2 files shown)


Src/CSharpier.Cli/IgnoreFile.cs
```diff
@@ -7,7 +7,7 @@ public class IgnoreFile
 {
     protected Ignore.Ignore Ignore { get; }
     protected string IgnoreBaseDirectoryPath { get; }
-    private static string[] alwaysIgnored = { "**/node_modules/*.cs" };
+    private static readonly string[] alwaysIgnored = { "**/node_modules/**/*.cs" };
```
shocklateboy92, line 10: 🤣

shocklateboy92 (approved):
This does bring something to mind. Do we have a mechanism in config to override the always ignored list?

I understand it's unlikely anyone will want to format any C# code in their `node_modules` directory, but I'm thinking about the future where this list might grow..

PR #782, opened 2023-01-07
Put constructor initializer on new line

### LINKED ISSUE #526: Consider always putting constructor initializers on their own line
According to [this stylecop rule](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1128.md) constructor initializers should always be on their own line.

This probably helps in a case like this
```c#
public class RealTimeCallTimeoutException
{
    public RealTimeCallTimeoutException(string call) : base(call + " timed out") { }

    public RealTimeCallTimeoutException(string call)
        : base(call + " timed out") { }
}
```
But feels a bit silly in a case like this
```c#
public class GenInt
{
    public GenInt() : base(1) { }

    public GenInt()
        : base(1) { }
}
```

Running the numbers on the csharpier-repos code, 67% of initializers are on a new line. But that doesn't take into account how long the constructor or initializer are.

### LINKED ISSUE #700: Unable to configure tab usage
Hi there,

I have went through documentation to check how to configure tab usage in files and it was pretty straightforward so my configuration is like this:

1. created `.csharpierrc.json` 
2. added configuration properties shown in attachment.

With this configuration I am still unable to see tabs being used instead of spaces as seen in second picture (git diff).
This file is located at the root directory and it is being used locally by git hooks where tool is installed as global tool and in GitLab pipeline so we ensure configuration consistency among environments.

Thanks in advance for any help.
Kind regards! 
configFile
tabToSpaces

belav:
closes #526

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/ConstructorDeclarations.cst
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/ConstructorDeclarations.cst
@@ -23,9 +23,11 @@ public class BasicClass
 
 public class Initializers : BasicClass
 {
-    public Initializers() : this(true) { }
+    public Initializers()
+        : this(true) { }
 
-    public Initializers(string value) : base(value) { }
+    public Initializers(string value)
+        : base(value) { }
 
     public Initializers(
         string longParameter_______________________________________________________________________________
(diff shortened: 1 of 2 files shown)


shocklateboy92 (approved):
Did you link the wrong bug? I don't see how that's related to this change at all. 🤷

belav:
> Did you link the wrong bug? I don't see how that's related to this change at all. 🤷

Ah yeah, I updated the commit but missed updating the PR.

PR #786, opened 2023-01-09
Put generic constraints on new line

### LINKED ISSUE #527: Consider always putting generic type constraints onto a new line
According to [this stylecop rule](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1127.md) generic type constraints should always be on a new line.

I think I am on board with modifying this
```c#
    public static T CreatePipelineResult<T>(
        T result,
        ResultCode resultCode,
        SubCode subCode,
        string message = null
    ) where T : PipeResultBase
    {
// vs
    public static T CreatePipelineResult<T>(
        T result,
        ResultCode resultCode,
        SubCode subCode,
        string message = null
    ) 
        where T : PipeResultBase
    {
```

But it makes less sense in cases like this
```c#
    public int Foo<T>(T obj) where T : U;
// vs
    public int Foo<T>(T obj) 
        where T : U;
```

From the files in csharpier-repos, only 39% of generic type constraints are on a new line. But that doesn't take into account super short examples vs long ones.

belav:
closes #527

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/ClassDeclarations.cst
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/ClassDeclarations.cst
@@ -14,7 +14,8 @@ public class ThisIsSomeLongNameAndItShouldFormatWell1
         AndYetAnotherLongClassName,
         AndStillOneMore { }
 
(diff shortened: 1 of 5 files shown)


shocklateboy92 (approved):
LGTM.

PR #787, opened 2023-01-09
Only read files that are UTF8 or UTF8 BOM

### LINKED ISSUE #768: Error in MSBuild package when code contains an acute accent
Let's say I have this enum:

```csharp
public enum MeetingLocation
{
  Café,
  Restaurant
}
```

I can build the project just fine even though it has the [acute accent](https://en.wikipedia.org/wiki/Acute_accent#English) (`é`).

But if I add the CSharpier MSBuild package, I get an error on build:

```
  Error <FULL_FILE_PATH> - Failed to compile so was not formatted.
```

Of course the file does not get formatted, but it does not block formatting other files or fail the build itself.

Adding the comment `// csharpier-ignore` to the beginning of the file or the previous line does not seem to help.

This is tested agains CSharpier.MsBuild@0.20.0, on .NET SDK 6.0.403.

The VS Code extension does not seem to have a problem (yes, I checked the extension logs).

belav:
closes #768

Previously UTF.Unknown was used to try to determine file encodings. This was problematic because if a file was too small, it would end up with several possible encodings. The encoding it considered most likely was not always correct. As opposed to trying to read a file as each encoding type, CSharpier now only supports UTF8 files. This is consistent with the IDE plugins, which stream files to CSharpier as UTF8. If a file is UTF8-BOM, it will retain the BOM.

### DIFF
--- a/Src/CSharpier.Cli/FileReader.cs
+++ b/Src/CSharpier.Cli/FileReader.cs
@@ -28,7 +28,10 @@ internal static class FileReader
             // this is kind of a "hack" - read the file without the BOM then
             // streamReader.CurrentEncoding below will correctly be UTF8 or UTF8-BOM
             // https://stackoverflow.com/a/27976558
-            using var streamReader = new StreamReader(fileStream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false));
+            using var streamReader = new StreamReader(
+                fileStream,
+                new UTF8Encoding(encoderShouldEmitUTF8Identifier: false)
+            );
 
             var fileContents = await streamReader.ReadToEndAsync();
 
--- a/Src/CSharpier.Cli/FileReader.cs
+++ b/Src/CSharpier.Cli/FileReader.cs
@@ -1,6 +1,5 @@
 using System.IO.Abstractions;
 using System.Text;
-using UtfUnknown;
 
 namespace CSharpier.Cli;
 
--- a/Src/CSharpier.Cli/FileReader.cs
+++ b/Src/CSharpier.Cli/FileReader.cs
@@ -28,7 +28,7 @@ internal static class FileReader
             // this is kind of a "hack" - read the file without the BOM then
             // streamReader.CurrentEncoding below will correctly be UTF8 or UTF8-BOM
             // https://stackoverflow.com/a/27976558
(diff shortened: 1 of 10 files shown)


Src/CSharpier.Tests/FormattingTests/TestFiles/FileEncoding_UTF8.cst
```diff
@@ -2,3 +2,9 @@ public class ClassName
 {
     public void MethodName() { }
 }
+
+public enum MeetingLocation
+{
+    Café,
```
shocklateboy92, line 8: Uh... do you want to make the test case `Café` like the bug? `CafÃ©` kind of makes no sense.

Src/CSharpier.Cli/FileReader.cs
```diff
@@ -25,24 +24,19 @@ CancellationToken cancellationToken
...
-            fileStream.Seek(0, SeekOrigin.Begin);
-
-            // this fixes an issue with ANSI encoded files like csharpier-repos\AutoMapper\src\UnitTests\Internationalization.cs
-            var encodingToRead = encoding.CodePage == 852 ? Encoding.GetEncoding(1252) : encoding;
-
-            using var streamReader = new StreamReader(fileStream, encodingToRead);
+            // this is kind of a "hack" - read the file without the BOM then
+            // streamReader.CurrentEncoding below will correctly be UTF8 or UTF8-BOM
+            // https://stackoverflow.com/a/27976558
+            using var streamReader = new StreamReader(fileStream, new UTF8Encoding(false));
```
shocklateboy92, line 31: Do you want to do?
```suggestion
            using var streamReader = new StreamReader(fileStream, new UTF8Encoding(encoderShouldEmitUTF8Identifier: false));
```

Because `false` makes no sense on an encoding without reading the documentation.

shocklateboy92 (dismissed):

Src/CSharpier.Tests/FormattingTests/TestFiles/FileEncoding_UTF8.cst
belav: I'm thinking whatever you are looking at this code with isn't interpreting the file correctly? This is what I see
image

PR #788, opened 2023-01-09
Fixed Condition Error

samtrion:
`error  : Specified condition "$(CSharpier_Check)" evaluates to "" instead of a boolean.  C:\Users\<username>\.nuget\packages\csharpier.msbuild\0.21.0\build\CSharpier.MsBuild.targets`

This error makes it impossible to unload and reload projects in a Multi Project Solution.

### DIFF
--- a/Src/CSharpier.MsBuild/build/CSharpier.MsBuild.targets
+++ b/Src/CSharpier.MsBuild/build/CSharpier.MsBuild.targets
@@ -2,7 +2,7 @@
     <PropertyGroup>
         <CSharpier_FrameworkVersion Condition="'$(CSharpier_FrameworkVersion)' == ''">$(TargetFramework)</CSharpier_FrameworkVersion>
         <CSharpierDllPath>$(MSBuildThisFileDirectory)../tools/csharpier/$(CSharpier_FrameworkVersion)/dotnet-csharpier.dll</CSharpierDllPath>
-        <CSharpierArgs Condition="$(CSharpier_Check)">$(CSharpierArgs) --check</CSharpierArgs>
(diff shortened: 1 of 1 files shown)


Src/CSharpier.MsBuild/build/CSharpier.MsBuild.targets
```diff
@@ -2,7 +2,7 @@
     <PropertyGroup>
         <CSharpier_FrameworkVersion Condition="'$(CSharpier_FrameworkVersion)' == ''">$(TargetFramework)</CSharpier_FrameworkVersion>
         <CSharpierDllPath>$(MSBuildThisFileDirectory)../tools/csharpier/$(CSharpier_FrameworkVersion)/dotnet-csharpier.dll</CSharpierDllPath>
-        <CSharpierArgs Condition="$(CSharpier_Check)">$(CSharpierArgs) --check</CSharpierArgs>
+        <CSharpierArgs Condition="'$(CSharpier_Check)' = 'true'">$(CSharpierArgs) --check</CSharpierArgs>
```
belav, line 5: I'm thinking this should be an `==` instead? I've always used them.

belav (changes requested):
I ran into this once or twice when working on a different issue recently but wasn't sure how to reproduce it reliably and then forgot about it. Thanks for the contribution!

Src/CSharpier.MsBuild/build/CSharpier.MsBuild.targets
samtrion: Yes, you are right.
Error when transferring, had already fixed it locally.

shocklateboy92 (approved):

belav (approved):

PR #797, opened 2023-01-16
Checks compatible target framework

samtrion:
Checks if a compatible target framework is used. If not, .NET 7 is used.

This prevents error messages like the following.
```
C:\Users\<username>\.nuget\packages\csharpier.msbuild\0.22.0\build\../tools/csharpier/netstandard2.0/dotnet-csharpier.dll  C:\sources\Test1\src\Console1" exited with code 1.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 0.22.1
+## What's Changed
+#### Fix for CSharpier.MsBuild so it selects a compatible framework if the project does not target net6 or net7 [#797](https://github.com/belav/csharpier/pull/797)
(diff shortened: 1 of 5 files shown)


belav:
Ah crap, I forgot about this! I meant document that CSharpier_FrameworkVersion was required if your project wasn't targeting net6 or net7, but this is a better idea.
Although maybe it should default to net6, which is what 0.21.0 auto selected. But it's also pretty easy for someone to add this property and point to net6 (after I document that it exists, which I forgot to do).
I'm kinda leaning keep it at net7, but let me know your thoughts. I can push out a release for this tomorrow.

belav:
Ah, I take it back. I did document this, but still, auto selecting this is the better option and helps avoid any confusion for people updating.

samtrion:
I chose the solution because it is simply the most _compatible_ to seem. Since it also handles a possible misconfiguration of `CSharpier_FrameworkVersion`.

With the question `net6.0` or `net7.0` as default value, it was simply a decision for the performance of `net7.0`.

PR #798, opened 2023-01-17
Offer a pre-commit hook that runs the .NET tool

Kurt-von-Laven:
.NET pre-commit hooks rely on `dotnet pack`, which uses the .NET SDK version specified in `global.json`. Relax the .NET SDK version constraint from ~7.0.100 to ^7.0.100, maximizing compatibility.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "7.0.100"
+    "version": "7.0.100",
+    "rollForward": "latestFeature"
   }
 }
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+- id: csharpier
+  name: Run CSharpier on C# files
+  entry: dotnet tool run dotnet-csharpier
+  language: dotnet
+  types:
+    - c#
+  description: CSharpier is an opinionated C# formatter inspired by Prettier.
--- a/Docs/Pre-commit.md
+++ b/Docs/Pre-commit.md
@@ -5,8 +5,22 @@ hide_table_of_contents: true
 
 CSharpier can be used with a pre-commit hook to ensure that all staged files are formatted before being committed.
 
+## [pre-commit](https://pre-commit.com)
+
+Add the following to your `.pre-commit-config.yaml`:
+
+```yaml
+repos:
+    - repo: https://github.com/belav/csharpier
+      rev: 0.23.0 # Replace with the latest Git tag.
+      hooks:
+        - id: csharpier
+```
+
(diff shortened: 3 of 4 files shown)


shocklateboy92 (changes requested):
First of all, thanks for your contribution.

Relaxing the version constraint to maximize compatibility is an awesome change. :rocket: :guitar:  

The `.pre-commit-hooks.yaml` is a bit more icky.
We don't use pre commit hooks in this repo, and having a config file around that does nothing would be confusing.

If you want it to serve as a reference for other people to configure their pre-commit hooks, please move its content into our documentation. Otherwise, remove it.

Kurt-von-Laven:
Thank you for the prompt review! It sounds like you are talking about `.pre-commit-config.yaml`, which indeed belongs in downstream repositories. This pull request only pertains to `.pre-commit-hooks.yaml`, which resides upstream and does not configure a repository to use pre-commit hooks. Nor does it serve as a reference for downstream repositories to configure their pre-commit hooks, but rather implements CSharpier as a pre-commit hook for consumption by downstream repositories. Pre-commit maintains [a centralized list of a very small fraction of the repositories offering hooks](https://pre-commit.com/hooks.html). Downstream repositories can implement hooks locally as a workaround in cases where the upstream repository does not wish to support [pre-commit](https://pre-commit.com). This largely defeats the goal of the tool though (to the extent that no such hooks may be disseminated via the aforementioned list). When the hook is supported upstream, this enables pre-commit to automatically manage development dependencies for a wide variety of programming languages and tools, including .NET tools. In the case of CSharpier, a local hook may require setting up `.config/dotnet-tools.json` and an additional hook to run `dotnet tool restore`. There are some pitfalls to avoid if you don't want to push hassle onto the end developer that I won't get into, and everyone ends up iteratively figuring it out for themselves (or not realizing it's possible), so no two solutions are the same. On account of the convenience of being able to run thousands of different tools written in any of the most popular programming languages on any repository by adding four lines of YAML to a config file, pre-commit has spread like wildfire, reaching 9.6k stars since its initial release in 2017.

global.json
```diff
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "7.0.100"
+    "version": "7.0.100",
+    "rollForward": "latestFeature"
```
belav, line 4: I ran into issues with `rollForward` in combination with some of the source generators. I have since moved away from the problematic ones, so I think this should be safe now.

belav (approved):
Thanks for the explanation on pre-commit, it wasn't immediately obvious to me how it worked.

I don't have any objections to including it.

shocklateboy92 (approved):
@Kurt-von-Laven for the detailed explanation. Makes now!

PR #813, opened 2023-01-30
Detect extension running remotely and provide better install message

### LINKED ISSUE #789: VSCode CSharpier extension - "Configured as formatter but cannot format 'C#'-files
# Issue description
When saving a .cs file, the notification `VSCode CSharpier extension - "Configured as formatter but cannot format 'C#'-files` pops up in the bottom right.

# Issue causes/solutions
### Fix
1. Add the `./config` directory with the dotnet-tools (of CSharpier) to the project root.
2. Add `RUN dotnet tool restore` to the Containerfile.
### Cause
When rebuilding the DevContainer, the installed dotnet-tools are no longer installed, thus the extension cannot execute the CSharpier commands. Running `dotnet tool restore` is therefore needed in order to re-install the CSharpier commands when the container is rebuilt.
### Fix
1. Add `"editor.formatOnSaveMode": "file"` to the `settings.json` of VS Code.

E.g.:
```json
"[csharp]": {
	"editor.defaultFormatter": "csharpier.csharpier-vscode",
	"editor.formatOnSave": true,
	// Needed for `CSharpier` to work correctly
	"editor.formatOnSaveMode": "file"
}
```
### Cause
The extension only seems to work with `"editor.formatOnSaveMode": "file"`. The aforementioned error will be thrown when using `"editor.formatOnSaveMode": "modifications"` or `"editor.formatOnSaveMode": "modificationsIfAvailable"`. I suspect this may be due to the Roslyn analyzer needing the full contents of the file to parse correctly in order to be able to format?

# Issue goal
The goal of creating this issue is to create awareness for the possible causes of this error. In my opinion, it would be wise to add a `Possible issues` or similar section to the `Details` of the VS Code CSharpier extension, so users of the extension will be able to see which modifications they must make in order to be able to use this extension.

belav:
closes #789

### DIFF
--- a/Src/CSharpier.VSCode/README.md
+++ b/Src/CSharpier.VSCode/README.md
@@ -42,3 +42,12 @@ You can turn on format-on-save on a per-language basis by scoping the setting:
 ## Limitations
 
 Format Selection is not supported.
+
+Only `"editor.formatOnSaveMode" : "file"` is supported. If using other modes, you can set `file` by scoping the setting:
+```json
+"editor.formatOnSave": true,
+"editor.formatOnSaveMode": "modifications"
+"[csharp]": {
+    "editor.formatOnSaveMode": "file"
+}
+```
--- a/Src/CSharpier.VSCode/README.md
+++ b/Src/CSharpier.VSCode/README.md
@@ -39,6 +39,19 @@ You can turn on format-on-save on a per-language basis by scoping the setting:
 }
 ```
 
+### Devcontainers
+
+CSharpier supports DevContainers if it is installed as a local dotnet tool.
+```bash
+# if no .config/dotnet-tools.json file exists
+dotnet new tool-manifest
+
+# add csharpier to manifest
+dotnet tool install csharpier
+
+# rebuild container image
+```
+
 ## Limitations
 
 Format Selection is not supported.
--- a/Src/CSharpier.VSCode/src/InstallerService.ts
+++ b/Src/CSharpier.VSCode/src/InstallerService.ts
@@ -26,7 +26,7 @@ export class InstallerService {
         this.logger.info(this.extension.extensionKind);
 
         if (this.extension.extensionKind === vscode.ExtensionKind.Workspace) {
-            window.showErrorMessage("CSharpier must be installed as a dotnet tool on the local host using `dotnet tool install csharpier` and the devcontainer rebuilt to support formatting files.");
+            window.showErrorMessage("CSharpier must be installed as a local dotnet tool in ./.config/dotnet-tools.json using `dotnet tool install csharpier` and the devcontainer rebuilt to support formatting files.");
             return;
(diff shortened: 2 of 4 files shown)


Src/CSharpier.VSCode/README.md
```diff
@@ -39,6 +39,28 @@ You can turn on format-on-save on a per-language basis by scoping the setting:
...
 ## Limitations
 
 Format Selection is not supported.
+
+Only `"editor.formatOnSaveMode" : "file"` is supported. If using other modes, you can set `file` by scoping the setting:
+```json
+"editor.formatOnSave": true,
+"editor.formatOnSaveMode": "modifications"
+"[csharp]": {
+    "editor.formatOnSaveMode": "file"
```
shocklateboy92, line 64: Is there a way to specify in the extension manifest/code that format mode should always be file?

Src/CSharpier.VSCode/src/InstallerService.ts
```diff
@@ -11,22 +12,34 @@ export class InstallerService {
...
     }
 
     public displayInstallNeededMessage = (directoryThatContainsFile: string) => {
         this.logger.error("CSharpier was not found so files may not be formatted.");
 
-        const items = ["Install CSharpier Globally"];
+        this.logger.info(this.extension.extensionKind);
+
+        if (this.extension.extensionKind === vscode.ExtensionKind.Workspace) {
+            window.showErrorMessage("CSharpier must be installed as a dotnet tool on the local host using `dotnet tool install csharpier` and the devcontainer rebuilt to support formatting files.");
```
shocklateboy92, line 29: What exactly do you mean by 'local host'? Machine outside the dev container?

shocklateboy92 (commented):
Some of the messages are a little confusing.

Src/CSharpier.VSCode/src/InstallerService.ts
belav: Yeah. VSCode uses that verbiage. Not sure if just calling it "local machine" or "host" would make it more clear. Maybe local machine because no one really refers to a container as a machine, but a container can host other things.

Src/CSharpier.VSCode/README.md
belav: The extension only calls `registerDocumentFormattingEditProvider` but not `provideDocumentRangeFormattingEdits`. Which I think is essentially saying it only supports a whole file edit. Apparently setting `formatOnSaveMode` to `modifications` can still cause problems though.

Src/CSharpier.VSCode/src/InstallerService.ts
belav: another way to say it I guess, is there must be a dotnet tools manifest in the directory that includes csharpier. Hmmm. Maybe that would be more clear.

PR #814, opened 2023-01-30
Allow empty lines in query syntax

### LINKED ISSUE #754: Allow blank lines in query syntax
### Discussed in https://github.com/belav/csharpier/discussions/750

<div type='discussions-op-text'>

<sup>Originally posted by **TwentyFourMinutes** November  2, 2022</sup>
Currently when you have a linq syntax query it will produce something like the following:

```cs
var result = await (
    from post in dbContext.Posts
    where post.Id == 1
    select post
)
    .AsNoTracking()
    .FirstAsync();
```
which seems fine at first, however, these queries can get quite large fairly quickly. We currently manually format it with empty lines e.g.
```cs
var result = await (
    from post in dbContext.Posts
    join blog in dbContext.Blogs on post.BlogId equals blog.Id
    
    let count = dbContext.Posts.Count(p => p.Name == post.Name)
    
    where post.Id == 1
    select new 
    {
         Post = post,
         Blog = blog,
         SamePostNameCount = count
    }
)
    .AsNoTracking()
    .FirstAsync();
```
Anyhow, csharpier would strip out empty lines which can get very unreadable on even larger queries. 

Proposal:
Allow single blank lines between two statements, but truncate more than that down to 1. 

Do you have had any thoughts on this yet? Or on this rather small proposal?</div>

belav:
closes #754

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/QueryExpressions.cst
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/QueryExpressions.cst
@@ -73,6 +73,21 @@ class ClassName
                 CustCount____________________________ = g.Count()
             };
 
+        var complexQuerySupportsNewLines =
+            from c in customers
+
+            let d = c
+
+            where d != null
+
+            join c1 in customers1 on c1 equals c
+
+            group c by c.Country into g
+
+            orderby g.Count() ascending
+
+            select new { Country = g.Key, CustCount = g.Count() };
+
         var asyncQuery = await
             from c in customers
             from u in users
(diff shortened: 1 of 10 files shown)


Src/CSharpier.Tests/FormattingTests/TestFiles/QueryExpressions.cst
```diff
@@ -73,6 +73,21 @@ class ClassName
...
             };
 
+        var complexQuerySupportsNewLines =
+            from c in customers
+
+            let d = c
+
+            where d != null
+
+            join c1 in customers1 on c1 equals c
```
shocklateboy92, line 83: You should have some lines without spaces to test the negative case

shocklateboy92 (approved):

PR #825, opened 2023-02-06
Making compilation errors public for CSharpier.Core

### LINKED ISSUE #799: Make compile errors public when using CSharpier.Core
This is a suggestion from #783 

> I'd like to get the error using the API, but it seems this is not possible currently. (https://github.com/hofstadter-io/hof/blob/_dev/formatters/tools/csharpier/Program.cs)

> Looking at https://github.com/belav/csharpier/blob/master/Src/CSharpier/CodeFormatter.cs it seems the public methods only return the result.Code field. Could we add something like FormatWithErrors and FormatWithErrorsAsync that exposes the more complex return CodeFormatterResult?

belav:
closes #799


Src/CSharpier/CodeFormatter.cs
```diff
@@ -4,51 +4,50 @@
...
 namespace CSharpier;
 
-public class CodeFormatter
+public static class CodeFormatter
 {
-    public static LanguageVersion LanguageVersion = LanguageVersion.Preview;
+    internal static readonly LanguageVersion LanguageVersion = LanguageVersion.Preview;
 
-    public static string Format(string code, CodeFormatterOptions? options = null)
+#pragma warning disable RS0026 // I know these are breaking and I can't seem to get Unshipped/Shipped happy with me doing it
```
shocklateboy92, line 11: Why don't you just add a new set of overloads without the `options` arg? That way it doesn't have to be an optional parameter and might resolve this nicely.

shocklateboy92 (approved):
LGTM, mostly.

Src/CSharpier/CodeFormatter.cs
belav: It is complaining about the options but that isn't what changed. I changed the return type and it seems to be confused by that. Adding those overloads may get it happy, but then it may just complain again if I remove the overloads in the next release.

This seemed like the easiest way past this and I just need to remove the warning after 0.23.0 comes out.

shocklateboy92: I'll take your word for it :shrug:

shocklateboy92 (approved):

PR #831, opened 2023-02-18
deal with comments and #if's in argument lists

### LINKED ISSUE #410: When #if starts with comma, it formats oddly
In the given code, the `,` cannot be outside the `#if`. This leads to it being formatted like so.
Can we catch this and avoid breaking after the `,`?
```c#
            [Option(
                "server_port"
#if DEBUG
                ,
                Default = 50052
#endif
            )]
```

belav:
closes #410

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/ArgumentListLikeSyntax.cs
+++ b/Src/CSharpier/SyntaxPrinter/ArgumentListLikeSyntax.cs
@@ -12,14 +12,17 @@ internal static class ArgumentListLike
         var docs = new List<Doc> { Token.Print(openParenToken, context) };
 
         if (
-            arguments.Count == 1
-            && arguments[0].Expression
-                is ParenthesizedLambdaExpressionSyntax
-                    {
-                        ParameterList.Parameters.Count: 0,
-                        Block: { }
-                    }
-                    or SimpleLambdaExpressionSyntax { Block: { } }
+            arguments
+            is [
+                {
+                    Expression: ParenthesizedLambdaExpressionSyntax
+                        {
+                            ParameterList.Parameters.Count: 0,
+                            Block: { }
+                        }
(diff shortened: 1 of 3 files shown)


Src/CSharpier/SyntaxPrinter/ArgumentListLikeSyntax.cs
```diff
@@ -12,14 +12,17 @@ FormattingContext context
...
-                is ParenthesizedLambdaExpressionSyntax
-                    {
-                        ParameterList.Parameters.Count: 0,
-                        Block: { }
-                    }
-                    or SimpleLambdaExpressionSyntax { Block: { } }
+            arguments
+            is [
+                {
+                    Expression: ParenthesizedLambdaExpressionSyntax
```
shocklateboy92, line 18: oooh. Fancy list pattern match syntax :beach_umbrella:

shocklateboy92 (commented):
TBH, I'm not sure #831 should be a bug.
It's clear in this case that there's a comma that's in the `#if`.
But it doesn't really belong to the next parameter, and we've never absorbed that line break in any other circumstance. 

IMHO I think the current behavior is better than putting the comma on the same line as the next argument. What do you think?

Src/CSharpier/SyntaxPrinter/ArgumentListLikeSyntax.cs
belav: that was all rider...

belav:
I think I'm okay with just not doing this. I ran into it a couple times at work with our new multitargeting but it is super rare. I think there were ~30 instances of it across all of the code in https://github.com/belav/csharpier-repos. It's kinda ugly as is, but also weird with the preceding comma. So let's just not add extra code to deal with the super rare thing when the new formatting is debatable.

PR #858, opened 2023-03-27
Adding XML formatting

### LINKED ISSUE #819: Support formatting .csproj files and other types of xml
It could use the same indent and EOL settings.
Not sure about width, it could always not line break for now.
It would be nice for keeping csproj files consistent, I did that on a work project that was ~190 files and it was a pain, and things still get inconsistent.

belav:
closes #819


belav:
The first pass at this was to just use the `XmlTextWriter`, which fixes indentation to be consistent, but also destroys line breaks.

https://github.com/belav/csharpier-repos/pull/63/files?diff=split&w=1

A notable example
image

Maybe instead for a first pass this can
- Get indentation consistent
- Clean up extra new lines - maybe new lines separating elements should be allowed, but if there are multiples, then remove them. It seems pretty common for people to leave a blank line between `<ItemGroup>` elements.
- Get the end element tags consistent - `<Tag/>` should be `<Tag />`

I think that should be pretty easy to do using the existing `Doc` system.

Then the two remaining things I think are also needed

- A way to specify indentation options by file type. At work we use 2 spaces in csproj, but 4 spaces in cs.
- Maybe a way to specify which file types csharpier should format. I am leaning towards auto including cs, csproj, props and targets. The csharpierignore file is one way someone could opt out of formatting the none c# files. Maybe there should be an option instead? Something like
```yaml
# this would be the default value
filesToFormat: 
  - cs
  - csproj
  - props
  - targets

Do you have any thoughts @shocklateboy92 ?

**Comment by belav** [comment]
Actually, maybe more like what prettier does. Which can be combined with options for files, and at some point also allow custom formatters

```yaml
tabWidth: 4
overrides:
  - files: "*.csproj"
    options:
      tabWidth: 2
  - files: ["*.props", "*.targets"]
    formatter: none
  - files: "*.ps1"
    formatter: PluginThatFormatsPowershell

**Comment by gdlol** [comment]
😃 Hi there, I am using both csharpier and prettier in my projects and wondering if this would be some duplicate effort.

I have been using prettier for everything besides .cs files (like JSON files), and for .csproj (XML) files there is [@prettier/plugin-xml](https://www.npmjs.com/package/@prettier/plugin-xml).

The plugin uses [linguist-languages](https://www.npmjs.com/package/linguist-languages) to identify XML file types base on extensions, and it already includes file types like .csproj, .props and .targets. The list of extensions for XML can be found [here](https://github.com/github-linguist/linguist/blob/7ca3799b8b5f1acde1dd7a8dfb7ae849d3dfb4cd/lib/linguist/languages.yml#L7701). So far the experience is pretty good.

There may also be tooling concerns (VSCode settings) for including csproj formatting in csharpier. For instance, users may need to set csharpier as the default XML formatter, but is it a good idea?

Anyway, csharpier is a nice tool, and thanks for the effort to keep improving it.

**Comment by belav** [comment]
> 😃 Hi there, I am using both csharpier and prettier in my projects and wondering if this would be some duplicate effort.
> 
> I have been using prettier for everything besides .cs files (like JSON files), and for .csproj (XML) files there is [@prettier/plugin-xml](https://www.npmjs.com/package/@prettier/plugin-xml).

My initial pass for this was just intended to clean up whitespace and indentation, but I already ran into reasons why I think it needs to do a bit more than that and can probably use their code as a starting point. Thanks for sharing the link!

Even if prettier already supports this type of formatting I still see value in csharpier providing the functionality. There are situations where someone may not want to deal with node/npm and still be able to format xml. For a web app, they probably already have node but there are plenty of other types projects out there.

**Comment by belav** [comment]
@shocklateboy92 this is a bit too big to probably give a proper review, but it is mostly ready. If you have any feedback on the notes below I think that is more important than code review.

Things to do before actually releasing
- address a couple TODOs I still had in code
- clean up a lot of the commented out stuff from porting prettiers code for xml to c#
- write up all the doc for this
- maybe add some kind of xml validation to make sure csharpier isn't losing anything when it formats xml - I feel like I ran into a ton of edge cases in csharpier-repos so I don't know if it is worth the effort to do.

The most important bits

- adds formatting for xml files by default
- tries to format `".csproj", ".props", ".targets", ".xml", ".config"` as if they were xml
  - If one of them is not valid xml it is treated as a warning.
- defaults to 2 spaces for indentation instead of 4
- defaults to 100 width - I am tempted to make this higher. When reviewing all the files in csharpier-repos that ends up breaking a lot of elements with two attributes that look fine on a single line.
- does not break the starting tag for element with a single attribute
- does not break an element with no attributes and a single text child (with no new lines in it)
```xml
<!-- examples of the two above rules, they appear a lot in csprojs -->
<Project>
  <Reference Include="Newtonsoft.Json.Bson, Version=1.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
    <HintPath>..\..\packages\Newtonsoft.Json.Bson.1.0.2\lib\net45\Newtonsoft.Json.Bson.dll</HintPath>
  </Reference>
  <AspNetCoreShippingAssembly Include="Microsoft.AspNetCore.Server.Kestrel.Transport.NamedPipes.SomeLongerName" />
</Project>
```
- does not add/remove new lines in attributes. new lines are used often by csproj conditions to make them more readable, but officially xml spec calls for turning them into a single space
- treats whitespace as strict within a text element
```xml
<!-- given an element like this -->
<ElementWithAttribute Attribute="AttributeValue__________________">TextValue</ElementWithAttribute>
<!-- breaks this way to not add new lines -->
<ElementWithAttribute Attribute="AttributeValue__________________"
  >TextValue</ElementWithAttribute
>
<!-- vs treating it as non-strict and adding line breaks wihin the text -->
<ElementWithAttribute Attribute="AttributeValue__________________">
  TextValue
</ElementWithAttribute>

belav:
I'm going to merge this into 1.0.0 so I can start doing some proper testing with an alpha nuget package.

PR #863, opened 2023-04-01
Trying to force a blank line after file scoped namespace

belav:
This is actually kind of a pain because of #if's
Maybe this isn't worth the trouble. Or this takes priority, and the extra lines get added to #if's, because they aren't as common.


shocklateboy92:
> kind of a pain

Kind of a pain like "man... trying to solve this challenge really gets my blood pimping"?
Or more like "aw man, this is sucking all the joy out of this project"

If it's the latter, always remember the `help-wanted` label exists. :wink:

belav:
> > kind of a pain
> 
> Kind of a pain like "man... trying to solve this challenge really gets my blood pimping"? Or more like "aw man, this is sucking all the joy out of this project"
> 
> If it's the latter, always remember the `help-wanted` label exists. 😉

In this case the issue self reported by me so I could have just ignored it, but it was actually pretty easy to solve after stepping away for a bit. But I do really appreciate the advice for the "sucking the joy out of this project" types of issues.

shocklateboy92 (approved):

PR #889, opened 2023-05-06
Some cleanup

belav:


Src/CSharpier/SyntaxPrinter/FormattingContext.cs
```diff
@@ -1,9 +1,14 @@
...
+    public FormattingContext(string lineEnding)
+    {
+        this.LineEnding = lineEnding;
+    }
+
     public int PrintingDepth { get; set; }
     public bool NextTriviaNeedsLine { get; set; }
     public bool ShouldSkipNextLeadingTrivia { get; set; }
-    public string LineEnding { get; init; } = Environment.NewLine;
+    public string LineEnding { get; init; }
```
shocklateboy92, line 13: you know, you can use the new `required` keyword to avoid having to create a constructor:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required

shocklateboy92 (approved):

Src/CSharpier/SyntaxPrinter/FormattingContext.cs
belav: I tried that, but it was unhappy. I think because of the `netstandard2.0` target.

But I just ran into https://github.com/Sergio0694/PolySharp this morning which I want to try out, it sounds like it supports `required` and could replace my own version of Polyfilling some other language features.

PR #899, opened 2023-05-19
Fixing issue with the msbuild detection not dealing with CPM very well.

belav:


Src/CSharpier.Cli.Tests/CliTests.cs
```diff
@@ -315,6 +315,47 @@ public async Task Should_Not_Fail_On_Bad_Csproj()
         result.Output.Should().StartWith("Warning The csproj at ");
     }
 
+    [Test]
+    public async Task Should_Not_Fail_On_Mismatched_MSBuild_With_No_Check()
+    {
+        await this.WriteFileAsync(
+            "Test.csproj",
+            @"<Project Sdk=""Microsoft.NET.Sdk"">
```
shocklateboy92, line 323: Since the contents of these files are static, you should consider creating them as actual files and committing them.
Then you wouldn't need to escape all the ""`

shocklateboy92 (approved):
LGTM. Made a comment for you to consider in the future

PR #903, opened 2023-05-22
Getting csharpier.msbuild to run once even with multiple target frameworks

belav:


shocklateboy92 (dismissed):
Oaf. Reading this brought back the PTSD I got from dealing with MSBulid.
You're doing god's work :wink:

PR #910, opened 2023-06-10
Don't format files in obj folders

belav:


shocklateboy92 (approved):

shocklateboy92:
Out of curiosity, do we offer any sort of mechanism to opt out of the hard coded ignore paths?

belav:
> Out of curiosity, do we offer any sort of mechanism to opt out of the hard coded ignore paths?

It isn't documented, but I am fairly certain you can opt out in the same way as a gitignore. Meaning if the gitignore ignores bin, there is some way to do `!path/to/bin`, I don't know if that is the proper syntax or not.

Issue #911, opened 2023-06-11
Strong Name Assemblies


TwentyFourMinutes:
I am trying to use the `Csharpier.Core` package in a signed project. However I do get runtime exception for `Csharpier.Core` as it is not signed:
> Referenced assembly 'CSharpier, Version=0.24.2.0, Culture=neutral, PublicKeyToken=null' does not have a strong name.

In essence to fix this [create a `.snk` file](https://learn.microsoft.com/en-us/dotnet/standard/assembly/create-public-private-key-pair) and enter the following command in a `Developer Command Prompt for VS`:
> sn -k csharpier.snk 

Move the file to the same location as the `Directory.Build.props` and add the following inside the `Directory.Build.props` file:
```xml
<AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)/csharpier.snk</AssemblyOriginatorKeyFile>
<SignAssembly>true</SignAssembly>
```

Would be an easy fix and doesn't hurt anyone :)

belav:
I have this basically done but ran into https://www.nuget.org/packages/Ignore not being signed. It is a small library under MIT so I'll pull in the couple of files that I need to finish this off.

TwentyFourMinutes:
Oh, I see. If you want to offload any work, feel free to tell me. Thanks for looking into it though!

PR #913, opened 2023-06-13
Strong name sign assemblies

belav:


shocklateboy92:
Since you're pulling in the Ignore library, I would recommend against integrating the logic into our project.

Instead, create a top level directory called 'lib` or `External` or `Dependencies` or something.
When you put the ignore library in there, you should very clearly write down the version (or ideally, commit hash) that you're pulling from. You will definitely want to refer to this in future when taking updates of Ignore in future.
Maybe you can use git submodules, if the file structure of their repo doesn't need to change much.

That however, is a backup plan. My strongest suggestion is to have @TwentyFourMinutes (seems I can't tag him directly out side of issue #911 ) to follow up with the owner of the Ignore library and get them to sign their package. Then we can continue taking security fixes / updates from them via nuget, and allow ours to be used by signed only projects.

TwentyFourMinutes:
That's fine by me. 👍

TwentyFourMinutes:
https://github.com/goelhardik/ignore/issues/36

belav:
> [goelhardik/ignore#36](https://github.com/goelhardik/ignore/issues/36)

I assumed because the project hadn't been updated for a while it would take a while for it to be updated, the quick response on it was awesome to see!

TwentyFourMinutes:
You are god damn right, that's quite nice to see!

shocklateboy92 (approved):
LGTM

PR #915, opened 2023-06-17
Adding ability to bypass CSharpier

belav:


Src/CSharpier.MsBuild/build/CSharpier.MsBuild.targets
```diff
@@ -18,7 +18,7 @@
 
   <!-- getting this to run a single time for projects that target multiple frameworks requires all of this
    see https://github.com/dotnet/msbuild/issues/2781 -->
-  <Target Name="CSharpierFormat" BeforeTargets="DispatchToInnerBuilds;BeforeBuild">
+  <Target Name="CSharpierFormat" BeforeTargets="DispatchToInnerBuilds;BeforeBuild" Condition="'$(CSharpier_Bypass)' != 'true'">
```
shocklateboy92, line 21: Do we want to print a message for projects that have this enabled?
I feel like it's something that's easy to not notice coming from a `.props` that gets auto-injected from a nuget package and could potentially cause a massive headache to debug.

shocklateboy92 (approved):

PR #920, opened 2023-06-25
Reduce indentations for nested loops

### LINKED ISSUE #867: Nested loops without brackets may be not indented
```C#
// With CSharpier 23
foreach(var subsequnce in sequance)
    foreach(var item in subsequance)
        item.DoSomething();

// How it might look like 
// and how Rider 2023 formats it
foreach(var subsequnce in sequance)
foreach(var item in subsequance)
    item.DoSomething();
```

```C#
// With CSharpier 23
foreach(var subsequnce in sequance)
    foreach(var item in subsequance) {
        item.DoSomething();
        item.DoSomething();
    }

// How it might look like 
// and how Rider 2023 formats it
foreach(var subsequnce in sequance)
foreach(var item in subsequance) {
    item.DoSomething();
    item.DoSomething();
}
```

The same for `await foreach`, `while` and `for` loops;

The second variant removes one indentation level
and look pretty for me.

Also, CSharpier already use the second variant for `using` statements
and if the same will be done for loops it will be more consistent.

```C#
using (var fileStream1 = File.OpenRead("SomeFile"))
using (var fileStream2 = File.OpenRead("OtherFile"))
{
    Console.WriteLine("Hello");
    Console.WriteLine("World");
}
```

Rudomitori:
closes #867


belav:
A couple thoughts I had after looking at this change.

If the first `foreach` breaks, should the 2nd one still be indented? see [this](https://github.com/belav/csharpier-repos/pull/72/files#diff-4a1819bb1d2c943fab9ec0639f72bd2db4aa18c918374c049e2d69c3da17f177)
image


Are we sure mixing while/for/foreach should not keep the indentation? When I see two `foreach` at the same level it makes sense what is happening. `while` followed by `for` is less obvious, but perhaps because I'm not used to it. see [this](https://github.com/belav/csharpier-repos/pull/72/files#diff-28e20f07e126a9e668dc0b8a620fb76cb0a112f3e3c31187573e1314a46a750f)
image


I do like what it does to code like [this](https://github.com/belav/csharpier-repos/pull/72/files#diff-95df8d4451e7aaf14c727a9718e480c1d524b0b55e7c26492ba8719993ae5496). I've traditionally been "always braces" but with an auto formatter you don't have the problem of indentation being incorrect. And nested foreachs without the extra indentation is a lot cleaner.
image

shocklateboy92:
I definitely think this should only apply to nested loops of the same type. 

If that makes things more complicated to implement (since we have to keep track of what the previous node was), I think we should apply this behavior to `forceach` **only**.

Rudomitori:
> If the first `foreach` breaks, should the 2nd one still be indented?

I think, it will be better to keep more simple behavior and don't indent even if `foreach` breaks. It is more predictable and consistent.

> Are we sure mixing while/for/foreach should not keep the indentation? When I see two `foreach` at the same level it makes sense what is happening. `while` followed by `for` is less obvious, but perhaps because I'm not used to it.

`while` and `for` reflect the same intention as `foreach` does and the intention is to iterate over something. `foreach` can be used only if we deal with classes that implement `IEnumerable` or has `GetEnumerator` and if they are not implemented we need to fall back to `for` or even `while` if there is no indexed access. 

If we will keep the indentation for the case of the mixing loop types then we will force a developer to convert loops to one type.
* if the type is `foreach` then the developer need to implement `GetEnumerator`.
It is a good way to incapsulate the iteration logic, but the logic is not always the same for all usage of a class. Also `IEnumarator` is not a zero-cost abstruction and can be not applicable in performance sencitife code.
* `for` requires an indexed access that can be hard to implement or meaningless at all.
* `while` can be uncomfortable to use.

It will better to allow the developer to mix loop types in a way that is bast for his case and still has no unnecessary indentations.

But unfortunately, Rider keeps the indentation in this case and I didn't find out a option to change it. Also, Rider show a warning if there is no the indentation. 

I think, we have to don't conflict with Rider behavior and keep the indentation in this case for now, but try to remove it in the future.

shocklateboy92:
First of all, thank you very much for the detailed write-up - it's awesome when people are passionate about open source projects :+1: 

After taking your detailed points into consideration (making developers convert loop types, Rider rules, etc.), I think the best way forward _for now_ is this:
- Only reduce indentation for nested `foreach` loops without braces
- Don't change the behavior based on whether there's a `break` involved

We'll see how that performs in the wild, what people's feedback is, etc.
At _a later date_, we can decide to reduce indentation for other loops as well.

That way,
- Developers can easily opt-out of this behavior by putting braces (which is required per the style guide of a lot of projects anyway).
- It's simple to implement
- Doesn't cause anyone to convert loop types


I think it makes the most sense (even long term) because it covers the most common pattern: (iterating over nested iterators/collections). When someone sees a set of `foreach` statements on the same indentation level they can tell at a glance that someone is iterating over a nested collection. If the user chooses a different type of loop, that means they're doing something different and we should have the regular indented syntax to signal so.

Rudomitori:
I adjusted the code to reduce indentions only for the same loop type and did it for `while`, `for` and `foreach` loops. It was a very small change to do it not only for `foreach`

shocklateboy92 (approved):
Thanks for your contribution!

PR #923, opened 2023-07-01
Sorting usings

### LINKED ISSUE #661: Sorting of using directives
I would like using directives to be sorted with these rules:
- `System` on top.
- The rest sorted alphabetically.
- All `using static` at the end which follows the same rules.

Example:
```cs
using System;
using Bar;
using Baz;
using Foo;
using static System.Math;
using static Bar.BarHelper;
using static Foo.FooHelper;
```

belav:
closes #661


Src/CSharpier.Tests/FormattingTests/TestFiles/cs/UsingDirectives_FileComments.test
meenzen, line 1: I think there should be some test cases for code that does crazy things with comments, for example:

```cs
// Licensed to something

using Microsoft.AspNetCore;
// I like putting
using Microsoft;
/*
 * random comments everywhere
 */
using Apple;
/* because it's valid C# code */
using Microsoft.EntityFrameworkCore;
using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
// Look mom, I'm using a really fancy authentication library:
using Microsoft.Identity.Web.UI;
using Microsoft.Identity.Web; // another comment just for good measure
```

I'm unsure how this case should be handled, these are the options I could think of:

- Keep everything as is, whoever put a comment between the using statements probably has a good reason (Maybe show a warning when formatting? I am not familiar with this codebase so i'm not sure if showing warnings is even possible)
- Sort the individual blocks of usings between the comments
- Be opinionated and just remove the comments, lol

As far as I can tell tell it's impossible to sort usings with comments while preserving the original intent of the comments in all cases.

I don't really care what the final solution is, just that it's tested and works with all the possible comment formats.

belav: Keeping comments at the top was the only tricky part. Most comments are leading trivia which just goes with the node it is above. Trailing comments also stick with the node they are after. From all the code I reviewed, comments in usings aren't that common. `#if` is hard to deal with, but I have the basic cases covered. The .net analyzers themselves have some open issues around `#if`s, so I'm not worried about the couple of edge cases that I couldn't resolve which appeared in the mono code.

```
// Licensed to something

/*
 * random comments everywhere
 */
using Apple;
// I like putting
using Microsoft;
using Microsoft.AspNetCore;
/* because it's valid C# code */
using Microsoft.EntityFrameworkCore;
using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
using Microsoft.Identity.Web; // another comment just for good measure
// Look mom, I'm using a really fancy authentication library:
using Microsoft.Identity.Web.UI;

File: Src/CSharpier.Tests/FormattingTests/TestFiles/cs/UsingDirectives.test [hunk 1]
```diff
@@ -1,29 +1,13 @@
...
-using M = System.Math;
-using Point = (int x, int y);
-
-using static System.Math;
-
-global using System;
-
-using First;
-using Second;
+B;
```
shocklateboy92, line 5: wtf is this?

Src/CSharpier/SyntaxPrinter/UsingDirectives.cs
```diff
@@ -0,0 +1,259 @@
...
+                    systemUsings.Add(usingData);
+                }
+                else
+                {
+                    regularUsings.Add(usingData);
+                }
+            }
+        }
+
+        yield return globalUsings.OrderBy(o => o.Using!, Comparer).ToList();
```
shocklateboy92, line 201: Why do you need the null suppression operator for `o.Using!`? 
`.OrderBy()` usually doesn't need it.

Src/CSharpier/SyntaxPrinter/UsingDirectives.cs
```diff
@@ -0,0 +1,259 @@
...
+        yield return systemUsings.OrderBy(o => o.Using!, Comparer).ToList();
+        yield return aliasNameUsings.OrderBy(o => o.Using!, Comparer).ToList();
+        yield return regularUsings.OrderBy(o => o.Using!, Comparer).ToList();
```
shocklateboy92, line 204: I suspect if you just do `.OrderBy(o => "" + o.Using?.Alias.ToFullString() + o.Using?.Name.ToFullString())`, you won't need that comparer.
Though it's long enough that it might be worth extracting  to a helper method, and it might be slower because of the string concat, so :shrug:

shocklateboy92 (approved):
I hate the a `RawSyntaxKind` is an extension method and not a property. Otherwise we could use awesome pattern matches in so many cool places.

Src/CSharpier.Tests/FormattingTests/TestFiles/cs/UsingDirectives.test
belav: I believe that was me going nuts trying comments everywhere a while back.
Apparently that is valid c#!

Src/CSharpier/SyntaxPrinter/UsingDirectives.cs
belav: Looks like because I was missing the ? on `IComparer<UsingDirectiveSyntax?>`

PR #927, opened 2023-07-14
Adding support for editorconfig

### LINKED ISSUE #630: Csharpier .editorconfig
I'm trying to use both dotnet-format and csharpier on a project, as there are some fixes that dotnet format can provide that csharpier doesn't handle and vice versa. However, I'd like dotnet format to defer to csharpier's rules whenever there's any crossover.

Is it possible that we could get an .editorconfig file that describes csharpier's current rules as closely as possible?

belav:
closes #630


belav:
I still need to update the doc, but this is ready for review..... actually looks like there are some test failures. Possibly a linux vs windows path issue.
When in the same directory, csharpierrc takes precedence over editorconfig.
Otherwise if no configuration is in the same directory the first parent configuration (csharpierrc or editorconfig) will win.
Includes support for editorconfig inheritance (not sure if they call it that). Rules from parent files can apply if a child file doesn't supply a value for that rule.

belav:
@shocklateboy92 this was ready for review if you wanted to take a gander. It is a bigger change so I didn't want to skip over your watchful eye.

shocklateboy92 (commented):
I haven't read through the whole PR yet, but I figured the one comment I had so far was important enough to commit before going to bed tonight :slightly_smiling_face:

Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
```diff
@@ -51,13 +38,39 @@ ConfigurationFileOptions configurationFileOptions
...
-        ILogger? logger = null
+        ILogger logger
     )
     {
-        var directoryInfo = fileSystem.DirectoryInfo.FromDirectoryName(baseDirectoryPath);
+        var results = new List<CSharpierConfigData>();
+        var directoryInfo = fileSystem.DirectoryInfo.FromDirectoryName(directoryName);
+
+        var filesByDirectory = directoryInfo
+            .EnumerateFiles(".csharpierrc*", SearchOption.AllDirectories)
```
shocklateboy92, line 52: So, to confirm: you're looking for all `.csharpierrc` files in subdirectories, but you're only going to apply them to files that are in that directory (and below), right?

Because it would be weird if someone created a `.csharpierrc` (or `.editorconfig`) for a specific project in a monorepo and suddenly files in its parent directories started formatting that way.

shocklateboy92 (commented):
Okay, I actually did have time to finish the review tonight (_finally_).

Src/CSharpier.Cli/EditorConfig/ConfigFileParser.cs
```diff
@@ -0,0 +1,78 @@
+using System.IO.Abstractions;
+using System.Text.RegularExpressions;
+
+namespace CSharpier.Cli.EditorConfig;
+
+internal static class ConfigFileParser
+{
+    private static readonly Regex SectionRegex = new(@"^\s*\[(([^#;]|\\#|\\;)+)\]\s*([#;].*)?$");
```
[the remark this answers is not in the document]
belav: I was originally using that project, but it didn't support IFileSystem - https://github.com/editorconfig/editorconfig-core-net/pulls

I also realized it was built to find a editorconfig for a given file, which would be too slow for what csharpier needs.

That regex comes out of editorconfig-core-net, I did find https://github.com/rickyah/ini-parser just now. I will see if it is easy to pull in.

Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
belav: > you're only going to apply them to files that are in that directory (and below), right?

Correct. I didn't test it, but it seemed safe to assume that it would be faster to first find all possible configuration files in the directory tree being formatted. Then when formatting a given file, figure out which configuration applies to it.

belav:
Using ini-parser cleaned up the code and should do a much better job of parsing the files correctly.

Src/CSharpier.Cli/EditorConfig/ConfigFileParser.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        using var stream = fileSystem.File.OpenRead(filePath);
+        using var streamReader = new StreamReader(stream);
+        var configData = parser.ReadData(streamReader);
+
+        var directory = fileSystem.Path.GetDirectoryName(filePath);
+        var sections = new List<Section>();
+        foreach (var section in configData.Sections)
+        {
+            sections.Add(new Section(section, directory));
```
shocklateboy92, line 20: nit: I feel like this whole loop could have been a `.Select()`.

shocklateboy92 (approved):
minor nit: feel free to complete this PR yourself without waiting for another review if you choose to address it.

Src/CSharpier.Cli/EditorConfig/EditorConfigParser.cs
```diff
@@ -0,0 +1,84 @@
...
+        IFileSystem fileSystem
+    )
+    {
+        if (directoryName is "")
+        {
+            return new List<EditorConfigSections>();
+        }
+
+        var directoryInfo = fileSystem.DirectoryInfo.FromDirectoryName(directoryName);
+        var editorConfigFiles = directoryInfo
```
samtrion, line 19: Thus only a part of the `.editorconfig` of the project is considered. Since the project system / Visual Studio with [`<EditorConfigFiles />`](https://github.com/dailydevops/dotnet-engineering/blob/main/settings/sln.build.targets#L94-L95) allows to load further files. In addition Visual Studios extension the [`Global AnalyzerConfig`](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig).

belav: This comes from the doc on AnalyzerConfig, so I don't there need to be any changes to support it.
> global configuration files can't be used to configure editor style settings for IDEs, such as indent size or whether to trim trailing whitespace. Instead, they are designed purely for specifying project-level analyzer configuration options.

I wasn't aware of `EditorConfigFiles` but I'll look into it.

belav: It sounds like usage of `EditorConfigFiles` is not recommended and `Global AnalyzerConfig` should be used instead. https://github.com/MicrosoftDocs/visualstudio-docs/issues/6051#issuecomment-1092262347 Which makes it sounds like this is more intended to be used for configuring analyzers.

There is an easy workaround of just adding an `.editorconfig` or `.csahrpierrc` to the root, so it doesn't seem critical to support `EditorConfigFiles`.

PR #933, opened 2023-08-17
Adding format selection for VSCode plugin

### LINKED ISSUE #659: Format only the selected code

Exiled1:
Aims to resolve #659

Replaces the old languages registerDocumentFormattingEditProvider for registerDocumentRangeFormattingEditProvider to add range formatting support. Note: registerDocumentRangeFormattingEditProvider still acts as a document formatter so the old one is not needed.

Potential issues: Currently only accepts valid AST code, so selecting an entire local function doesn't work, however selecting the contents of the function should work the same since that makes a valid AST.


shocklateboy92:
Do you want to merge this as-is (behind an experimental config and a warning that depending on the selection, things will behave weirdly?
Or is this just to gather feedback on code style while you work on the actual implementation modifying the backend?

Exiled1:
> Do you want to merge this as-is (behind an experimental config and a warning that depending on the selection, things will behave weirdly? Or is this just to gather feedback on code style while you work on the actual implementation modifying the backend?

@shocklateboy92  I think this can be merged as-is currently. However it should be noted that I'm going to be implementing the options we talked about into the process for mapping AST nodes to their place pre and post formatting. The reason I bring this up is because I'll probably have to revisit this at a later date in an issue to update the format selection.

For the future:
- I'll be opening a new Issue to add in support for the change that I'll be doing with the process options since I think that the plugin will need to pass it over to the process.

shocklateboy92:
Implemented by #1376

PR #934, opened 2023-08-20
Avoid breaking only around binary expression but not binary expression itself

### LINKED ISSUE #924: Splitting previously split lines on logical separators if available
Splitting lines not solely on the width but also on width + logical separators makes the code more readable (subjectively).

## Example
Example code with logical checks, just exceeding a 100-character config limit.

Original Line, 103 char long. Over the 100-char config limit.
```C#
            else if (VariableNameOneAaaaaaaa > VariableNameTwoBbbbbb && VariableNameOneAaaaaaaa > 0.0)
```
When formatted ends up like the following, 97 char long at the widest, under the 100-char config limit.
```C#
            else if (
                VariableNameOneAaaaaaaa > VariableNameTwoBbbbbb && VariableNameOneAaaaaaaa > 0.0
            )
```
In my opinion, this would look better by either keeping it at 103 and not breaking or, preferably, breaking it further like so:
```C#
            else if (
                VariableNameOneAaaaaaaa > VariableNameTwoBbbbbb 
                && VariableNameOneAaaaaaaa > 0.0
            )
```

Currently, the last example will be formatted back to the middle example.

belav:
closes #924

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/IfStatements.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/IfStatements.test
@@ -56,22 +56,31 @@ public class ClassName
 
         if (true && (true || false)) { }
 
-        if (exactly100 || moreTextToMakeItWork_____________________________________________________)
+        if (someLongStatement == true || someOtherStatement_______________________________ == false)
         {
-            return;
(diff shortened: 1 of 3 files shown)


shocklateboy92 (approved):
LGTM

PR #940, opened 2023-08-23
Only wait 3 seconds for csharpier to format, otherwise kill it and restart it.

### LINKED ISSUE #926: Chsarpier freezes Rider
When I refactor Code, Csharpier freezes the UI sometimes.
Unfortunately it's impossible to reproduce reliably.
It mostly happens when I reformat code and the code is broken for a moment.

Can I find logs somewhere?

Usually killing the Csharpier Task via Taskmanager and then reenabling the Plugin unfreezes the Ui.

belav:
closes #926

### DIFF
--- a/Src/CSharpier.Rider/CHANGELOG.md
+++ b/Src/CSharpier.Rider/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # csharpier-rider Changelog
 
+## [1.3.9]
+- Wait at most 3 seconds for csharpier to format otherwise consider it hung and restart it.
+
 ## [1.3.8]
 - Add displayName attribute to CSharpier options window to speed up Settings dialog.
 
--- a/Src/CSharpier.Rider/README.md
+++ b/Src/CSharpier.Rider/README.md
@@ -11,7 +11,6 @@ To use it:
 - Optionally configure CSharpier to `Run on Save` under Preferences/Settings | Tools | CSharpier
 
 Please report any [issues](https://github.com/belav/csharpier/issues)
-<!-- Plugin description end -->
 
 ## Installation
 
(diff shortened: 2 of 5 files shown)


shocklateboy92 (approved):
Uh... do you know _why_ the csharpier process hangs? 
Are we using synchronous waits? deadlocks? 

Maybe it's worth some more investigation before jumping to a workaround?

belav:
> Uh... do you know _why_ the csharpier process hangs? Are we using synchronous waits? deadlocks?
> 
> Maybe it's worth some more investigation before jumping to a workaround?

We aren't really sure, and I've only seen it maybe once a week or so, with nothing in the logs to help figure out wtf is going on. The VS plugin has basically the same code to kill hung processes so it seemed safe enough. Maybe someone else that saw it more often could be convinced to fire up the plugin in debug, otherwise I'm not really sure how to track down the problem.

shocklateboy92:
> > Uh... do you know _why_ the csharpier process hangs? Are we using synchronous waits? deadlocks?
> > Maybe it's worth some more investigation before jumping to a workaround?
> 
> We aren't really sure, and I've only seen it maybe once a week or so, with nothing in the logs to help figure out wtf is going on. The VS plugin has basically the same code to kill hung processes so it seemed safe enough. Maybe someone else that saw it more often could be convinced to fire up the plugin in debug, otherwise I'm not really sure how to track down the problem.

So, logs probably won't help you in tracking down a deadlock. You have to take a memory dump when that happens and look at the stacks:
https://learn.microsoft.com/en-us/dotnet/core/diagnostics/debug-deadlock

belav:
I have a little bit of experience with dumps, but more around memory usage.

Based on the most recent comment on the freezing ticket, it could be that when a file is in just the wrong state, the streaming of data back and forth between the plugin and csharpier gets into a stuck state because one of the delimiters on either end never gets sent. It is probably about time to switch over to something like protobuf. I should also modify this to keep track of the file that fails when it hangs.

PR #944, opened 2023-08-26
Adding experimental support for GRPC for the extensions to communicate with CSharpier

belav:


shocklateboy92:
So, is there a reason you want GRPC _specifically_, other than a learning experience? 
If you create a server that listens on an TCP port, you have to deal with a fair bit of extra concurrency issues.
It would be much simpler if you just pipe the protocol buffers into stdin/stdout like you did with the plain text earlier.

And of course, since our proto file contains a single string, we might get away with just using JSON like other all the language servers seem to do.

PS: If the primary reason for using GRPC specifically _was_ for a learning experience, would you consider this PR as having given that experience? Or do you want to merge it into master and ship it?

belav:
> It would be much simpler if you just pipe the protocol buffers into stdin/stdout like you did with the plain text earlier.

This seems like it would require the same thing as CSharpier currently does, which is looking for a specific character/byte to know when a request to format a file has been received. I really just starting googling protobuf which lead me to grpc and the initial POC for just the c# side of things was super easy.

> If you create a server that listens on an TCP port, you have to deal with a fair bit of extra concurrency issues.

I'm hoping that all of these issues are hidden from me from using the grpc library. I also wanted to try out named pipes, which seems like the "proper" way to do IPC. But I don't have a clear answer on if it works with linux or not.

The code in the POC so far seems quite a bit simpler than running a process ourselves, I'm gonna play around a bit more because it is motivating right now. I am worried about what happens when you have multiple IDEs running, and you need to use a dynamic port or a dynamic named pipe. I was also having issues connecting to the server when running csharpier via a regular node child process which lead me to using vscode tasks, which may be the proper way to start up the csharpier process. Although it is unclear if you can pipe stdin/out with them.

TLDR - having fun learning for now, none of this may actually make it into the real code base.

belav:
The proto version doesn't do everything yet, but it does look way more straightforward right now

https://github.com/belav/csharpier/blob/proto/Src/CSharpier.Cli/ProtoFormatter.cs
vs
https://github.com/belav/csharpier/blob/proto/Src/CSharpier.Cli/PipingFormatter.cs

belav:
@shocklateboy92 I think this is ready to be considered experimental. It works well from what I can tell in VS & VSCode, although I don't plan on releasing updated extensions for them just yet.

It isn't clear why with VSCode using `spawn` results in not being able to connect to the GRPC server, I assume it is some kind of permissions issue. And using a task results in not being able to get anything from stdout which resulted it needing to find a free port on the VSCode side.

shocklateboy92:
Cheers. I'm kind of busy at the moment and this is a big PR. 

I'll get to it next week if that's alright.

belav:
> I'll get to it next week if that's alright.

Not a problem at all! I wasn't sure if you saw when PRs switched away from draft, and I knew you had some concerns with this one originally so wanted to make sure you saw it.

PR #950, opened 2023-08-31
Add support for ARM64 version of VS2022

ktos:
According to the [official docs](https://learn.microsoft.com/en-us/visualstudio/extensibility/arm64/target-arm64-visual-studio-extension?view=vs-2022) there should be only a need for a specific InstallationTarget to support ARM64 version of Visual Studio. 

I've added it and it seems to be working correctly in my experimental release.

I know that I am probably the only user of ARM version of VS2022, but still I'd really like to use CSharpier ;)


belav (approved):

belav:
I am not able to test this with ARM64, but I don't see any reason not to support it if it really is this easy. I released it as 1.4.4 which should be available shortly. Thanks for the contribution! You can't be the only one who would use it.

PR #958, opened 2023-09-10
Fix for the rider plugin hanging. When the error stream filled up without being read it couldn't read from the input stream was waiting indefinitely

belav:
closes 926


shocklateboy92 (approved):
LGTM

PR #960, opened 2023-09-10
Re-add support for line endings being configurable.

### LINKED ISSUE #935: Line endings not set
It seems that running `dotnet-csharpier .` on our project folder does not do anything about line endings. Files that have "lf" line endings keep them, files that have "crlf" keep them, too.

I'd expect that one of the two is chosen and applied.

Am I doing something wrong?

belav:
closes #935


Src/CSharpier.Cli/Options/CaseInsensitiveEnumConverter.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        if (reader.TokenType != JsonTokenType.String)
+        {
+            throw new JsonException(
+                $"Expected a string for enum parsing, but got {reader.TokenType}."
+            );
+        }
+
+        var enumText = reader.GetString();
+        if (Enum.TryParse(enumText, true, out TEnum result))
```
shocklateboy92, line 23: nit: booelan literal values are much more readable if you use named paramteters:
```suggestion
        if (Enum.TryParse(enumText, ignoreCase: true, out TEnum result))
```

shocklateboy92 (approved):

PR #962, opened 2023-09-22
fix: support EditorConfig comments

meenzen:
The spec very clearly states that comments are supported. Without this fix a basic config file like the example given on the EditorConfig website will cause a crash.

```
# EditorConfig is awesome: https://EditorConfig.org

# top-most EditorConfig file
root = true

# Unix-style newlines with a newline ending every file
[*]
end_of_line = lf
insert_final_newline = true

# ...
```

Spec: https://spec.editorconfig.org/#file-format

### DIFF
--- a/Src/CSharpier.Tests/OptionsProviderTests.cs
+++ b/Src/CSharpier.Tests/OptionsProviderTests.cs
@@ -215,6 +215,35 @@ end_of_line = crlf
         result.EndOfLine.Should().Be(EndOfLine.CRLF);
     }
 
+    [Test]
+    public async Task Should_Support_EditorConfig_With_Comments()
+    {
+        var context = new TestContext();
+        context.WhenAFileExists(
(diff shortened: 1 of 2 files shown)


Src/CSharpier.Tests/OptionsProviderTests.cs
```diff
@@ -215,6 +215,35 @@ public async Task Should_Support_EditorConfig_Basic()
...
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+max_line_length = 10
+; Windows-style line endings
```
belav, line 234: I never realized this was a valid comment.

belav (dismissed):
Thanks for catching this before it got released!

belav (approved):

PR #966, opened 2023-09-29
Support for c# 12 collection expressions

### LINKED ISSUE #964: Support C# 12 collection expressions
C# 12 [Collection expressions](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md) can't be formatted yet.

Example:

```cs
public class FooBar
{
    public void Foo()
    {
        List<string> list = ["Foo", "Bar"];
        Print(list);
    }

    public void Bar()
    {
        string[] array = ["Foo", "Bar"];
        Print(array);
    }

    private static void Print(IEnumerable<string> strings)
    {
        foreach (string str in strings)
        {
            Console.WriteLine(str);
        }
    }
}
```

Errors:

```
FooBar.cs - Failed to compile so was not formatted.
  (5,29): error CS1525: Invalid expression term '['
  (11,26): error CS1525: Invalid expression term '['
```

rel: #883

belav:
closes #964


belav:
It looks like I need to add code to ensure that syntax tree validation accounts for collection expressions.

.github/workflows/validate_pull_request.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/setup-dotnet@v1
         with:
           dotnet-version: |
-            7.0.304
+            7.0.400
```
shocklateboy92, line 14: On an unrelated note, I think you can the whole "default variables" feature to share this constant between all the different yaml files you have, so you only have to update it once in future:
https://docs.github.com/en/actions/learn-github-actions/variables#defining-configuration-variables-for-multiple-workflows

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM mcr.microsoft.com/dotnet/aspnet:7.0 AS base
 WORKDIR /app
 EXPOSE 80
 
-FROM mcr.microsoft.com/dotnet/sdk:7.0.304 AS build
+FROM mcr.microsoft.com/dotnet/sdk:7.0.400 AS build
```
shocklateboy92, line 5: if the whole variables thing I said in the other comment works out, you can make this an `ARG` and pass it from the yaml.

shocklateboy92 (approved):

shocklateboy92:
I'm going to merge this anyway, since the comments I made weren't directly related to this PR.

PR #986, opened 2023-11-07
Fix typo in Configuration.md

parched:


belav (approved):
Thanks!

belav:
I'm unsure why the check formatting action isn't passing, but the problem is definitely not related to this PR

PR #991, opened 2023-11-07
An existing csharpierrc file always takes priority over editorconfig

### LINKED ISSUE #987: EditorConfig hierarchy not working as expected
In the root directory I have `.csharperrc.yaml` which specifies the line length as 120.
Also, in the root directory I have `.editorconfig` which doesn't specify the line length.
In subdirectory `A` I have an `.editorconfig` which doesn't specify the line length either. (Note it also does not have `root = true`)

When I run `dotnet csharpier .` in the root directory, all the files in `A` get formatted with line length `100` (the default).

I expected the line length in the root `.csharpierrc.yaml` to be used, but it only does this when the `A/.editorconfig` doesn't exist.

I've now added `max_line_length = 120` to the root `.editorconfig` and it works, but this was a surprise when upgrading to 0.26.

belav:
closes #987

### DIFF
--- a/Docs/Configuration.md
+++ b/Docs/Configuration.md
@@ -7,8 +7,9 @@ CSharpier has support for a configuration file. You can use any of the following
 - A ```.csharpierrc.json``` or ```.csharpierrc.yaml``` file.
 - A ```.editorconfig``` file. See EditorConfig section below.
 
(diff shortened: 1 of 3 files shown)


shocklateboy92 (approved):
LGTM

PR #1004, opened 2023-11-10
Sort global System usings before other global usings

### LINKED ISSUE #1003: Global System using directives should be sorted first
First of all, great project and big fan of automating formatting.

I noticed an inconsistency when it comes to sorting regular `using System*` directives vs `global using System*` directives:

When sorting the regular `using` directives the System using directives are ordered before the non-system using directives.

However when sorting `global using` all the directives are grouped together and sorted as one set. 

Basically sorting the following:

```csharp
using Library;
using System;
using AnotherLibrary;
```

Results in 
```csharp
using System;
using AnotherLibrary;
using Library;
```

Where as sorting the following:

```csharp
global using Library;
global using System;
global using AnotherLibrary;
```

Results in 
```csharp
global using AnotherLibrary;
global using Library;
global using System;
```

My suggestion would be to use the same pattern with global System using directives so that they would be sorted before other non-System global using directives.

I don't mind doing the work if this sounds like a reasonable idea.

https://github.com/belav/csharpier/blob/a877f7d3b701fdd5a1a0f52a581d458897f6d570/Src/CSharpier/SyntaxPrinter/UsingDirectives.cs#L217-L223

vipentti:
Fixes #1003

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/UsingDirectives_SortsGlobalSystemToTop.expected.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/UsingDirectives_SortsGlobalSystemToTop.expected.test
@@ -1,0 +1,4 @@
+global using System;
+global using System.Web;
+global using AWord;
+global using ZWord;
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/UsingDirectives_SortsGlobalSystemToTop.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/UsingDirectives_SortsGlobalSystemToTop.test
@@ -1,0 +1,4 @@
+global using ZWord;
+global using AWord;
+global using System.Web;
+global using System;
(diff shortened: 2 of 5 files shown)


vipentti:
I'm not sure why either of the failed checks actually fail. 

`Check Formatting` seems to fail to something. Locally formatting checks out okay except for the expected after running other tests:

```
Error ./Src/CSharpier.Cli.Tests/bin/Debug/net8.0/TestFiles/CheckUnformatted.cs - Was not formatted.
  The file did not end with a single newline.
Formatted 302 files in 1254ms.
```

Without that 

```
dotnet csharpier . --check 
Formatted 301 files in 813ms.
```

`Run Tests` seems to fail due so some unrelated issue:

```
 Unhandled exception: System.AggregateException: One or more errors occurred. 
 (The process cannot access the file '/tmp/1fd44a6a-9e27-45bb-ad6c-bcaaf578e68d/project.assets.json' because it is being used by another process.)
 ```

shocklateboy92:
The tests are actually failing due to dotnet/sdk#35989

shocklateboy92:
@belav can you take a look / update to the latest .net RC? That should fix it.

belav:
> @belav can you take a look / update to the latest .net RC? That should fix it.

I've been getting past it for now by just running the action again, it usually works the second time. I can hopefully get things updated to the new net8 this weekend.

@vipentti thanks for the fix!

belav (approved):

Issue #1027, opened 2023-11-18
`CSharpier_FrameworkVersion` is unreliable


Tyrrrz:
I don't have .NET 7.0 SDK installed anymore, since I've moved my projects off to .NET 8.0. `CSharpier_FrameworkVersion` sometimes incorrectly resolves to .NET 7.0 due to unreliable heuristics:

https://github.com/belav/csharpier/blob/ef795385ba4b83a0b61a56287fd380cc63000dee/Src/CSharpier.MsBuild/build/CSharpier.MsBuild.targets#L3-L4

The two most common scenarios are:
- The project is a library that targets .NET Standard 2.1, but .NET 8.0 SDK is used to build it.
- The project is targeting a superset of .NET 8.0, for example `net8.0-windows`.

In either of these cases, CSharpier resolves to the .NET 7.0 version of the tool, which may not be able to execute (as in my case).

My suggestion is to not rely on the `TargetFramework` property for this, but instead use the current SDK version, which can be extracted from the `NETCoreSdkVersion` property, as defined by one of the built-in imports.

![image](https://github.com/belav/csharpier/assets/1935960/e4ce5a95-22b6-4b36-998f-ff79ce8c24e6)

domn1995:
I believe I'm running into this issue in my source generator project. Here's the build output if it helps:
https://github.com/domn1995/dunet/actions/runs/6900832597/job/18802795139?pr=187

belav:
> I believe I'm running into this issue in my source generator project. Here's the build output if it helps: https://github.com/domn1995/dunet/actions/runs/6900832597/job/18802795139?pr=187

That sounds like a problem with the nuget package not being restored correctly, I don't think I've seen that. I know one of the net8 RC's had some issues around restoring tools, but it looks like you are on the official release.

Issue #1022, opened 2023-11-16
CSharpier.MSBuild uses TargetFramework to choose which CSharpier tool to run which results in odd behavior when building multi-target libraries in Microsoft container images


adc-cjewett:
We use Microsoft's container images to build our code. CSharpier.MSBuild does not seem to play nicely with those images in more recent versions and the reason for that is it uses the TargetFramework to determine which version of the CSharpier tool is run. That seems incorrect because a machine with only .NET 8.0.100 SDK can build and package multi-target libraries. However, it cannot run .NET 6 or .NET 7 tools as the below error describes.

Would it make sense to expose a property that allows the users to override and specify which version of the CSharpier tool to be run? 

Reproducible here: https://github.com/adc-cjewett/CSharpierTest

```
------
 > [build 1/1] RUN dotnet build "CSharpierTest.csproj" -c Release --no-restore:
0.633 MSBuild version 17.8.3+195e7f5a3 for .NET
1.456   You must install or update .NET to run this application.
1.457
1.457   App: /root/.nuget/packages/csharpier.msbuild/0.26.2/tools/csharpier/net7.0/dotnet-csharpier.dll
1.457   Architecture: x64
1.457   Framework: 'Microsoft.NETCore.App', version '7.0.0' (x64)
1.457   .NET location: /usr/share/dotnet/
1.457
1.457   The following frameworks were found:
1.457     8.0.0 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]
1.457
1.457   Learn more:
1.457   https://aka.ms/dotnet/app-launch-failed
1.457
1.457   To install missing framework, download:
1.457   https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=7.0.0&arch=x64&rid=linux-x64&os=ubuntu.22.04
1.458 /root/.nuget/packages/csharpier.msbuild/0.26.2/build/CSharpier.MsBuild.targets(19,5): error MSB3073: The command "dotnet "/root/.nuget/packages/csharpier.msbuild/0.26.2/build/../tools/csharpier/net7.0/dotnet-csharpier.dll"  --check --no-msbuild-check "/CSharpierTest" > /dev/null " exited with code 150. [/CSharpierTest/CSharpierTest.csproj::TargetFramework=netstandard2.0]
1.719   You must install or update .NET to run this application.
1.720
[...]

Issue #1019, opened 2023-11-16
Improve logic for default target framework version for msbuild


belav:
The logic in `CSharpier.MsBuild.targets` will need to be updated every time a new .NET version comes up.

It should be possible to make use of something like this.
```
$([MSBuild]::VersionGreaterThanOrEquals($(_TargetFrameworkVersionWithoutV), '6.0'))
```

It will have to take into account the ability to override via `CSharpier_FrameworkVersion`

It is also not clear why this was required when adding net8 - `CSharpier_FrameworkVersion=$(CSharpier_FrameworkVersion)`. Figure out if there is a better way to deal with this all. Right now a lot of it feels like a kludge.

If possible get some more test scenarios. Right now the only scenario uses net8.

PR #1034, opened 2023-11-21
Use NETCoreSdkVersion to determine which CSharpier tool version to use during builds

Cjewett:
Relevant issues:
https://github.com/belav/csharpier/issues/1027
https://github.com/belav/csharpier/issues/1022
https://github.com/belav/csharpier/issues/1019

Instead of using the current TargetFramework that is being built as part of the MSBuild process we use the SDK version that the MSBuild process is using to determine which version of the CSharpier dotnet tool we should use.

The goal is to allow those who only keep one version of an SDK and runtime installed on their machine the ability to use CSharpier in a more reliable manner.


shocklateboy92 (approved):
LGTM. Thanks!

PR #1037, opened 2023-11-21
Fixing some issues with collection expressions

### LINKED ISSUE #1002: CSharpier doesn't respect editorconfig csharp_space_between_empty_square_brackets = false
### Expected
```csharp
        List<int> ids = [];
        var store = await Get<List<PositionAdd>>(Key) ?? [];
```

### Actual
**First line:** extra space before square bracket and space between empty square brackets
**Second line:** space between empty square brackets
```csharp
        List<int> ids =  [ ];
        var store = await Get<List<PositionAdd>>(Key) ?? [ ];
```

### LINKED ISSUE #1009: C#12 Collection expressions are prefixed with two spaces
I wanted to send some feedback about the formatting of the new collection expressions. I think it is inconsistent to prefix these with 2 spaces (I'm also not sure why such formatting was selected) when compared to how e.g. collection/object initializers are formatted:
```cs
public class Reproduction
{
    private readonly Dictionary<string, int> counts =  [ ];
    private readonly List<string> keys = { };
}
```

I'm also not sure I agree with the choice to format collection expressions with spaces between the square brackets and the elements but I suppose it's consistent with how collection/object initializers are formatted so I find it harder to argue against. It is worth noting that [MSDN formats these without spaces](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#collection-expressions) (while formatting collection/object initializers with spaces between the brackets) and the IDE also suggests such formatting though:
image

belav:
closes #1002
closes #1009

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/CollectionExpressions.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/CollectionExpressions.test
@@ -1,6 +1,8 @@
-int[] a =  [ 1, 2, 3, 4, 5, 6, 7, 8 ];
+List<int> ids = [];
 
-Span<int> b =  [ 'a', 'b', 'c', 'd', 'e', 'f', 'h', 'i' ];
+int[] a = [1, 2, 3, 4, 5, 6, 7, 8];
+
+Span<int> b = ['a', 'b', 'c', 'd', 'e', 'f', 'h', 'i'];
 
 string[] c =
 [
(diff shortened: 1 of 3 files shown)


Src/CSharpier.Tests/FormattingTests/TestFiles/cs/CollectionExpressions.test
```diff
@@ -1,6 +1,8 @@
-int[] a =  [ 1, 2, 3, 4, 5, 6, 7, 8 ];
```
belav, line 1: I had a test case with the two spaces here and didn't even notice
: (

shocklateboy92 (approved):

Src/CSharpier.Tests/FormattingTests/TestFiles/cs/CollectionExpressions.test
shocklateboy92: If it's any consolation, I didn't notice in the code review either.

PR #1030, opened 2023-11-20
Ignore .editorconfig file included in the .csharpierignore file

sebastieng84:
This will resolves issue #1023.

Any .editorconfig file included in the .csharpierignore file will not be parsed. By default we ignore everything in node_modules, but it still tries to parse the .editconfig file.

PR #1039, opened 2023-11-22
Optimize editorconfig lookups when piping files

belav:
This is gonna cause conflicts with #1030


belav:
I'd wait to merge this until after #1030, then I can just clean up the conflicts in here.

PR #1052, opened 2023-11-30
Sort usings with Ordinal culture (#1051)

### LINKED ISSUE #1051: Usings sorting differs based on system culture
It seems that system culture affects the usings ordering. I use `cs-CZ` culture on my PC, but our CI has `en-US`. This creates different lists in different environments.

For example this:

```csharp
using Microsoft.ApplicationInsights.Channel;
using Microsoft.ApplicationInsights.DataContracts;
using Microsoft.ApplicationInsights.Extensibility;
```

gets sorted like this in Czech:

```csharp
using Microsoft.ApplicationInsights.DataContracts;
using Microsoft.ApplicationInsights.Extensibility;
using Microsoft.ApplicationInsights.Channel;
```

This is because in Czech the `ch` is a "single letter" which is part of our alphabet and is placed between `h` and `i`.

When ordering the usings it should probably just always use something like `OrdinalIgnoreCase` (https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#recommendations-for-string-usage).

This started happening after upgrading from version `0.25.0` to `0.26.3` as part of our .net 8 upgrade.

davidkudera:
Fixes #1051.

I used `Ordinal` here just because it was what Rider suggested:

image

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/UsingDirectives.cs
+++ b/Src/CSharpier/SyntaxPrinter/UsingDirectives.cs
@@ -280,10 +280,18 @@ internal static class UsingDirectives
 
             if (x.Alias is not null && y.Alias is not null)
             {
-                return x.Alias.ToFullString().CompareTo(y.Alias.ToFullString());
+                return String.Compare(
+                    x.Alias.ToFullString(),
+                    y.Alias.ToFullString(),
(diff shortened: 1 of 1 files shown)


shocklateboy92 (approved):
Makes sense. Thanks!

PR #1065, opened 2023-12-09
Remove extra newline for collection expression property initializers

### LINKED ISSUE #1063: Extra newline added when using a collection expression with { get; }
I see some of the formatting issues with C# 12 collection expressions were fixed in #1047 but I think I have found another one.

I'm using version 0.26.4 and have found that the following code:

```csharp
public List<HttpRequestMessage> Requests { get; } =
    [
        new()
        {
            Content = new StringContent("content"),
            Headers = { { "header", "value" } },
            Method = HttpMethod.Post
        }
    ];
```

will get formatted to

```csharp
public List<HttpRequestMessage> Requests { get; } =

    [
        new()
        {
            Content = new StringContent("content"),
            Headers = { { "header", "value" } },
            Method = HttpMethod.Post
        }
    ];
```

i.e. it leaves a blank line between the property declaration and the value. This isn't consistent with the behaviour for fields or auto-properties (`=>`) which don't add a blank line so I assume it's unintentional.

SapiensAnatis:
Closes #1063 

Avoids adding a newline before a `CollectionExpressionSyntax` when its parent is an `EqualsValueClause` whose own parent is a `PropertyDeclarationSyntax`.

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/CollectionExpressions.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/CollectionExpressions.test
@@ -97,4 +97,15 @@ class MyClass
             HttpMethod.None,
             HttpMethod.Connect,
         ];
+
+    public List<DayOfWeek> DaysOfWeek { get; } =
+        [
+            DayOfWeek.Sunday,
+            DayOfWeek.Monday,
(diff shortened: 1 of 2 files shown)


belav (approved):
Looks good, seems that condition was causing the problem.

Issue #1077, opened 2023-12-18
Bugs caused by new formatting of lambda expressions


Rudomitori:
The new formatting implemented in #1066 have brought several new bugs.

### First
An empty argument list is splitted to several lines if the closing bracket is exactly the 100th character of the line

**Input:**
```C#
private static readonly ModelBinderProviderCollection _providers =
    CreateDefaultCollection();
```
**Output:**
```C#
private static readonly ModelBinderProviderCollection _providers = CreateDefaultCollection(

);
```
**Expected behavior:**
```C#
private static readonly ModelBinderProviderCollection _providers =
    CreateDefaultCollection();
```

### Second
In an argument list with one lambda the closing bracket is moved to the next line if it's exactly the 100th character of the line

**Input:**
```C#
                            It.Is<ExceptionLoggerContext>(
                                c => c.ExceptionContext == expectedContext
                            )
```
**Output:**
```C#
                            It.Is<ExceptionLoggerContext>(c => c.ExceptionContext == expectedContext
                            )
```
**Expected behavior:**
```C#
                            It.Is<ExceptionLoggerContext>(c =>
                                c.ExceptionContext == expectedContext
                            )
```

### Third
An argument list with a lambda is incorecctly formatted if the lambda contains other nested argument list with a lambda

**Input:**
```C#
        this.Where______________________________________(
            selector: static longName_________________ =>
            {
                return Method(x => x.Prop);
            }
        );
```
**Output:**
```C#
        this.Where______________________________________(
            selector: static longName_________________ =>
            {
                return Method(x => x.Prop);
            });
```
**Expected behavior:**
```C#
        this.Where______________________________________(
            selector: static longName_________________ =>
            {
                return Method(x => x.Prop);
            }
        );
```

Rudomitori:
@belav
I've fixed the first and third buhs, but I need more time to fix the second

belav:
@Rudomitori thanks for taking care of all of these!

PR #1066, opened 2023-12-10
feat: Improve formatting of lambda expressions

Rudomitori:
#766
#836
#672

PR #1078, opened 2023-12-18
fix: bugs from #1066

Rudomitori:
Fix the first and third bugs described in #1077

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/ArgumentListLikeSyntax.cs
+++ b/Src/CSharpier/SyntaxPrinter/ArgumentListLikeSyntax.cs
@@ -9,14 +9,12 @@ internal static class ArgumentListLike
         FormattingContext context
     )
     {
-        var docs = new List<Doc> { Token.Print(openParenToken, context) };
         var lambdaId = Guid.NewGuid();
 
-        switch (arguments)
+        var args = arguments switch
         {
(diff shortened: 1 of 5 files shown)


shocklateboy92 (dismissed):
LGTM.

Do you want to merge this as is, and fix the second bug mentioned in #1077 in a separate PR?

Or should I wait for you to push the fix for the second one also into this branch?

Src/CSharpier/SyntaxPrinter/ArgumentListLikeSyntax.cs
```diff
@@ -10,14 +10,15 @@ FormattingContext context
     )
     {
         var docs = new List<Doc> { Token.Print(openParenToken, context) };
+        var lambdaId = Guid.NewGuid();
 
         switch (arguments)
```
shocklateboy92, line 15: I forgot to mention this in your original PR, but this all branches of this switch statement just add something to the list.
Wouldn't it look better if you used the switch expression instead?

Rudomitori: The expression looks almost the same for me

Rudomitori:
I will fix the second bug in other PR

Rudomitori:
Well, I've fixed the last bug before PR is merged, so I've decided to add the changes in this PR

shocklateboy92 (approved):
Cheers!

PR #1080, opened 2023-12-19
refactor: Use switches instead of if-else in DocFitter

Rudomitori:
While I was working on the fixes for #1077, I needed to debug `DocFitter`. There is the long chained if-else and when you press the `Step over` button in RIder the debugger stops on a condition of each branch, but in a switch statement the debugger jumps right to the executed branch and it speed up debugging.

I don't mind if the PR will be rejected. It contains only syntax changes and doesn't modify the logic of `DocFitter`

### DIFF
--- a/Src/CSharpier/DocPrinter/DocFitter.cs
+++ b/Src/CSharpier/DocPrinter/DocFitter.cs
@@ -30,18 +30,11 @@ internal static class DocFitter
                 return false;
             }
 
-            PrintCommand command;
-            if (newCommands.Count > 0)
+            var (currentIndent, currentMode, currentDoc) = newCommands switch
             {
-                command = newCommands.Pop();
-            }
-            else
-            {
-                command = remainingCommands.ElementAt(x);
-                x++;
-            }
-
-            var (currentIndent, currentMode, currentDoc) = command;
+                { Count: > 0 } => newCommands.Pop(),
+                _ => remainingCommands.ElementAt(x++)
+            };
 
             switch (currentDoc)
             {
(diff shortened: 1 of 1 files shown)


Src/CSharpier/DocPrinter/DocFitter.cs
```diff
@@ -116,8 +111,9 @@ void Push(Doc doc, PrintMode printMode, Indent indent)
                             : ifBreak.FlatContents;
 
                     Push(contents, currentMode, currentIndent);
+                    break;
                 }
-                else if (currentDoc is LineDoc line)
+                case LineDoc line:
                 {
```
shocklateboy92, line 117: Do you need `{}` on all these cases?
Switch cases normally don't have them, so someone reading the code will think you're purposefully creating a scope because you're creating new variables that will conflict with the parent scope.

shocklateboy92 (commented):
I'm okay with this. There's nothing inherently wrong with using a switch-case, and if that helps improve debugging that's a good enough reason.

I did have a comment about your braces though.

Src/CSharpier/DocPrinter/DocFitter.cs
Rudomitori: I'm just used to adding the braces to switch branches if there are several statements in them. No matter whether there is variable declaration or not. 

I've removed them

shocklateboy92 (approved):
That was quick! Thanks

Src/CSharpier/DocPrinter/DocFitter.cs
```diff
@@ -30,64 +30,46 @@ void Push(Doc doc, PrintMode printMode, Indent indent)
...
+                        continue;
+
+                    if (returnFalseIfMoreStringsFound)
+                        return false;
+
+                    output.Append(stringDoc.Value);
+                    remainingWidth -= stringDoc.Value.GetPrintedWidth();
+                    break;
+                case LeadingComment
+                or TrailingComment:
```
belav, line 55: I created #1084 because this looks a bit odd.

PR #1088, opened 2023-12-24
Fix tests for environments with another culture

stefannikolei:
* When the tests run in an environment with e.g. a German operating system "expected" is translated, so this test fails.


shocklateboy92 (approved):
Seems legit. Thanks!

PR #1090, opened 2023-12-24
Update System.IO.Abstractions to latest

stefannikolei:


Src/CSharpier.Cli/CommandLineFormatter.cs
```diff
@@ -155,7 +155,7 @@ CancellationToken cancellationToken
                 : directoryOrFilePath;
 
             var optionsProvider = await OptionsProvider.Create(
-                directoryName,
+                directoryName!,
```
shocklateboy92, line 158: I really don't like the null forgiving `!` operator.
Would it be feasible to put null checks in all these instead?

stefannikolei: Yeah i can do that.

shocklateboy92 (dismissed):

shocklateboy92:
Oh, I just realized there are conflicts that need to be resolved. The code is good otherwise.

belav (approved):

PR #1091, opened 2023-12-24
Update Nunit to latest

stefannikolei:


shocklateboy92 (approved):
Heh, it's like we have a smarter dependabot :stuck_out_tongue:

PR #1098, opened 2023-12-25
Update DiffEngine to latest

stefannikolei:


Src/CSharpier.Tests/FormattingTests/SetUpFixture.cs
```diff
@@ -10,16 +10,7 @@ public class SetUpFixture
...
-        DiffTools.AddToolBasedOn(
-            DiffTool.WinMerge,
-            name: "WritableWinMerge",
-            arguments: (temp, target) =>
-            {
-                var leftTitle = Path.GetFileName(temp);
-                var rightTitle = Path.GetFileName(target);
-                return $"/u /wr /e \"{temp}\" \"{target}\" /dl \"{leftTitle}\" /dr \"{rightTitle}\"";
-            }
-        );
```
belav, line 23: This is used to make `[TestName].test` the writable file and is still needed.

If you add an extra space to `ClassDeclarations.test` and run the test, you want to be able to copy the changes from `ClassDeclarations.actual.test` back to `ClassDeclarations.test`.

belav (changes requested):

Src/CSharpier.Tests/FormattingTests/SetUpFixture.cs
stefannikolei: ok. I will have a look how to port this arguments to this version

stefannikolei: @belav I tried to achieve this. But on my Mac I could not get the behavior to work. I looked at the changes of DiffEngine and tried to port the arguments. Could you have a look on Windows with an installed version of WinMerge?

belav: This still just has the right side editable.

Is there any need to update? If not it doesn't seem worth the effort to figure out how to make this work on the latest version of DiffEngine.

PR #1104, opened 2023-12-26
Adding an option to format generated files that are typically excluded.

### LINKED ISSUE #1055: Add option to allow formatting auto generated files.
There are cases where code is auto generated outside of c# and a user will want to run csharpier to format them.

Adding a CLI option `--include-generated` to bypass the check for if a file is auto-generated will allow them to format those files.

Some notes

1. Are there generated files in the `obj` folder that would get accidentally included by adding this option?
2. If a user asks csharpier to format a single file, should that assume they want to format the file even if it is generated?

belav:
closes #1055


shocklateboy92 (approved):
LGTM.

PR #1115, opened 2024-01-07
Make long parenthesized expressions not always break.

### LINKED ISSUE #921: ParenthesizedExpression breaking inconsistently
Having a single invocation on a ParenthesizedExpression  that is too long causes it to break like below. It would look better if it stayed consistent with the second example.

```c#
(
    someObject as SomeLongType__________________________________________
).CallMethod();

(someObject as SomeLongType__________________________________________)
    .CallMethod()
    .CallMethod();
```

A potential solution is "never break ParenthesizedExpression unless its contents break"

belav:
closes #921

The goal of this ticket was to prefer breaking trailing members on a parenthesized expression before breaking the expression.
For example.
```c#
// input
(someObject as SomeLongType__________________________________________).CallMethod();

// expected
(someObject as SomeLongType__________________________________________)
    .CallMethod();

// actual
(
    someObject as SomeLongType__________________________________________
).CallMethod();
```

The way it is currently implemented though results in the following
```c#
var asyncMethodsWithToken = (
    from method in asyncMethods
    where method.GetParameters().Any(pi => pi.ParameterType == typeof(CancellationToken))
    select method
)
    .ToList();

// previously
var asyncMethodsWithToken = (
    from method in asyncMethods
    where method.GetParameters().Any(pi => pi.ParameterType == typeof(CancellationToken))
    select method
).ToList();
```

This does mean things are more consistent with longer chains
```c#
var asyncMethodsWithToken = (
    from method in asyncMethods
    where method.GetParameters().Any(pi => pi.ParameterType == typeof(CancellationToken))
    select method
)
    .ToList________________________()
    .ToList________________________()
    .ToList________________________();
```

Another example where I think the old version was probably better
```c#
Assert.Equal(
    CoreStrings.FindValueTypeMismatch(0, "IntKey", "string", "int"),
    (
        await Assert.ThrowsAsync<ArgumentException>(
            () =>
                Finder
                    .FindAsync<IntKey>(cancellationType, context, new object[] { "77" })
                    .AsTask()
        )
    )
        .Message
);

// previously
Assert.Equal(
    CoreStrings.FindValueTypeMismatch(0, "IntKey", "string", "int"),
    (
        await Assert.ThrowsAsync<ArgumentException>(
            () =>
                Finder
                    .FindAsync<IntKey>(cancellationType, context, new object[] { "77" })
                    .AsTask()
        )
    ).Message
);
```

I'm not sure how much work it would be to keep the tailing call next to the `)` if the contents of the parenthsized expression break.

FWIW prettier does not have any logic around this.

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/ParenthesizedExpressions.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/ParenthesizedExpressions.test
@@ -1,16 +1,34 @@
-class ClassName
-{
-    void MethodName()
-    {
-        var x =
-            query
-            && (
-                currentChar == '='
-                || currentChar == '<'
-                || currentChar == '!'
-                || currentChar == '>'
-                || currentChar == '|'
-                || currentChar == '&'
-            );
-    }
-}
+var x =
+    query
+    && (
+        currentChar == '='
+        || currentChar == '<'
+        || currentChar == '!'
+        || currentChar == '>'
+        || currentChar == '|'
+        || currentChar == '&'
+    );
+
+// some comment won't break the next line
+(someObject as SomeType).CallMethod();
+
+(someObject as Exactly100____________________________________________________________).CallMethod();
+(someObject as SomeLongType___________________________________________________________)
+    .CallMethod();
+(someObject as SomeLongType________________________________________________________________________)
+    .CallMethod();
+
+(
+    someObject
+    as UnlikelyButThisIsWhatItWouldLookLike_________________________________________________
+)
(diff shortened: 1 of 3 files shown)


shocklateboy92:
Do we have a group type that breaks if its children are longer than its siblings' children?

belav:
> Do we have a group type that breaks if its children are longer than its siblings' children?

No and the way things work I don't know if it would be possibly.

A ConditionalGroup may accomplish something like that. If the first possible group breaks the children and does not break the siblings and is considered to fit, then it gets printed.

shocklateboy92 (approved):

PR #1137, opened 2024-01-19
GRPC really bloated the nuget package, this accomplishes the same thing.

belav:
Nuget package went from 18mb to 70mb when the gRPC packages were included. This is 20mb as a package.

This should accomplish the same thing without all the bloat and the code is mostly the same.


Src/CSharpier.Cli/ServerFormatter.cs
```diff
@@ -1,39 +1,46 @@
...
-        };
-        server.Start();
-
+        var builder = WebApplication.CreateBuilder();
+        builder
+            .WebHost
+            .ConfigureKestrel(
+                (_, serverOptions) =>
+                {
+                    serverOptions.Listen(IPAddress.Any, thePort);
```
shocklateboy92, line 27: So, you can actually use [unix sockets](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions.listenunixsocket?view=aspnetcore-8.0) (Linux, possibly Mac) and [named pipes](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.core.kestrelserveroptions.listennamedpipe?view=aspnetcore-8.0) (Windows) to avoid having to chose a port.

shocklateboy92: That will make it much easier if you ever plan on having multiple editors share an instance of csharpier.

shocklateboy92 (approved):
So, does that mean you stopped using protocol buffers, but still use the `.proto` files to generate models across languages? If so, OpenAPI (formally swagger) might be alternative to consider.

Src/CSharpier.Cli/ServerFormatter.cs
belav: The proto models are no longer needed, I probably forgot to delete them.

I took a look at both a bit. One concern was windows/linux/mac + java/js/c# is a lot of combinations of things to figure out.

Also with named pipes/unix sockets, I assume I still need to come up with a unique name for each instance of CSharpier that runs. If you are running rider with 0.27 and vscode with 0.27 right now that is two instances of csharpier. I assume they would be unhappy all trying to share the same socket/pipe.

shocklateboy92: Yes. With UNIX sockets, you'd be using (probably):
```bash
"${XDG_RUNTIME_DIR}/csharpier/${version}.sock"
```

Which (in my machine) expanded to:
```
/run/user/1000/csharpier/0.27.2.sock
```

I'm not sure what the conventions are for windows named pipes, but I assume a similar thing is possible.

PR #1138, opened 2024-01-21
Change check to make sure versions with hash pass

alanssitis:
Method validateInstall in CustomPathInstaller for CSharpier.Rider would fail if the version passed has a hash.

I changed the method to ensure the plugin would still work if the version is outlined in the dotnet-tools.json file.


belav (changes requested):
I hadn't seen the hash version show up in `dotnet-tools.json`, but it does appear to accept the hash version.

How does one end up with the hash there?

Either way it does make sense to account for it, but I think it would be better to modify `CSharpierProcessProvider` to remove the hash when it pulls the version out of the `dotnet-tools.json`. Then the version string getting passed around should always be the bare version.

alanssitis:
That makes sense; I copied the whole version itself, hehe.

Rudomitori:
The hash is added by .NET 8 SDK.

It adds a Git commit hash to the version of the assembly on build. It's needed for [Source links](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink).
As I understand, it's mostly for libraries, but not for tools and apps. 
In our projects, we remove the hash by putting the following lines into csproj files:
```XML
<!-- .NET 8 SDK adds Git commit hash to the version, but we expect that version is exactly that we pass -->
<IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
```

shocklateboy92 (approved):

shocklateboy92:
Looks like @belav needs to re-approve. I'll wait for that instead of dismissing his review.

belav:
> ```
> <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
> ```

Doh, I wish I had done more googling before modifying all of the extensions to work with the hash version. We are kind of stuck with it for a while, but I'm going this so it can go away eventually.

belav (approved):

belav:
I submitted the new version 1.5.2 to JetBrains, it may take a day or two for them to approve .t

PR #1142, opened 2024-01-23
Also format csx files

### LINKED ISSUE #1141: Support for CSharp Script
Hi guys. I'm having some issues trying to format my C# scripts. 
C# can be used for scripting. There are tools like [dotnet-script](https://github.com/dotnet-script/dotnet-script/) to run those files. C# scripts use the `.csx` extension but **csharper** doesn't support it. Is there a way to force **csharper** to format those files?

**Input:**
 dotnet-csharpier ./main.csx

**Output:**
Warning ./main.csx - Is an unsupported file type.

**Expected behavior:**
The file formatted.

belav:
closes #1141

### DIFF
--- a/Src/CSharpier.Cli/CommandLineFormatter.cs
+++ b/Src/CSharpier.Cli/CommandLineFormatter.cs
@@ -240,9 +240,10 @@ internal static class CommandLineFormatter
                 var tasks = fileSystem
                     .Directory.EnumerateFiles(
                         directoryOrFilePath,
-                        "*.cs",
+                        "*.*",
                         SearchOption.AllDirectories
                     )
+                    .Where(o => o.EndsWithIgnoreCase(".cs") || o.EndsWithIgnoreCase(".csx"))
                     .Select(o =>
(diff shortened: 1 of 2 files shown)


belav:
Merging so I can get 0.27.1 out shortly. The searching by *.* may be slow, but I didn't see a way to search for globs. And I assume two searches for different searchPatterns would be slower than *.*.

There will be more extensions to search for if I ever find time to get back to the csproj branch.

Src/CSharpier.Cli/CommandLineFormatter.cs
```diff
@@ -240,9 +240,10 @@ async Task FormatFile(string actualFilePath, string originalFilePath)
                 var tasks = fileSystem
                     .Directory.EnumerateFiles(
                         directoryOrFilePath,
-                        "*.cs",
+                        "*.*",
```
shocklateboy92, line 243: you could make the pattern `*.cs*`, which will probably match a lot less files.

PR #1176, opened 2024-02-12
Sync exposed CodeFormatterOptions with currently supported cli options

Phault:
Tried to incorporate your feedback. Personally think it's just fine to have an enum with just two values, so I went along with it. Had to make EndOfLine public as well, hence I also moved it out of PrinterOptions.cs, but wasn't sure where to place it.

Not familiar with how that PublicAPI.Unshipped.txt stuff works, but I let my IDE update it. Let me know if it's all wrong :)

### DIFF
--- a/Src/CSharpier.Tests/CodeFormatterTests.cs
+++ b/Src/CSharpier.Tests/CodeFormatterTests.cs
@@ -61,30 +61,39 @@ internal class CodeFormatterTests
 
         result.Code.Should().Be("var someVariable =\n    someValue;\n");
     }
-    
+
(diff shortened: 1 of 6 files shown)


belav:
> Not familiar with how that PublicAPI.Unshipped.txt stuff works, but I let my IDE update it. Let me know if it's all wrong :)

I don't recall exactly how it works, but I believe that is the first step. Then when the next release goes out all of those changes need to move into the Shipped.txt version of the file.

belav (approved):
Looks good, thanks!

PR #1179, opened 2024-02-17
Adding formatting for spread element

### LINKED ISSUE #1167: Spread (in collection expression) are not formatted
**Input:**
```cs
List<KeyValuePair<string, string>> list =
[..    attribute.Targets.Select(target =>    KeyValuePair.Create(target, context.EntityDefinitions.TryGetValue(target, out var type) ? type.ClassName : null)).Where(kvp => kvp.Value != null)];
```

**Output:**
```cs
List<KeyValuePair<string, string>> list =
[
    ..    attribute.Targets.Select(target =>    KeyValuePair.Create(target, context.EntityDefinitions.TryGetValue(target, out var type) ? type.ClassName : null)).Where(kvp => kvp.Value != null)
];
```

**Expected behavior:**
The spread doesn't seem to be formatted at all. 
For example, extra spaces after the `..` or after `=>` are not condensed. 
The line is extremely long but it's not broken.

I would expect something like
```cs
List<KeyValuePair<string, string>> list =
[
    ..attribute
        .Targets.Select(target =>
            KeyValuePair.Create(
                target,
                context.EntityDefinitions.TryGetValue(target, out var type) ? type.ClassName : null
            )
        )
        .Where(kvp => kvp.Value != null)
];
```

belav:
closes #1167

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/SpreadElements.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/SpreadElements.test
@@ -1,0 +1,21 @@
+int[] someArray = [.. someOtherArray];
+int[] someOtherArray = [.. value1, .. value2, .. value3];
+
+int[] someOtherArray =
+[
+    .. value1________________________________,
+    .. value2________________________________,
+    .. value3________________________________
+];
+
+List<KeyValuePair<string, string>> list =
+[
+    .. attribute
+        .Targets.Select(target =>
+            KeyValuePair.Create(
+                target,
+                context.EntityDefinitions.TryGetValue(target, out var type) ? type.ClassName : null
+            )
+        )
+        .Where(kvp => kvp.Value != null)
+];
--- a/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/SpreadElement.cs
+++ b/Src/CSharpier/SyntaxPrinter/SyntaxNodePrinters/SpreadElement.cs
@@ -1,0 +1,13 @@
+namespace CSharpier.SyntaxPrinter.SyntaxNodePrinters;
+
+internal static class SpreadElement
+{
+    public static Doc Print(SpreadElementSyntax node, FormattingContext context)
+    {
+        return Doc.Group(
+            Token.Print(node.OperatorToken, context),
+            " ",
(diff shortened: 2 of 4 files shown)


belav:
@jods4 I went with a space after the operator because that is what is in all the official docs 
```c#
int[] someArray = [.. list];
```

There were only two instances of spread in here - https://github.com/belav/csharpier-repos/pull/104 - or at least only two that changed formatting. I'm going to update the code there and then do some proper querying to see how the spread operator has been used so far. I'm open to either space or no space. JS/TS is no space as far as I know.

jods4:
@belav I wrote my example without a space after `..` because I'm used to JS/TS, but I don't have a strong preference. Space is fine, too.

Most important is that you've fixed the blind spot in formatting the spreaded value! 👍

prettyv:
As already mentioned due to how the spread element/operator is used without an additional space in at least JS/TS the option without space would feel more "natural" to me, especially when switching between languages often. The official [documentation](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/collection-expressions#spread-element) is also a bit inconsistent here where in its explanatory text it doesn't use spaces either ("The spread element `..vowels`", "The spread element `..consonants`"), which makes more intuitive sense to me akin to prefix increment and decrement operators.

Would be happy if this could be reconsidered.

jerone:
Excuses to respond to an older PR.

I came to this repo to report (or ask to remove) the space between the spread operator and the operand.

This just looks weird:
```csharp
internal class Test
{
    private static readonly int[] Array = [1, 2, 3];

    private static void Main()
    {
        int[] _ =
        [
            .. Test.Array,
            .. Test.Array,
            .. Test.Array,
            .. Test.Array,
            .. Test.Array,
            .. Test.Array,
        ];
    }
}
```

Me too, am familiar with [JS syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax), but then again, the spread operator in JS uses three dots anyways.

After reading https://github.com/dotnet/roslyn/discussions/72863 & https://github.com/dotnet/roslyn/issues/71219 it appears that **the space is intentional**.

> Ranges and spreads are entirely different and intentionally have different spacing to reflect that. 

> it's not a unary expression, so it gets a space

TIL

PR #1188, opened 2024-02-24
Better support for dotnet commands.

### LINKED ISSUE #893: Plugins - better support for dotnet tool commands
still exists for csharpier v0.24.1 + rider plugin v1.3.7 + macOS + globally installed tool.

since it's hard to set `PATH` environment variable for macOS GUI apps, the `PATH` env is default and does not contain `dotnet tools install -g` dir. `~/.dotnet/tools/`

```sh
$ which dotnet-csharpier
/Users/magicdawn/.dotnet/tools/dotnet-csharpier
```

suggestions:

- replace `dotnet csharpier --version` with `sh -c "dotnet csharpier --version"` on posix systems (unix & macOS), let shell loads `PATH` from `.bash_profile` / `.zshenv` etc
- add search support for `~/.dotnet/tools`, since it's the default global install location. see https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-tool-install#description

https://github.com/belav/csharpier/blob/rider-1.3.7/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessProvider.java#L161

-----------

And a note for other macOS users want to set PATH for GUI apps

`launchctl config user path "xxxx1:xxxx2:xxxx3"`, requires reboot

see https://github.com/hschmidt/EnvPane#why-cant-i-set-path-with-envpane

_Originally posted by @magicdawn in https://github.com/belav/csharpier/issues/716#issuecomment-1532178103_

belav:
This makes use RiderDotNetActiveRuntimeHost to find the path to the dotnet exe, and sets DOTNET_ROOT where appropriate. This should remove any need for the PATH to be correct.

closes #893

### DIFF
--- a/Src/CSharpier.Rider/CHANGELOG.md
+++ b/Src/CSharpier.Rider/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 # csharpier-rider Changelog
 
+## [1.6.0]
+- Better support for dotnet commands.
+  - Uses the Rider setting for '.NET CLI executable path' for running dotnet commands
+  - If unable to run dotnet commands, show an error message
+
 ## [1.5.3]
 - Add experimental support for CSharpier server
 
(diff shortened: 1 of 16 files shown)


belav:
Tested this on ubuntu and I believe it is good.

PR #1200, opened 2024-03-08
Update Website to docusaurus v3

### LINKED ISSUE #1198: Docs use C# instead of CS for code-blocks
Hi! Minor nitpick!

I belive your should run a find/replace to replace ` ```c#` with ` ```cs` to make syntax highlighting work on the csharpier website ( https://csharpier.com).

The recognized codes for prism that you use for code higlighting in the docusaurus setup looks for one of these: `csharp, cs, dotnet`

Example: https://csharpier.com/docs/About the first code example here is not colored correctly and it would look better with a tiny change :)

Feel free to close this issue if you think its not worth fixing, or add a comment and I'll fix it and submit a PR

Strepto:
This fixes #1198 , but is also a normal maintenance task.

`npm audit` now reports 0 issues (the `main` branch has `26 vulnerabilities (15 moderate, 11 high)`) (I know this does not really matter for a static page, but its nice to keep up to date ayhow)

Let me know if this is a useful contribution, as this was not discussed beforehand. If not feel free to close it!

I have tested this locally only, but I cannot see any visual differences between this and csharpier.com (prod) 
Except for the code highlighting actually working 🌟.

Question: This code updates the docs in the website folder but they seem to be duplicated from the root Docs folder. Which is the "master" of these?

How to test:
 
```cmd
git pull branch
cd to the src/Website folder
npm install
npm run start
```


Src/Website/docusaurus.config.js
```diff
@@ -110,9 +109,9 @@ const config = {
                 ],
             },
             prism: {
-                additionalLanguages: ["csharp"],
-                theme: lightCodeTheme,
-                darkTheme: darkCodeTheme,
+                additionalLanguages: ["csharp", "bash", "powershell", "json", "editorconfig", "yaml", "ignore"],
+                theme: prismReactRenderer.themes.github,
```
Strepto, line 113: Consider if this is the correct theme to use? Its a bit bright and low contrast for my eyes, but this is at least the same as you had before.

belav: I'm not opposed to it being changed, I think I just based the current one off of how the prettier site looked

belav:
> Let me know if this is a useful contribution, as this was not discussed beforehand. If not feel free to close it!

It is appreciated! I tend to neglect things like this because they don't seem as urgent.  Getting the syntax highlighting working is great, I may have just not noticed it because I usually look at the docs in my ide/gh where they worked.

> Question: This code updates the docs in the website folder but they seem to be duplicated from the root Docs folder. Which is the "master" of these?

The root Docs folder is the next unreleased version of them, which may have changes for unreleased features. I copy them over the ones in the Website folder as part of the release. As of right now I believe the files should be the same, so if you copy the website ones into that folder and only see your changes we are all good.

belav (changes requested):
Only change requested is getting the root docs updated, otherwise this looks good!

Src/Website/docusaurus.config.js
Strepto: That reasoning works for me! I'd consider using the vsLight theme for increased contrast, but this is all preference so lets merge this and if you want to switch just do it separately 👍

belav (approved):
Thanks for the contribution!

Src/Website/docusaurus.config.js
belav: vsLight looked pretty similar to me. I apparently disabled the toggle that lets the user enable dark mode. So instead of changing the theme I brought that back and got it looking good. Going back to the light mode after seeing it with dark mode is tough on my eyes

Strepto: Agreed! I just assumed there was a reason for the no dark mode since it was disabled 🌞

The dark themes look better, so I support enabling it! 👍

PR #1208, opened 2024-03-13
Another fix

belav:


belav:
Submitted 1.6.3 to jetbrains, it should be available soon.

PR #1211, opened 2024-03-18
Update Configuration.md

belav:
Adding the editorconfig values csharpier uses for its using sorting logic.


belav (commented):
also update website doc

PR #1237, opened 2024-04-20
Consider wide unicode characters to be size of 2

### LINKED ISSUE #260: Fullwidth unicode characters should be accounted for in print width
Unicode characters can be fullwidth, which takes the same space as two normal characters. This isn't taken into account when printing right now. A method was abstracted away to /Src/CSharpier/Utilities/StringExtensions.cs that can be modified to implement this feature.

See https://github.com/prettier/prettier/blob/main/src/utils/get-string-width.js for how prettier does it.

There doesn't seem to be much about this that I can find in c#/.net
https://stackoverflow.com/questions/65410411/how-to-reliably-determine-the-width-of-a-character-in-c

belav:
closes #260

### DIFF
--- a/Src/CSharpier.Tests/CodeFormatterTests.cs
+++ b/Src/CSharpier.Tests/CodeFormatterTests.cs
@@ -4,6 +4,8 @@ using NUnit.Framework;
 
 namespace CSharpier.Tests;
 
+using CSharpier.Utilities;
+
 // TODO xml move these around
 [TestFixture]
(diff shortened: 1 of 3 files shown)


shocklateboy92 (approved):
Damn, that was not a problem I ever expected to encounter

belav:
> Damn, that was not a problem I ever expected to encounter

Didn't actually encounter it, but prettier had code for handling it when I ported things over and I finally got around to implementing it.

PR #1251, opened 2024-05-03
Basics of new override ability by extension, which will also allow other file extensions to be formatted.

### LINKED ISSUE #1220: Request for support of custom file extensions to scan/fix
**Request**

As it currently stands, it is not possible to scan C# files that have custom extensions (for build/deployment processes etc.)
The only work around for this, is to have a special script/process: that changes the extension, runs CSharpier, then puts the extension back.

**Concept**

Add a config section to the '.csharpierrc.*' that allows the user to pick the file extensions to target ie.
```yml
printWidth: 120
useTabs: false
tabWidth: 2
fileExtensions:
  - .cs
  - .cs.custom
```

belav:
closes #1220

I mostly went with how prettier works for overrides. Using a file glob, a user can specify options to be used for a file. One of those options is which formatter to use. Right now this means non-standard files can be formatted with the csharp formatter. Some day the xml formatting PR will finally get some attention.


Src/CSharpier.Cli/CommandLineFormatter.cs
```diff
@@ -206,25 +213,33 @@ async Task FormatFile(string actualFilePath, string originalFilePath)
...
-                    commandLineFormatterResult,
-                    writer,
-                    commandLineOptions,
-                    printerOptions,
-                    formattingCache,
-                    cancellationToken
-                );
+                if (printerOptions is not null)
+                {
+                    printerOptions.IncludeGenerated = commandLineOptions.IncludeGenerated;
```
shocklateboy92, line 219: On an unrelated note, isn't it weird that everything else in CommandLineOptions is about control flow, and just this one is about formatting?

Should we maybe consider removing this option from CLI and have it only in the config files?

Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
```diff
@@ -16,128 +12,70 @@ internal class ConfigurationFileOptions
...
+                EndOfLine = matchingOverride.EndOfLine,
+                Formatter = matchingOverride.Formatter
+            };
         }
 
-        // already found any in this directory above
-        directoryInfo = directoryInfo.Parent;
-
-        while (directoryInfo is not null)
+        if (filePath.EndsWith(".cs") || filePath.EndsWith(".csx"))
```
shocklateboy92, line 33: this check has come up in multiple places (including `EditorConfigSections.cs`). Do you want to extract the extensions to a constant?
Or even create a helper method?

Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
```diff
@@ -16,128 +12,70 @@ internal class ConfigurationFileOptions
...
             {
-                results.Add(
-                    new CSharpierConfigData(
-                        firstFile.DirectoryName!,
-                        Create(firstFile.FullName, fileSystem, logger)
-                    )
-                );
-            }
+                TabWidth = matchingOverride.TabWidth,
+                UseTabs = matchingOverride.UseTabs,
```
shocklateboy92, line 26: out of curiosity, if the matching override contains a `TabWidth` set, but not `UseTabs`, will it just go the default value of `UseTabs`? Or will it pick the value from the global printer options that have been set elsewhere?

I.e. is it okay that you're not doing
```csharp
TabWidth = matchingOverride.TabWidth ?? this.TabWith,
/* same for the rest of the properties */
```

You'll also have to make the relevant properties in `Override` nullable of course.

shocklateboy92 (approved):
Bunch of suggestions, including whether partial overrides are supported.

Src/CSharpier.Cli/CommandLineFormatter.cs
belav: I see it as similiar to directoryOrFile option `dotnet csharpier <directoryOrFile>` which is used to decide which files/folders to format. Currently all the options in in the config file are about what to do when a file is formatted, as opposed to what files to format. Although the new overrides option allows you to include extensions that aren't normally included.

`IncludeGenerated` does feel weird on `PrinterOptions`. I couldn't really find a better place to put it.

Long term I think csharpier should transition to what prettier does, where `<directoryOrFile>` uses globs. So the current command of `dotnet csharpier .` would be the equivalent of `dotnet csharpier **/*.{cs|cst}`

Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
belav: I need to do something with it in https://github.com/belav/csharpier/pull/858/files, because that is going to add csproj/xml formatting. Maybe that is the time to switch the CLI option to use globs. I'll leave it for now and deal with it there, I'm sure there are plenty of conflicts on those lines already.

Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
belav: It appears that it is poorly named. I ran across these TODOs when noodling on renaming it. Apparently I had the idea already. Gonna just sneak it in with this PR, it seems easy enough to do.

```
// TODO rename this to PrintingContext
// TODO and rename PrinterOptions.TabWidth to PrinterOptions.IndentSize
internal class FormattingContext
```

The configuration file does use the `tabWidth` name, which seems to have come from prettier. Changing that seems not worth the headache.

PR #1259, opened 2024-05-16
Allow to work with semantic versioning

huer12:
The VS extension is not able to work with preview versions. If you have a version like 0.27.3-beta, then the extension does not work. This code is fixing this issue. It allows to work with preview versions.

### DIFF
--- a/Src/CSharpier.VisualStudio/CSharpier.VisualStudioShared/CSharpierProcessProvider.cs
+++ b/Src/CSharpier.VisualStudio/CSharpier.VisualStudioShared/CSharpierProcessProvider.cs
@@ -1,17 +1,11 @@
+using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Net;
-using System.Security.Policy;
-using System.Windows.Forms;
+using System.Text.RegularExpressions;
 using System.Xml;
-using Microsoft.VisualStudio.Imaging;
-using Microsoft.VisualStudio.OLE.Interop;
-using Microsoft.VisualStudio.Shell;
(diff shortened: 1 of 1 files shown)


belav:
The extension installs the version of csharpier that it detects into a custom location. There have been no releases of csharpier on nuget that use preview number versioning. If someone were to reference a beta version like this wouldn't they also need a nuget source to tell VS where to find said version?

huer12:
We have released such a nuget version for ourself on our nuget server. So you are right VS needs to find it, but it is able to do so in our environment. We thought that may be others are doing this too and decided to bring a fix into the master.

shocklateboy92:
It looks like you still have formatting errors.

shocklateboy92 (approved):

Src/CSharpier.VisualStudio/CSharpier.VisualStudioShared/CSharpierProcessProvider.cs
```diff
@@ -278,6 +277,27 @@ private ICSharpierProcess SetupCSharpierProcess(string directory, string version
...
+        {
+            if (Version.TryParse(version, out var installedVersion) && installedVersion != null)
+            {
+                return installedVersion;
+            }
+
+            // handle semver versions
+            // regex from https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+            var regex = Regex.Match(
+                "0.27.3-dg.13",
```
parched, line 290: Shouldn't this be `version`?

martindisch: You're right, fixed it in #1264.

PR #1261, opened 2024-05-17
Pull trailing trivia out of forced flat portion of interpolated string

### LINKED ISSUE #1258: Incorect formatting of ternary expresion with a comment after an interpolated string
**Input:**

```csharp
var x = string.IsNullOrEmpty("")
    ? $"empty" // comment after interpolated string
    : "not empty";
```

**Output:**
The second branch of the ternary expression gets moved after the comment on one line.

```csharp
var x = string.IsNullOrEmpty("") ? $"empty" // comment after interpolated string : "not empty";
```

**Expected behavior:**
The expression should be kept separated into multiple lines.
The problem only occurs when the first branch is an interpolated string. For a string with no interpolation, it works as expected and the following is kept as given:

```csharp
var x = string.IsNullOrEmpty("")
    ? "empty" // without interpolation it's fine
    : "not empty";
```

belav:
closes #1258

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/InterpolatedStringExpressions.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/InterpolatedStringExpressions.test
@@ -8,6 +8,10 @@ public class ClassName
         // comment stays on this line
         $"two";
 
+    public string TrailingComment = someCondition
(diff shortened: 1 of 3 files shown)


shocklateboy92 (commented):
LGTM

shocklateboy92 (approved):

PR #1264, opened 2024-05-21
Fix semver version detection for VS plugin

martindisch:


shocklateboy92 (approved):
Lol. Thanks @parched and @martindisch

Issue #1249, opened 2024-04-30
How can I make CSharpier work with Visual Studio?


Dennis-N8:
**Environments**
- IDE Version: Microsoft Visual Studio Community 2022 (64-bit) - Current
Version 17.9.6
- Extension Version: 1.7.1
- CSharpier Version:  0.28.2
- Operating System: Windows 10 Pro
- .csharpierrc Settings: Not present
- .editorconfig Settings: Not present

**Log Output**
[Info - 4/30/2024 19:30:53] Formatting started for C:\...\CarController.cs using CSharpier 0.28.2
[Warn - 4/30/2024 19:30:53] Failed posting to the csharpier server. System.Net.WebException: Unable to connect to the remote server ---> System.Net.Sockets.SocketException: The requested address is not valid in its context 127.0.0.1:0
   at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
   at System.Net.ServicePoint.ConnectSocketInternal(Boolean connectFailure, Socket s4, Socket s6, Socket& socket, IPAddress& address, ConnectSocketState state, IAsyncResult asyncResult, Exception& exception)
   --- End of inner exception stack trace ---
   at System.Net.HttpWebRequest.GetRequestStream(TransportContext& context)
   at System.Net.HttpWebRequest.GetRequestStream()
   at CSharpierProcessServer.formatFile(FormatFileParameter parameter)
[Info - 4/30/2024 19:30:53] Formatted in 1ms

**Steps to reproduce**
Right click inside any .cs file in Visual Studio and select 'Format With CSharpier'

**Expected behavior**
It should work as expected
 
**Actual behavior**
It doesn't format the .cs file.

**The output of 'dotnet --list-sdks':**
7.0.401 [C:\Program Files\dotnet\sdk]
8.0.204 [C:\Program Files\dotnet\sdk]

**global.json content**
{ "sdk": { "version": "7.0.401" } }

With and without `global.json` the CSharpier extension doesn't work.

Also, how can I make the CSharpier work with latest dotnet sdk regardless which sdk version is in global.json?

belav:
Based on the fact it is trying to connect using port 0, I believe it is failing to start the csharpier server but not detecting that it failed to start. Can you enable debug logging? That will hopefully point me in the direction of the problem.

[...]

PR #1276, opened 2024-05-31
Let kestrel pick the open port

belav:
See [this](https://github.com/belav/csharpier/issues/1249#issuecomment-2129993017) for why there are problems with trying to find a free port on our own. Letting kestrel pick the port seems like a much better option.

references #1249


shocklateboy92 (approved):
Damn, I would not have guessed that at all without that guy's helpful comment.

Issue #668, opened 2022-05-17
Enforce trailing commas in object and collection initializer


NonSpicyBurrito:
Take the following code:
```cs
var foos = new[]
{
    new Foo(),
    new Bar()
};
```
If later a new item is added:
```cs
var foos = new[]
{
    new Foo(),
    new Bar(),
    new Baz()
};
```
This causes the following issues:
- A misleading commit diff of `+2/-1` even though there isn't really any code deletion.
- Polluting git blame.

These issues also occur when deleting/reordering the last item in initializers.

These issues can be fixed by enforcing trailing commas.
Relevant Prettier option: https://prettier.io/docs/en/options.html#trailing-commas

belav:
I had considered this early on - #23. I think it was closed because we decided to not have csharpier changing anything besides whitespace. But there are good arguments for transforming code in some cases, and there are good arguments for forcing tailing commas. The discussions I found around making trailing commas the default in prettier - https://github.com/prettier/prettier/issues/9369 and https://github.com/prettier/prettier/issues/68

I think I am in favor of adding this. If it were to follow the prettier way, it would only include the trailing comma when the initializer breaks.

jods4:
Conversely, it would also be nice if CSharpier removed the trailing comma when the object initializer fits on single line.

I think it's a common style guide to not have the final comma on a single line, and include it on multiple lines, i.e. this format:
```cs
Dictionary<string, string> headers;

// No trailing comma after If-Match
headers = new()  { { "OData-Version", "4.0" }, { "If-Match", "*" } };

// Trailing comma after If-Match
headers = new() 
{
  { "OData-Version", "4.0" },
  { "Accept", "application/json;odata.metadata=minimal" },
  { "If-Match", "*" },
};
```

I think this makes a strong argument for CSharpier adding/removing that comma automatically:
- Depending on the length of initializers, CSharpier will reflow on single or multiple lines.
- Currently if you want to adhere to the style above, you need to _manually_ add/remove that comma based on CSharpier reflow, so _after_ CSharpier runs.
- This is not doable if you run CSharpier automatically, e.g. as part of a git commit hook. Even if you Format on Save, it's easy to miss.

tkukushkin:
Also csharpier could act as [Black](https://github.com/psf/black) (popular python code formatter) about trailing commas. It forces multiline output if sees trailing comma.
So this code in python
```python
def foo(a: int, b: str,) -> None:
```
will be formatted to
```python
def foo(
    a: int,
[...]

PR #1284, opened 2024-06-08
Prettier style trailing commas

dawust:
This pull request addresses the issue described in [Issue #668](https://github.com/belav/csharpier/issues/668) and [Issue #1285 ](https://github.com/belav/csharpier/issues/1285).
A flag UsePrettierStyleTrailingCommas is introduced which ensures a trailing comma for anonymous object/collection/initializer/switch expressions, list patterns and enum declarations in the case a line break occurs.

I hope the implementation of the feature isn't too invasive. Obviously it would be easier to implement the feature non-configurable.

I'm not sure how to deal with the tests, for now I enabled the flag and adjusted the corresponding tests.

Thanks for any feedback!

### DIFF
--- a/Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
+++ b/Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
@@ -16,6 +16,8 @@ internal class ConfigurationFileOptions
     [JsonConverter(typeof(CaseInsensitiveEnumConverter<EndOfLine>))]
     public EndOfLine EndOfLine { get; init; }
 
+    public bool UsePrettierStyleTrailingCommas { get; init; }
+
     private static readonly string[] validExtensions = { ".csharpierrc", ".json", ".yml", ".yaml" };
 
     internal static PrinterOptions CreatePrinterOptionsFromPath(
@@ -38,7 +40,8 @@ internal class ConfigurationFileOptions
             TabWidth = configurationFileOptions.TabWidth,
             UseTabs = configurationFileOptions.UseTabs,
             Width = configurationFileOptions.PrintWidth,
-            EndOfLine = configurationFileOptions.EndOfLine
+            EndOfLine = configurationFileOptions.EndOfLine,
+            UsePrettierStyleTrailingCommas = configurationFileOptions.UsePrettierStyleTrailingCommas
         };
     }
 
--- a/Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
+++ b/Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
@@ -16,8 +16,6 @@ internal class ConfigurationFileOptions
     [JsonConverter(typeof(CaseInsensitiveEnumConverter<EndOfLine>))]
     public EndOfLine EndOfLine { get; init; }
 
-    public bool UsePrettierStyleTrailingCommas { get; init; }
-
     private static readonly string[] validExtensions = { ".csharpierrc", ".json", ".yml", ".yaml" };
 
     internal static PrinterOptions CreatePrinterOptionsFromPath(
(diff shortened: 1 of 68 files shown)


Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
```diff
@@ -16,6 +16,8 @@ internal class ConfigurationFileOptions
     [JsonConverter(typeof(CaseInsensitiveEnumConverter<EndOfLine>))]
     public EndOfLine EndOfLine { get; init; }
 
+    public bool UsePrettierStyleTrailingCommas { get; init; }
```
belav, line 19: I'd get rid of the option and just always format this way. Prettier needed it as an option because fo the different flavors of ES. If some day c# adds new trailing comma support that isn't in old versions then CSharpier may need an option for "CSharpVersion".

Src/CSharpier/SyntaxPrinter/TrailingComma.cs
```diff
@@ -0,0 +1,16 @@
+namespace CSharpier.SyntaxPrinter;
+
+internal static class TrailingComma
+{
+    public static Doc Print(SyntaxToken nextSyntaxToken, FormattingContext context)
+    {
+        return
+            context.UsePrettierStyleTrailingCommas
+            && !nextSyntaxToken.LeadingTrivia.Any(o => o.IsDirective)
```
belav, line 9: Is there a test for this? I don't quite follow why the trailing comma is only added when there is not a directive.

Src/CSharpier/SyntaxPrinter/TrailingComma.cs
```diff
@@ -0,0 +1,16 @@
...
+            ? Doc.IfBreak(
+                Token.Print(SyntaxFactory.Token(SyntaxKind.CommaToken), context),
+                Doc.Null
+            )
```
belav, line 13: It is worth adding a couple of tests that validate the trailing comma is added/removed. That would require a test file with the expressions before the comma is added/removed, along with an expected file.

After rescanning the comments in #668 I believe most people agreed this logic was appropriate. IE - do not support for a magic trailing comma that forces something to break.

belav (changes requested):
Looking good so far! Thanks for taking this on.

The one big piece missing, is making sure `SyntaxNodeComparer` doesn't report the added/removed comma as a warning along with adding some tests in `SyntaxNodeComparerTests`.

I'm hoping the logic can go into one of the methods in `SyntaxNodeComparer` that compares lists.

Src/CSharpier.Cli/Options/ConfigurationFileOptions.cs
dawust: I absolutely agree with you about the decision. Removed the option.

Src/CSharpier/SyntaxPrinter/TrailingComma.cs
dawust: Added some tests with an expected file for the expected behavior

Src/CSharpier/SyntaxPrinter/TrailingComma.cs
dawust: Added a test for this behavior.
And yeah, this one is kinda lazy, because I don't really know how to make sure we don't add a comma if there is a leading comma in the directive, e.g.:
`public enum EnumDeclarationWithLeadingCommaInDirectiveBeforeBracket { 
    Foo = 1
#if DEBUG
    ,
    Bar = 2
#endif
}`

dawust:
> I'm hoping the logic can go into one of the methods in `SyntaxNodeComparer` that compares lists.
I guess it's ok to just omit the last separator (expression count - 1) when comparing the separators of the original node and the formatted node!?

Src/CSharpier/SyntaxPrinter/TrailingComma.cs
belav: ah yeah, conditional compilation causes a few headaches like this

belav:
> I guess it's ok to just omit the last separator (expression count - 1) when comparing the separators of the original node and the formatted node!?

Yeah that seems good enough, I don't see the csharpier code somehow being so screwed up that it replaces the last comma with a different separator.... and if it did then I don't think we could actually generate a syntax tree from the result.

belav (approved):
Nice work on all the tests, thanks for the contribution!

PR #1286, opened 2024-06-08
Updating all the extensions to only use csharpier server for 0.29.0+,…

belav:
… adding option to bypass csharpier server.

### DIFF
--- a/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessProvider.java
+++ b/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessProvider.java
@@ -219,16 +219,26 @@ public class CSharpierProcessProvider implements DocumentListener, Disposable, I
             // but there are still 0.12 and 0.14 downloads happening
             var installedVersion = version.split("\\.");
             var versionWeCareAbout = Integer.parseInt(installedVersion[1]);
+            var serverVersion = 29;
 
             ICSharpierProcess csharpierProcess;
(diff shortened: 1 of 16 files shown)


Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessProvider.java
```diff
@@ -219,16 +219,26 @@ private ICSharpierProcess setupCSharpierProcess(String directory, String version
...
+            ) {
                 csharpierProcess = new CSharpierProcessServer(customPath, version, this.project);
             } else if (versionWeCareAbout >= 12) {
                 var useUtf8 = versionWeCareAbout >= 14;
 
+                if (versionWeCareAbout >= serverVersion
+                    && CSharpierSettings.getInstance(this.project).getDisableCSharpierServer()
+                ) {
+                    this.logger.debug(
+                    "CSharpier server is disabled, falling back to piping via stdin"
```
shocklateboy92, line 236: on an unrelated note, are you using an auto-formatter for Java? 
Otherwise, maybe this string should be indented?

shocklateboy92 (approved):

Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessProvider.java
belav: Every time I'm in there I wish that I was, I should finally figure out how to set one up

PR #1297, opened 2024-06-26
Fix CLI message regarding directoryOrFile

### LINKED ISSUE #1296: The CLI has contradictory message about directoryOrFile being required
The `directoryOrFile` parameter is required for the command-line interface to work, however, the message in the output says that it might be ommited.

**Environments:**

- CSharpier Version:
- Running CSharpier via: cli
- Operating System: Fedora 40
- .csharpierrc Settings: defaults
- .editorconfig Settings: default

**Steps to reproduce:**
Run `dotnet csharpier`

**Expected behavior:**

It should print uniform message about parameters being required or not.

**Actual behavior:**

It prints:

```
>: dotnet csharpier
directoryOrFile is required when not piping stdin to CSharpier

dotnet-csharpier

Usage:
  dotnet-csharpier [options] [<directoryOrFile>...]

Arguments:
  <directoryOrFile>  One or more paths to a directory containing c# files to format or a c# file to format. If a path is not specified the current directory is used

# the rest doesn't matter....
```

Specifically, look at:

>  If a path is not specified the current directory is used

It isn't true. When path is not specified, the message above gets shown.

marcinjahn:
Closes #1296

### DIFF
--- a/Src/CSharpier.Cli/CommandLineOptions.cs
+++ b/Src/CSharpier.Cli/CommandLineOptions.cs
@@ -43,7 +43,7 @@ internal class CommandLineOptions
             {
                 Arity = ArgumentArity.ZeroOrMore,
                 Description =
-                    "One or more paths to a directory containing c# files to format or a c# file to format. If a path is not specified the current directory is used"
+                    "One or more paths to a directory containing C# files to format or a C# file to format. It may be ommited when piping data via stdin. If a path is not specified the current directory is used"
             }.LegalFilePathsOnly(),
             new Option(
                 new[] { "--check" },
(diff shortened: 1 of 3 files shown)


shocklateboy92:
@belav is this accurate?
I thought a dummy path was required even when piping stdin.

belav:
> @belav is this accurate? I thought a dummy path was required even when piping stdin.

This PR is accurate, looks like the help text never got updated when piping stdin was added.
[Code that enforces directoryOrFile](https://github.com/belav/csharpier/blob/437d36b4dd63d64e675ca04c6d5a972feb9f9da4/Src/CSharpier.Cli/CommandLineOptions.cs#L105)

Thanks @marcinjahn !

belav (approved):

Issue #957, opened 2023-09-08
MSBuild - when a file fails to compile csharpier interferes with getting you clickable links to the compilation errors.


belav:
> So, it turns out doing this broke error output:
Since CSharpier now runs before compilation, when a unit fails to compile, the regular compilation target is not run.
This means the error messages that IDEs (including VSCode) can pick up are not emitted.

![image](https://github.com/belav/csharpier/assets/3606529/8eb73424-b052-4c61-9f97-2b8d4369f227)

https://github.com/belav/csharpier/commit/15061deaad33600550c5e92ddcd3c2435c9837ca#r126341275

We can probably just ignore the result of running csharpier, or output the result but still continue on.

PR #1311, opened 2024-07-27
Treat files that don't compile as errors.

### LINKED ISSUE #1131: How to abort bash script if warnings from dotnet CLI tool?
What would be the proper way to abort a bash script if the dotnet CLI tool issues any warnings? I'm going to be formatting files in a CI pipeline during PRs and committing the changes back to the branch, so I do not want the script to move forward if CSharpier returns any warnings during the formatting process.

This is what I came up with initially, but it feels hackish and only works in my regular terminal not my CI pipeline, which seems to still include everything from stdout into the $errors variable, not just warnings and errors.
```
errors=$(dotnet csharpier . --write-stdout --loglevel Warning 2>&1)

if [ $? -ne 0 ] || [ -n "$errors" ]; then
    echo "$errors"
    echo "CSharpier formatting failed; aborting."
    exit 1
fi
```

I also tried this to see if any files remained unformatted, but CSharpier still returns exit code 0 with the `--check` flag when it warns there are files it cannot format.
```
dotnet csharpier .
dotnet csharpier --check .
if [ $? -ne 0 ]; then
    echo "CSharpier formatting failed; aborting."
    exit 1
fi
```

belav:
closes #1131


shocklateboy92:
So, I read through the linked issue and have the same feeling you do: that CSharpier used to do this before and we changed it on purpose.

Could you confirm that the behavior of CSharpier.MsBuild doesn't change as a result of this? I think there used to be an issue where CSharpier would stop be build before MsBuild go to the actual compile stage and we would only see the CSharpier errors and not the actual compiler errors.

rodion-m:
Could you please merge it and update the tool? I need it to properly use it with aider.

belav:
> So, I read through the linked issue and have the same feeling you do: that CSharpier used to do this before and we changed it on purpose.
> 
> Could you confirm that the behavior of CSharpier.MsBuild doesn't change as a result of this? I think there used to be an issue where CSharpier would stop be build before MsBuild go to the actual compile stage and we would only see the CSharpier errors and not the actual compiler errors.

I did a bit of digging on my phone and there was this issue - https://github.com/belav/csharpier/issues/957 - and the fix appears to be ignoring the output of csharpier when run via msbuild. I'll confirm that to be sure. Good call out!

belav:
> Could you please merge it and update the tool? I need it to properly use it with aider.

I'm not sure what aider is but I'm hoping to have a new release out within a week or two.

belav:
@shocklateboy92 you were correct that msbuild would be unhappy with this. Based on everything I just did I think #957 was about suppressing the warning output. The compilation failures would show up as a warning followed by an error without those changes. Getting msbuild happy with this change required adding a flag. I couldn't find another way to do it.

shocklateboy92 (dismissed):

PR #1314, opened 2024-08-06
Do not overwrite `CSharpier_Check`, when already set.

PetSerAl:
At this point `CSharpier_Check` can be already set. For example: `Directory.Build.props` imported before `CSharpier.MsBuild.props`, and `CSharpier_Check` can be set here. Set default value only when current is empty.


shocklateboy92 (approved):
LGTM. Thanks for your contribution.

Issue #1267, opened 2024-05-23
Modify plugins to use dotnet tool list -g to find globally installed version


virgiliu-evidos:
**Environments**
- IDE Version: Rider 2024.1.2 Build #RD-241.15989.179
- Extension Version: 1.7.3
- CSharpier Version:  0.28.2
- Operating System: macOS Sonoma 14.5
- .csharpierrc Settings: none
- .editorconfig Settings: none

**Log Output**

```
2024-05-23 20:12:35,314 [ 157128]   FINE - #c.i.c.CSharpierLogger - Using dotnet found from RiderDotNetActiveRuntimeHost at /usr/local/share/dotnet/dotnet
2024-05-23 20:13:12,162 [ 193976]   FINE - #c.i.c.CSharpierLogger - Ensure there is a csharpier process for /Users/myUser/repoRoot/src/projectOne
2024-05-23 20:13:12,176 [ 193990]   FINE - #c.i.c.CSharpierLogger - Looking for /Users/myUser/repoRoot/src/projectOne/.config/dotnet-tools.json
2024-05-23 20:13:12,176 [ 193990]   FINE - #c.i.c.CSharpierLogger - Looking for /Users/myUser/repoRoot/src/.config/dotnet-tools.json
2024-05-23 20:13:12,176 [ 193990]   FINE - #c.i.c.CSharpierLogger - Looking for /Users/myUser/repoRoot/.config/dotnet-tools.json
2024-05-23 20:13:12,177 [ 193991]   FINE - #c.i.c.CSharpierLogger - Looking for /Users/myUser/.config/dotnet-tools.json
2024-05-23 20:13:12,177 [ 193991]   FINE - #c.i.c.CSharpierLogger - Looking for /Users/.config/dotnet-tools.json
2024-05-23 20:13:12,177 [ 193991]   FINE - #c.i.c.CSharpierLogger - Looking for /.config/dotnet-tools.json
2024-05-23 20:13:12,177 [ 193991]   FINE - #c.i.c.CSharpierLogger - Unable to find dotnet-tools.json, falling back to running dotnet csharpier --version
2024-05-23 20:13:12,178 [ 193992]   FINE - #c.i.c.CSharpierLogger - Running /usr/local/share/dotnet/dotnet csharpier --version in /Users/myUser/repoRoot/src/projectOne
2024-05-23 20:13:14,027 [ 195841]   FINE - #c.i.c.CSharpierLogger - Could not execute because the specified command or file was not found.
[...]

PR #1325, opened 2024-08-18
Using dotnet tool list for rider extension

belav:


belav:
Handles #1267 for rider

Issue #1321, opened 2024-08-17
Rework the CLI to properly use commands + arguments.


belav:
Currently all arguments are available to csharpier but not all of them apply depending on what you are trying to do with csharpier.

Reworking the cli to use the subcommands from `System.CommandLine` will help with this.

Some examples of commands reworked
```
dotnet csharpier format . --loglevel None

dotnet csharpier check . --loglevel None

dotnet csharpier server --port 30000

dotnet csharpier pipe-files
```

Consider also removing the need to prefix things with `dotnet`, it is an option when building/publishing a dotnet tool.

stefannikolei:
Should the log level be an Global option?

stefannikolei:
I would work on this. Could we define the available commands with corresponding options in here?

PR #1332, opened 2024-08-23
Add csharpier as ToolCommandName

stefannikolei:
Fixes part of #1321 

With this change it should be possible to just call csharpier instead of dotnet csharpier


belav:
I'm assuming this gets rid of the old command, if so we should hold off on merging this for a bit. There are a few things I want to get out in 0.29.x and after that is going to be 1.0.0

belav (approved):

PR #1333, opened 2024-08-23
Reimplement CLI Commands to SubCommands

stefannikolei:
I started working on creating subcommands. 

Could I get feedback for this? Is this the direction you want it to go?

It would be good for me when we could define all commands with the possible options


belav:
> Could I get feedback for this? Is this the direction you want it to go?

I only had time to scan it quick, but yeah I this looks like what I was thinking, breaking apart the commands so it is clear which options apply to which command.

stefannikolei:
Ok i am finished. You could go for a review. One thing I was not sure about was which option should be available for the check command.

shocklateboy92 (approved):
LGTM. I'm going to let @belav do the final merge though.

belav (approved):
I'm not totally sure what options are needed for check, I'm guessing there will be a few from format that need to be added to check. Probably these three

```
--no-msbuild-check
--include-generated
--compilation-errors-as-warnings
```
It is also possible that `--no-cache` would apply.

I can always merge for now and figure it out later unless you wanted to take it on in the same PR. Sorry for the delay in getting back to this, and thanks for the contribution!

stefannikolei:
As you want. You can also ping me when you figured it out.

belav:
Got the last few sub options cleaned up and settled on just making a new branch for 1.0.0 instead of waiting to merge this to main. Thanks again!

PR #1344, opened 2024-09-08
fixing issue when formatting with tabs that would convert tabs in com…

### LINKED ISSUE #1343: Comments don't follow tabs indent style
When formatting the document, any comments which have indenting, get updated to use space indenting, even when indent style is set to tabs. The indenting provided by the function is kept, but the extra indenting from the comment is spaces. This makes the indenting in this example `[tab][tab][space][space][space][space]` instead of `[tab][tab][tab]`. 

I found out this problem because VS Code highlights lines with inconsistent indenting, and I was writing pseudocode in a comment block. I apologise for the awful looking formatting, but it is the best way I could find to demonstrate the problem. 

**Input:**
```
	public void ExampleFunction()
	{
		/*
		The following line is an example with an indent:
			This line is indented by one tab.
		*/
	}
```
**Output:**
```
	public void ExampleFunction()
	{
		/*
		The following line is an example with an indent:
		    This line is indented by one tab.
		*/
	}
```
**Expected behavior:**
```
	public void ExampleFunction()
	{
		/*
		The following line is an example with an indent:
			This line is indented by one tab.
		*/
	}
```

belav:
…ments into spaces.

closes #1343

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/Comments_Tabs.expected.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/Comments_Tabs.expected.test
@@ -19,7 +19,7 @@ public class Foo
 		*/
 		/*
 		The following line is an example with an indent:
-		    This line is indented by 4 spaces but will be converted to 1 tab
+			This line is indented by 4 spaces but will be converted to 1 tab
 		*/
 		/*
 		The following line is an example with an indent:
--- a/Src/CSharpier/DocPrinter/DocPrinter.cs
+++ b/Src/CSharpier/DocPrinter/DocPrinter.cs
@@ -221,6 +221,12 @@ internal class DocPrinter
                         this.Output.Append(indent.Value);
                         spacesToAppend -= indentLength;
                     }
+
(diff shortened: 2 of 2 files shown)


belav:
This was a pretty quick fix, but leads to a question of what to do when the line in the comments is tabs followed by spaces. This is what I ended up doing. All the leading whitespace in here is tabs unless otherwise indicated by the comments. I think the altenative would be to turn anything less than a full tab into 1 tab, unless it matches the pattern of that first comment. 

Or another alternative would be to try to leave alone any whitespace that is "part of the block". Meaning the 3 comments below all start with 2 indents that are tabs, and any whitespace after that is left alone because it is in the block of the multiline comment.

```c#
public class Foo
{
	/**
	 * comment - existing behavior, keeps the single space to align the *
	 */
	public class Bar { }

	public void SomeFunction()
	{
		/*
		The following line is an example with an indent:
			This line is indented by one tab.
		*/
		/*
		The following line is an example with an indent:
			This line is indented by 4 spaces but will be converted to 1 tab
		*/
		/*
		The following line is an example with an indent:
		   This line is indented by 3 spaces but will be left as 3 spaces
		*/
	}
}

PR #1350, opened 2024-09-17
Update diffengine

stefannikolei:


Src/CSharpier.Tests/FormattingTests/SetUpFixture.cs
```diff
@@ -10,16 +10,26 @@ public class SetUpFixture
...
+        var targetTitle = Path.GetFileName(target);
+        return $"/u /wr /e \"{target}\" \"{temp}\" /dl \"{targetTitle}\" /dr \"{tempTitle}\" /cfg Backup/EnableFile=0";
+    }
+
+    static string RightArguments(string temp, string target)
+    {
+        var tempTitle = Path.GetFileName(temp);
+        var targetTitle = Path.GetFileName(target);
+        return $"/u /wl /e \"{temp}\" \"{target}\" /dl \"{tempTitle}\" /dr \"{targetTitle}\" /cfg Backup/EnableFile=0";
+    }
```
belav, line 34: The files show up in winmerge properly, but these changes result in the `TestCase.actual.test` file being editable, when it should be `TestCase.test` that is editable.

Or in other words "Copy from Left" is an option, but should not be, and "Copy to Left" is unavailable but should be.

image

belav (changes requested):

Src/CSharpier.Tests/FormattingTests/SetUpFixture.cs
stefannikolei: Ah... Probably temp and target need to change the position..

PR #1369, opened 2024-11-01
VS Extension now hot reloads its config

### LINKED ISSUE #1367: Suggestion - VS - Configuration hot reload
**Environments**
- IDE Version: vs2022 pro
- Extension Version: 1.8.0
- CSharpier Version: 0.28.2 
*will NOT be the same as the extension version*
- Operating System: win11
- .csharpierrc Settings: doesn't exist
- .editorconfig Settings: 
  ```
  [*.cs]
  max_line_length = 160
  ```

**Steps to reproduce**

1. Have the VS extension installed
2. Open VS + solution
3. outside of VS, create a `.vs/$(SolutionName)/v17/csharpier.json` containing `"RunOnSave": true`
4. `ctrl-s` on a c# file in the previously opened VS

**Expected behavior**

The c# file should be formatted.

**Actual behavior**

The file based configuration has not been loaded from disk when the file changed.

**Experiment**

Updated `CSharpierOptions.cs`'s [`LoadAsync`](https://github.com/belav/csharpier/blob/356394035b3fcccfaa272a399a9603ec3845f72a/Src/CSharpier.VisualStudio/CSharpier.VisualStudioShared/CSharpierOptions.cs#L95C3-L96C68) to contain for following code: 

```csharp
var filePath = await getFilePath();

static FileSystemWatcher Create(string filePath)
{
    FileSystemWatcher fsw = new FileSystemWatcher(Path.GetDirectoryName(filePath), Path.GetFileName(filePath));

    static void OnSomethingChanged(object sender, FileSystemEventArgs e)
    {
        ThreadHelper.Generic.BeginInvoke(() => Instance.Load());
    }

    fsw.Changed += OnSomethingChanged;
    fsw.Created += OnSomethingChanged;
    fsw.Renamed += OnSomethingChanged;
    fsw.EnableRaisingEvents = true;

    return fsw;
}

_fileSystemWatcher = _fileSystemWatcher ?? Create(filePath);

if (filePath == null || !File.Exists(filePath))
{
    return;
}
```

And now the config gets reloaded when the file changed. 

We are using some automation to enable CSharpier automatically when someone is using VS, the dev experience would be better if it didn't need a VS restart to be picked up.

I can open a PR with the updated code.

scabana:
This fixes #1367. 

The `.vs/$(SolutionName)/v17/csharpier.json` file is now watched for changes and will reload the configuration when the file is changed.

### DIFF
--- a/Src/CSharpier.VisualStudio/CSharpier.VisualStudio/source.extension.vsixmanifest
+++ b/Src/CSharpier.VisualStudio/CSharpier.VisualStudio/source.extension.vsixmanifest
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
-        <Identity Id="83d6b6a0-9e25-4034-80f3-38445d8a8837" Version="1.8.0" Language="en-US" Publisher="CSharpier" />
+        <Identity Id="83d6b6a0-9e25-4034-80f3-38445d8a8837" Version="1.9.0" Language="en-US" Publisher="CSharpier" />
         <DisplayName>CSharpier</DisplayName>
         <Description xml:space="preserve">CSharpier is an opinionated code formatter for c#. It uses Roslyn to parse your code and re-prints it using its own rules.</Description>
         <MoreInfo>https://github.com/belav/csharpier</MoreInfo>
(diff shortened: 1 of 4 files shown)


shocklateboy92 (approved):
LGTM. You've tested this thoroughly, right?

Neither @belav or I use Visual Studio, so we won't really notice regressions very quickly.

belav (approved):
I did find that the options page itself doesn't hot reload with changes, but that seems like an edge case unlikely to come up in the real world. I'll push out the update.

belav:
> LGTM. You've tested this thoroughly, right?
> 
> Neither @belav or I use Visual Studio, so we won't really notice regressions very quickly.

@shocklateboy92 I tested all the test cases I could think of, the code seems pretty straight forward.

scabana:
Hey, yeah, checked new code does as intended. Also validated updating the config from the UI worked and didn't trigger a chain reaction. 

Sorry for the format issue, I thought I had formatted it.

belav:
> Sorry for the format issue, I thought I had formatted it.

No worries! I feel like I somehow miss formatting a file in almost every PR. Thanks for your contribution!

PR #1380, opened 2024-11-17
Ensure we are ready for the next major version.

belav:


belav:
this won't necessarily allow csharpier to format new language features, but it will allow someone to format existing c# files with just the new sdk installed.

PR #1387, opened 2024-11-19
Use current `dotnet` host instead of global

PetSerAl:
Use current `dotnet` binary from `DOTNET_HOST_PATH` instead of just `dotnet`.
Reasons:
1. Global (in `PATH`) may not exist (when used _Binaries_, but not _Installers_).
2. Global can have different runtime version.
3. Consistent with outer tools (`csc` for example).
https://github.com/dotnet/roslyn/blob/324fd25331c969cd742ba68eee09ffd4b6fd29e3/src/Compilers/Shared/RuntimeHostInfo.cs#L61-L64
4. It is documented to be used for that purpose.
https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_host_path


shocklateboy92 (approved):
LGTM. Thanks for your contribution.

PR #1457, opened 2025-01-22
Support users who needs multiple dotnet architecturess

letientai299:
I'm using a Mac M2, and for some reasons, I need to use both arm64 and x64 dotnet, configured per project basis. Currently the Rider plugins will try to reinstall csharpier whenever I switch from a project that use x64 dotnet to another project that use arm64 dotnet. That slows down the initial format significantly.

I fix this by detecting and adding dotnet architecture to the `~/.cache/...` path.

Some implementation notes:

- First run of `dotnet --info` after machine startup is slow. Subsequent runs is faster. I don't think `dotnet --info` is run often. So, this might slow down the first installation a bit.

- Alternative to use `dotnet --info` is either to implement a class that support all common 3 executable formats (PE, ELF, and Mach-O), or use a 3rd lib (e.g., https://tika.apache.org/). I don't want to write and test the custom detector class. And, I don't think a new dependency is worth. So, I go with the simplest solution.

Let me know if there's any other idea to fix the root cause.


belav:
I like the simple solution instead of trying to support multiple executable formats. My guess is supporting multiple formats would be more work and more likely to cause problems. Also this simple fix would be easy to port into VSCode.

Just a few thoughts I had.

1. Would it make sense to consider x64 the default architecture and not include it in the file path? It would mean most users don't have a new path, but it also means the paths are less consistent.
2. Instead of detecting the architecture, what about having a project level option for selecting it? I think a downside is that if you don't notice csharpier uninstalling/reinstalling itself you'd never know to set the option. On the other hand it would get rid of the concern with `dotnet --info` being a bit slow.
3. I was worried about parsing the output of `--info` for other languages, but I figured out how to set the language using `DOTNET_CLI_UI_LANGUAGE` and `Architecture:` is not translated for other languages.

I'm leaning towards keeping things as is, but let me know what you think. And thanks for the contribution!

letientai299:
I prefer to keep it as is. My opinions on your thoughts:

1. I think it's better to keep the path more consistent. So, we don't have to document why x64 is different with other architecture. My implementation is also treating detection failure as no architecture path segment, so, making x64 special would nullify that fallback option.

2. If we add the configuration to select it, then how would we resolve if that option and the project-level dotnet SDK is incompatible? I also prefer to keep this simple. In case we want to port this feature to other IDE/plugins, we only deal with the installation path, no need to take care of the setting UI/UX.

belav (approved):

belav:
I agree with everything you brought up, went ahead and submitted it to JetBrains. They usually approve it within 24hours or so.

Thanks again!

PR #1458, opened 2025-01-23
Fix: `dotnet test` failure in `CliTests.Should_Return_Error_When_No_DirectoryOrFile_And_Not_Piping_StdIn`

ogaken-1:
# Problem

The test `CliTests.Should_Return_Error_When_No_DirectoryOrFile_And_Not_Piping_StdIn` was consistently failing when executed via the `dotnet test` command.

This issue appears to have been caused by CliWrap, which initializes `ProcessStartInfo` with `RedirectStandardInput = true` by default. This behavior interfered with the test, as it relies on `Console.IsInputRedirected` being `false`.

# Solution

This patch updates the test to use `ProcessStartInfo` directly instead of relying on CliWrap. This ensures that `RedirectStandardInput` is explicitly set to `false`.

### DIFF
--- a/Src/CSharpier.Cli.Tests/CliTests.cs
+++ b/Src/CSharpier.Cli.Tests/CliTests.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using CliWrap;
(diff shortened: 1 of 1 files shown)


belav (approved):
Not sure why this passes on the GH action, but it also did fail for me when I tried `dotnet test` locally. Maybe it is a windows vs linux thing. Either way this looks good, thanks for figuring out a fix!

PR #1465, opened 2025-02-01
Avoid excessive file system watches for --server

### LINKED ISSUE #1468: Excessive file system watches for --server
Created to track the resolution of the issue originally reported as part of PR #1465 . Text below is copied from that PR.

**`dotnet-csharpier --server` accidentally installs file system watches for the entire directory.**

Editor plugins, like the Rider extension, run the csharpier server in the root directory of a solution. Using the root directory as the content root for an ASP.NET application is a bad idea, because the default host setup installs a recursive file system watch on the entire directory. This file system watch does not honor any ignore files and will quickly exhaust OS resource with large solutions/multiple server processes. This is problematic, because IDEs heavily rely on file system watchers to ensure that their file explorers match the state of the file system. Once csharpier has gobbled up all of the available file system watches, IDEs either have to resort to expensive polling or start to miss updates that happen outside of the IDE.

# Observations
When using CSharpier in the rider extension with `--server` mode enabled, I regularly run out of inotify watches. My system (Fedora Linux 41) is configured to allow millions of watches, yet that still isn't sufficient. 

image
```
# sysctl -a | grep inotify
fs.inotify.max_user_instances = 1048576
fs.inotify.max_user_watches = 2400000
```

When I check which processes are using that many file system watches, I see that `csharpier` is **by far** the single biggest offender.

```
PID 9325 (/home/chris/.cache/csharpier/0.30.6/dotnet-csharpier --server ): 278088 watches
PID 9392 (/home/chris/.cache/csharpier/0.30.6/dotnet-csharpier --server ): 277027 watches
PID 9448 (/home/chris/.cache/csharpier/0.30.6/dotnet-csharpier --server ): 277115 watches
PID 9485 (/home/chris/.cache/csharpier/0.30.6/dotnet-csharpier --server ): 278089 watches
PID 9538 (/home/chris/.cache/csharpier/0.30.6/dotnet-csharpier --server ): 278088 watches
PID 9684 (/home/chris/.cache/csharpier/0.30.6/dotnet-csharpier --server ): 278049 watches
PID 9837 (/home/chris/.cache/csharpier/0.30.6/dotnet-csharpier --server ): 278088 watches
```

I investigated by running csharpier under `strace`:

```bash
sudo strace -e trace=inotify_add_watch -u chris /home/chris/.dotnet/tools/dotnet-csharpier --server
```

and it would immediately list hundreds of thousands of calls to `inotify_add_watch` (the system call that is used to implement `FileSystemWatcher` on Linux). In there you would find `inotify_add_watch` calls for absolutely _everything_. `.idea`, `bin`, `obj`, `.git` with all of their subdirectories.

That was strange, since csharpier itself doesn't use FileSystemWatcher (except for the vscode extension config file). Running csharpier under the debugger revealed that **it was the `WebApplication.CreateBuilder` call itself that set these file system watches up as part of the default configuration setup.**

# Solution Approaches
## Change process working directory
We could, for instance, change the extensions to configure the process working directory for the `--server` process to be the cache directory. The process working directory is used as the default directory for the content root. 

This approach has the big advantage that it would also fix the issue for older versions of csharpier. 

The csharpier cache directory is not entirely empty. There are hidden subdirectories for different versions of .NET and for different languages, but at least they won't change as the user is working (no big git checkouts).

## `CreateSlimBuilder` :x: 
Uses the same default configuration setup.

## `CreateEmptyBuilder` :grey_question: 
Would work in theory, but the csharpier team would have to maintain the custom setup of the host.

## `NullFileProvider` :x: 
The [NullFileProvider](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.fileproviders.nullfileprovider?view=net-9.0-pp&viewFallbackFrom=net-8.0) is a dummy implementation that would prevent the web app to look for its config files in the default directory.

Unfortunately, you can't set it up early enough. The file system watchers are installed as part of the `CreateBuilder` call already. The file provider configuration can only be added later.

## Only make this change for Linux :grey_question: 
We could easily scope this change to only apply to Linux, but I'm also concerned about the issue of accidentally picking up and honoring `appsettings.json` files from the current directory. While it's probably not _super_ common to have `appsettings.json` files at the root of a repository, it would nonetheless be very confusing (e.g., if the csharpier server would suddenly snatch away the same port as the real application).

And even if other OSs don't have similarly harsh consequences for running out of watches, csharpier still asks the OS to watch the entire repository, potentially slowing down unrelated operations like git fetch or npm install.

chklauser:
Fixes #1468 (`dotnet-csharpier --server` accidentally installs file system watches for the entire directory.)

We configure the server to run against a temporary, empty directory instead, so that the file system watcher that does get installed only has to watch a single, empty directory.

This also means that the server won't pick up the default appsettings.json files in the working directory. We probably don't want this anyway because these appsettings could be intended for the user's solution, not for csharpier.

Alternative approaches are discussed in #1468

### DIFF
--- a/Src/CSharpier.Cli/Server/ServerFormatter.cs
+++ b/Src/CSharpier.Cli/Server/ServerFormatter.cs
@@ -17,7 +17,21 @@ internal static class ServerFormatter
         string? actualConfigPath
     )
     {
-        var builder = WebApplication.CreateBuilder();
+        // Editor plugins, like the Rider extension, run the server in the root directory
+        // of a solution. Using the root directory as the content root for an ASP.NET application
+        // is a bad idea, because the default host setup installs a recursive file system watch on the entire
+        // directory. This file system watch does not honor any ignore files and will quickly exhaust OS resource
+        // with large solutions/multiple server processes.
+        // We thus configure the server to run against a temporary, empty directory instead. This also means that
+        // the server won't pick up the default appsettings.json files in the working directory.
+        // We probably don't want this anyway because these appsettings could be intended for the user's solution,
+        // not for csharpier.
+        var emptyContentRoot = Directory.CreateTempSubdirectory("csharpier-empty-content-root");
+        AppDomain.CurrentDomain.ProcessExit += (_, _) =>
+            Directory.Delete(emptyContentRoot.FullName, true);
+        var builder = WebApplication.CreateBuilder(
+            new WebApplicationOptions { ContentRootPath = emptyContentRoot.FullName }
+        );
         builder.WebHost.ConfigureKestrel(
             (_, serverOptions) =>
             {


chklauser:
I'll revert the unintentional formatting changes. Give me a sec.

belav:
Thanks for the thorough explanation of the issue!

I wasn't sure how the file watcher uses the project directory as the content root and had the thought that a possible fix would be to set the working directory of the process to the same directory that the csharpier process exists in.

[rider](https://github.com/belav/csharpier/blob/main/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessServer.java#L43) does not set a working directory at all. So if the default behavior of rider is to run the process from the project root that could explain how it ends up as the content root.

[vscode](https://github.com/belav/csharpier/blob/main/Src/CSharpier.VSCode/src/CSharpierProcessServer.ts#L38) does specifically set the project directory as the working directory when starting the process. That was changed in #543 but I'm not exactly sure why it was needed.

I think it's worth checking if setting the working directory to the same directory as the csharpier process solves this problem. It means the issue can be fixed by just updating the extensions. When a file is formatted the path of that file is passed to the server, so I am fairly certain we can get anything we need using that path.

If that doesn't work I'd be inclined have the extensions disable csharpier server on linux for versions without this fix.

chklauser:
**Yes, I can confirm that it is the working directory that gets used as the default content root** (checked with strace). So if the extensions would set the process working directory to the same directory as the `dotnet-csharpier` executable, the file system watches wouldn't be that much of a problem.

I still see a bunch watches for directories in the `~/.cache/csharpier/$version/` directory, `.store` with different subdirectories for languages and .net versions. Unnecessary work still happens. But at least those directories won't experience massive changes (such as git checkouts) while you are working.

But I ultimately don't have a strong opinion on how exactly the issue gets fixed. I have created an issue #1468 with the problem description above (in case there are going to be multiple PRs).

belav:
I think it makes sense to have the plugin changes + this change. This change is better long term and I'm assuming updating the plugins fixes it for users on older versions.

Thanks again for the contributions!

belav (approved):

PR #1469, opened 2025-02-02
Rider: Set server pwd to cache dir

### LINKED ISSUE #1468: Excessive file system watches for --server

chklauser:
Fixes #1468 by changing the process working directory of the `--server` process to the csharpier cache directory.

### DIFF
--- a/Src/CSharpier.Rider/CHANGELOG.md
+++ b/Src/CSharpier.Rider/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # csharpier-rider Changelog
 
+## [2.1.1]
+- Set working directory for CSharpier Server to the directory that contains CSharpier. Fixes an issue with a recursive file watch being run on the project root.
+
 ## [2.1.0]
 - Improved support for users that work with multiple dotnet architectures by splitting out installs of csharpier based on the architecture they were installed with. Thanks go to @letientai299 for the contribution
 
--- a/Src/CSharpier.Rider/gradle.properties
+++ b/Src/CSharpier.Rider/gradle.properties
@@ -1,4 +1,4 @@
-pluginVersion = 2.1.0
+pluginVersion = 2.1.1
 
 # See https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
 pluginSinceBuild = 222
--- a/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessServer.java
+++ b/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessServer.java
@@ -41,6 +41,11 @@ public class CSharpierProcessServer implements ICSharpierProcess2, Disposable {
         var argument = Semver.gte(this.version, newCommandsVersion) ? "server" : "--server";
         try {
             var processBuilder = new ProcessBuilder(this.csharpierPath, argument);
+
+            // Setting the working directory is important because older versions of csharpier will accidentally
+            // install a recursive file system watch on the entire working directory. If this is the repository root
+            // it will include extremely large and unnecessary directories like .git, .idea, bin, obj, node_modules.
+            processBuilder.directory(new File(this.csharpierPath).getParentFile());
             processBuilder.environment().put("DOTNET_NOLOGO", "1");
             processBuilder.environment().put("DOTNET_ROOT", this.dotNetProvider.getDotNetRoot());
             var csharpierProcess = processBuilder.start();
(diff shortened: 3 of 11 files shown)


belav:
I went ahead and made the same changes for VS and VSCode. VS seemed unnecessary if this is only an issue on linux but it was a simple change. The rider plugin requires a manual review every time which can take a day or two.

Thanks for the contribution!

belav (approved):

PR #1488, opened 2025-02-18
feat: check for " " values and return a static instance

TimothyMakkison:
Add a check to `Doc`'s implicit operator for " " returning a static instance if found. Saves 0.5 MB in tests aka around 0.5%

#### Before
| Method                | Mean     | Error   | StdDev   | Gen0       | Gen1      | Gen2      | Allocated |
|---------------------- |---------:|--------:|---------:|-----------:|----------:|----------:|----------:|
| Default_CodeFormatter | 238.8 ms | 4.75 ms | 11.92 ms | 11000.0000 | 4000.0000 | 1000.0000 |  95.82 MB |

#### After
| Method                | Mean     | Error   | StdDev   | Gen0       | Gen1      | Gen2      | Allocated |
|---------------------- |---------:|--------:|---------:|-----------:|----------:|----------:|----------:|
| Default_CodeFormatter | 236.6 ms | 4.72 ms | 11.32 ms | 11000.0000 | 4000.0000 | 1000.0000 |  95.27 MB |

### DIFF
--- a/Src/CSharpier/DocTypes/Doc.cs
+++ b/Src/CSharpier/DocTypes/Doc.cs
@@ -14,8 +14,10 @@ internal abstract class Doc
 
     public static implicit operator Doc(string value)
     {
-        return new StringDoc(value);
+        return value is " " ? SpaceString : new StringDoc(value);
     }
+
+    private static readonly StringDoc SpaceString = new(" ");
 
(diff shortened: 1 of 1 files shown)


TimothyMakkison:
See #1494

PR #1491, opened 2025-02-19
feat: reuse collection types for `DocFitter`

TimothyMakkison:
Instead of creating a new instance of `StringBuilder` and `Stack<PrintCommand>` every time `DocFitter.Fit` is called we instead create a single instance of each in `DocPrinter` and pass it into `DocFitter.Fit`.

This could be replaced by using `ValueStringBuilder` and `ValueListBuilder`, the latter would be beneficial as unlike `Stack<T>` we can access items by index instead of the `O(n)` lookup time with `ElementAt`.

### Before
| Method                | Mean     | Error   | StdDev  | Median   | Gen0       | Gen1      | Gen2      | Allocated |
|---------------------- |---------:|--------:|--------:|---------:|-----------:|----------:|----------:|----------:|
| Default_CodeFormatter | 231.5 ms | 4.62 ms | 8.78 ms | 228.0 ms | 11000.0000 | 4000.0000 | 1000.0000 |  95.82 MB |

### After
| Method                | Mean     | Error   | StdDev  | Gen0       | Gen1      | Gen2      | Allocated |
|---------------------- |---------:|--------:|--------:|-----------:|----------:|----------:|----------:|
| Default_CodeFormatter | 227.0 ms | 4.51 ms | 6.33 ms | 10000.0000 | 4000.0000 | 1000.0000 |  90.03 MB |

### DIFF
--- a/Src/CSharpier/DocPrinter/DocFitter.cs
+++ b/Src/CSharpier/DocPrinter/DocFitter.cs
@@ -9,19 +9,22 @@ internal static class DocFitter
         Stack<PrintCommand> remainingCommands,
         int remainingWidth,
         Dictionary<string, PrintMode> groupModeMap,
-        Indenter indenter
+        Indenter indenter,
+        Stack<PrintCommand> newCommands,
+        StringBuilder output
     )
     {
+        // Reset reusable collections before usage
+        newCommands.Clear();
+        output.Clear();
+
         var returnFalseIfMoreStringsFound = false;
-        var newCommands = new Stack<PrintCommand>();
         newCommands.Push(nextCommand);
 
         void Push(Doc doc, PrintMode printMode, Indent indent)
         {
             newCommands.Push(new PrintCommand(indent, printMode, doc));
         }
-
-        var output = new StringBuilder();
 
         for (var x = 0; x < remainingCommands.Count || newCommands.Count > 0; )
         {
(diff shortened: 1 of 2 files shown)


belav (approved):
I think it feels a bit weird to pass these to DocFitter.

Other options I can think of

1. Make DocFitter non-static and move the fields into it.
2. Move the single method from DocFitter into DocPrinter. I think it used to be there and I broke it out because DocPrinter was getting a bit large.

Not sure if it is really worth worrying about. I feel like CSharpier is a mix of static/functional style code along with OOP. There are definitely much weirder things in the codebase. All of the SyntaxNodePrinters/[NodeType].cs files used to be partial classes of one giant class. When I converted it from js to c# that was the easiest way to get things working and it stuck around for a while before I refactored it.

I'll hopefully get caught up today, thanks for digging in and finding all of these improvements!

TimothyMakkison:
> I think it feels a bit weird to pass these to DocFitter

> Move the single method from DocFitter into DocPrinter. I think it used to be there and I broke it out because DocPrinter was getting a bit large.

It does feel a little redundant 😅 I'd consider doing the latter, it combines readability and performance


> When I converted it from js to c# 

I thought it felt familiar


> I'll hopefully get caught up today, thanks for digging in and finding all of these improvements!

I have a couple more improvements;
- `Indenter.IncresseIndent` could probably reuse indents saving 4MB, but I'm a little scare of breaking things rn.
- `PrintCommand` could be a struct, this has massive savings of 16MB, but I want to hold off until I finish more changes as converting reference types to struct sometimes cause performance regressions from copying.
- If you are okay with using `ValueListBuilder` we can make some massive improvemetns to `SyntaxPrinter.Token`, this allocates a massive amount of empty lists. 
- I did look at the logic in a couple of places, I can't make any promises 😄 All the trivia logic is giving me flashbacks to my failed attempt at fixing code maid

PR #1503, opened 2025-02-22
feat: cache `Indent` and use `PadRight`

TimothyMakkison:
Add cache to `IncreaseIndent`and use `PadRight`. Saves 4MB 6% of memory.

- Could this be a `List<Indent>`? If I understand correctly csharpier won't change between tabs and space mode, so this should be okay right?
- Assumes that `TypesForTabs` is only used for tabs mode.
- `AddAlign` and `MakeIndentWithTypesForTabs` could be improved, I should run the benchmarks with tabs mode enabled to see if its worth it.

### Before
| Default_CodeFormatter | 234.9 ms | 11.21 ms | 33.05 ms | 7000.0000 | 3000.0000 | 1000.0000 |  66.17 MB |

### After
| Default_CodeFormatter | 223.6 ms | 9.96 ms | 29.36 ms | 212.7 ms | 7000.0000 | 3000.0000 | 1000.0000 |  62.18 MB |

### DIFF
--- a/Src/CSharpier/DocPrinter/Indent.cs
+++ b/Src/CSharpier/DocPrinter/Indent.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics;
 using System.Text;
 
 namespace CSharpier.DocPrinter;
@@ -26,6 +27,7 @@ internal class AlignType : IIndentType
 internal class Indenter(PrinterOptions printerOptions)
 {
     protected readonly PrinterOptions PrinterOptions = printerOptions;
+    protected readonly Dictionary<string, Indent> IncreaseIndentCache = new();
 
     public static Indent GenerateRoot()
     {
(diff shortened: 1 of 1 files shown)


belav (approved):

belav:
> Could this be a List<Indent>? If I understand correctly csharpier won't change between tabs and space mode, so this should be okay right?

Are you referring to the dictionary? I'm guessing the idea is using an index into the list where the index is the `indent.value`.

CSharpier doesn't switch between tabs and spaces in a file, but the `Align` doc would add some spaces after tabs in some cases. But I don't actually see any usages of it. 

I'm guessing it was used when there was some formatting that would mix a 2 space indent into the usual 4 space indents. In the case of tabs it would add 2 spaces after the tab indent level. I'm in favor of ripping out `Align`. There was pushback to the mixing of indent level and I don't forsee any new formatting that would use it.

I approved but let me know if you want to make more changes here. I can also take care of ripping out `Align` unless you'd want to deal with it at the same time. It may result in some conflicts.

belav:
I removed align in this PR https://github.com/belav/csharpier/pull/1516

PR #1504, opened 2025-02-22
feat: use interpolated string in `RightHandSide`

TimothyMakkison:
Prevent unnecessary `Guid.ToString()` allocation using interpolated string. Saves 0.3MB

### Before
| Default_CodeFormatter | 234.9 ms | 11.21 ms | 33.05 ms | 7000.0000 | 3000.0000 | 1000.0000 |  66.17 MB |

### After
| Default_CodeFormatter | 230.9 ms | 9.77 ms | 28.66 ms | 220.6 ms | 7000.0000 | 3000.0000 | 1000.0000 |  65.86 MB |

### DIFF
--- a/Src/CSharpier/SyntaxPrinter/RightHandSide.cs
+++ b/Src/CSharpier/SyntaxPrinter/RightHandSide.cs
@@ -12,7 +12,7 @@ internal static class RightHandSide
     {
         var layout = DetermineLayout(leftNode, rightNode);
 
-        var groupId = layout == Layout.Fluid ? layout.ToString() + Guid.NewGuid() : String.Empty;
+        var groupId = layout == Layout.Fluid ? $"{Layout.Fluid}{Guid.NewGuid()}" : string.Empty;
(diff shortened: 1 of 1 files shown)


belav (approved):
Related - in the 1.0.0 branch I have two new methods for generating group ids, because I was sick of reading guids.

One was based on an incrementing number but doesn't seem to be used. https://github.com/belav/csharpier/blob/1.0.0/Src/CSharpier/DocTypes/Doc.cs#L104

I assume I replaced it with this more useful one that increments an int for a given group name. Pretty sure this idea came from the prettier code base.
https://github.com/belav/csharpier/blob/1.0.0/Src/CSharpier/SyntaxPrinter/PrintingContext.cs#L10

The group names only really need to be descriptive when troubleshooting. Maybe it is worthwhile to change `GroupId` to an `int` and add a `GroupName` that is only used when in debug.

belav (approved):

PR #1509, opened 2025-02-24
feat: convert `Node.Print` to use `switch (node.Kind())`

TimothyMakkison:
- Use `switch (syntaxNode.Kind())` instead of `is`. Should be slightly faster 🤞 
- Add `Debug.Assert` in the default branch to ensure that future changes to roslyn / mistakes with `SyntaxKind` are caught

### Benchmarks
Ran `FormatAsync` in a for loop 10 times to try and average performance. Seems to show a small 1.5% increase in performance at the cost of more complexity and the potential for errors. Do you think its worth merging this pr @belav ? TBH I'm a little dissapointed 😅 

I did wonder if using chained `or` might affect the generated code 🤷 the massive `SyntaxKind` switch statements in roslyn use several `case` tags instead. Perhaps this is lowered into better code? Looking at the IL I think it creates a jump table, although I'm not sure - I'm not fluent in IL 😄 .

#### Before
| Method                | Mean    | Error    | StdDev   | Gen0       | Gen1       | Gen2      | Allocated |
|---------------------- |--------:|---------:|---------:|-----------:|-----------:|----------:|----------:|
| Default_CodeFormatter | 2.115 s | 0.0132 s | 0.0103 s | 77000.0000 | 35000.0000 | 9000.0000 | 661.69 MB |

#### After
| Method                | Mean    | Error    | StdDev   | Gen0       | Gen1       | Gen2      | Allocated |
|---------------------- |--------:|---------:|---------:|-----------:|-----------:|----------:|----------:|
| Default_CodeFormatter | 2.085 s | 0.0168 s | 0.0149 s | 77000.0000 | 35000.0000 | 9000.0000 |  661.7 MB |

### DIFF
--- a/Src/CSharpier.Generators/NodePrinterGenerator.cs
+++ b/Src/CSharpier.Generators/NodePrinterGenerator.cs
@@ -1,7 +1,7 @@
-﻿using System.IO;
-using System.Linq;
-using Generators;
+﻿using Generators;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace CSharpier.Generators;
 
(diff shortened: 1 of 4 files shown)


TimothyMakkison:
> The special cases seem to be a bit painful to manage but I don't really see special cases being changed though, possibly just new ones added as the language evolves. The last few language versions have been really straightforward for CSharpier to support though and a few of them didn't even require any changes beyond updating the `CodeAnalysis` version. With the extra complexity hidden away I'm not super concerned with it.

Fair enough, for whats its worth I have manually checked each printer type case I missed any `SyntaxNode` types that don't share their name with a `SyntaxKind`. It's how I discovered that `unchecked` didn't have unit tests.

> I thought CSharpier's code coverage was solid until you uncovered the missing tests for `unchecked`. I do have a process for formatting everything in `csharpier-repos` which has caught quite a few bugs. I'm confident that would catch any errors this may introduce.

Did you change anything with CLI tests? 4 days ago they started failing. It keeps happening after I updated rider to 2024
image

> There may be a tiny benefit to using [RawSyntaxKind](https://github.com/belav/csharpier/blob/main/Src/CSharpier/SyntaxTriviaExtensions.cs#L19) which I added a while back.

Good shout, it can't hurt to add it.

> I just heard of [FrozenDictionary](https://learn.microsoft.com/en-us/dotnet/api/system.collections.frozen.frozendictionary-2?view=net-9.0) today although with the dictionary you added only being in the generator I don't know how beneficial it would be.

Yeah, I've been looking for an excuse to use it for ages. Unfortunately it would be a bad fit for the generator 😞 . Briefly considered it for `Node.Print` but I'm pretty sure the compiler is faster.

belav:
> Did you change anything with CLI tests? 4 days ago they started failing.

Nothing I can think of on `main`, most of my work lately has been going into `1.0.0`. I fixed a bug on `main` just now and that PR ran the tests and after updating this PR it is also running the tests.

Something I ran into on `1.0.0` is the CLI tests run inside of the bin of the project, which was pulling in a config file the tests weren't expecting. Maybe you have a file in your directory tree causing the same problem. I'm going to pull my `1.0.0` changes into `main` that move where the cli tests are run.

> Yeah, I've been looking for an excuse to use it for ages. Unfortunately it would be a bad fit for the generator 😞 . Briefly considered it for Node.Print but the compiler is faster.

Ah yeah good point. We do make use of it in a few spots at work, and maybe those would be better off as switch statements.

belav:
It looks like there is something missing in this PR, all of these test files have a blank line before `goto` removed - https://github.com/belav/csharpier-repos/pull/127/files

I validated that the original of the first file before any formatting did have those blank lines - https://github.com/belav/csharpier-repos/blob/main/AspNetWebStack/src/System.Net.Http.Formatting/Formatting/Parsers/FormUrlEncodedParser.cs#L174

TimothyMakkison:
mb I see the problem, `GotoStatementSyntax` has multiple kinds associated with it. I'll fix it and double check everything in case there are other issues.

belav (approved):

belav (approved):

PR #1510, opened 2025-02-24
test: add test for unchecked statements/expressions

TimothyMakkison:
Noticed that tests weren't failing for #1509 when I forgot to account for `SyntaxKind.UncheckedStatement`. This will hopefully catch it


belav (approved):
🚀

PR #1513, opened 2025-02-26
perf: add `ValueStringBuilder`

TimothyMakkison:
Add `ValueStringBuilder` to replace the `StringBuilder` allocation in `SeparatedSyntaxList.Print` saves 0.5MB
- Uses `MemoryExtensions.Trim` to avoid allocating an extra string
  - .NET Standard 2.0 doesn't have this method. I opted to add a simple allocating version as I assume .NET Standard performance doesn't matter as much.
  - I can copy over the relevant performance code for .NET Standard if you'd prefer
- Uses `SkipLocalsInit` to get around the minor zeroing cost during prolog, this isn't essential

Do you think this PRs complexity is worth adding it's not a massive saving.


belav:
If the only benefit is 0.5MB I don't think it would be worth the added complexity.

PR #1525, opened 2025-03-06
perf: added `ConditionalWeakTable` for existing `StringDoc` pairs

TimothyMakkison:
Uses `ConditionalWeakTable` to cache `StringDoc` types for a given `string`. My understanding of `ConditionalWeakTable` is that it won't prevent the garabage collector from removing the `string` type when it isn't referenced by anything else. This ensures that csharpier doesn't accumulate memory from unneeded strings.

This saves another 1 MB for a total of 3 MB (and appears to have some decent 15% performance improvement, I think this is an error but it seems pretty consistent 🤷 ). However I've never used `ConditionalWeakTable` before so I don't know if I'm misusing it here. WDYT?

### DIFF
--- a/Src/CSharpier/DocTypes/Doc.cs
+++ b/Src/CSharpier/DocTypes/Doc.cs
@@ -12,7 +12,7 @@ internal abstract class Doc
         return DocSerializer.Serialize(this);
     }
 
-    public static implicit operator Doc(string value) => StringDoc.Create(value);
+    public static implicit operator Doc(string value) => StringDoc.ToStringDoc(value);
 
     public static NullDoc Null => NullDoc.Instance;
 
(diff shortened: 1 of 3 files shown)


belav (commented):
Reading up on ConditionalWeakTable I... think it would be okay?

When there are no more references to the key outside of the table, then the entry can be removed from the table. The StringDoc that is the value in the table is returned, so it will be referenced in other parts of the code. The key isn't really referenced in the code beyond the call to ToStringDoc.

So assuming that removing an entry from the table doesn't GC the values in the table that may still be referenced then I think this won't cause any issues.

I can run the formatting on csharpier-repos. I think that formats around 50k files and if it doesn't cause problems on that, I think it will be good.

Assuming the 15% is even close to accurate that is a pretty significant boost and I think worth it.

`StringDoc.ToStringDoc` sounds a bit weird - maybe `StringDoc.From`, `Parse`, `Create`?

PR #1527, opened 2025-03-07
Add DocKind and remove IsInstanceOf in DocFitter

belav:
This modified all `Doc` classes to define a `DocKind` which can then be used in place of `IsInstanceOf`.

Before
image

After changing `DocFitter`
image

### DIFF
--- a/Src/CSharpier/DocPrinter/DocFitter.cs
+++ b/Src/CSharpier/DocPrinter/DocFitter.cs
@@ -39,11 +39,12 @@ internal static class DocFitter
                 _ => remainingCommands.ElementAt(x++),
             };
 
-            switch (currentDoc)
+            switch (currentDoc.Kind)
             {
-                case NullDoc:
+                case DocKind.Null:
                     break;
-                case StringDoc stringDoc:
+                case DocKind.String:
+                    var stringDoc = (StringDoc)currentDoc;
                     // directives should not be considered when calculating if something fits
                     if (stringDoc.Value == null || stringDoc.IsDirective)
                         continue;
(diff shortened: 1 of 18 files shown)


belav:
After changing `DocPrinter` things are slower.... I'm thinking the change from `if/else` to `switch` may have caused it.

image

@TimothyMakkison this was inspired by your PR I reviewed earlier that changed some code to operate on `Kind()` instead of using `IsInstanceOf`

belav:
I confirmed that my initial test for the after seemed to be a fluke. Maybe this isn't worth exploring. I assume the overhead of adding `DocKind` doesn't provide enough of a benefit, especially when a number of the cases have to cast `Doc` to the specific type. One of my iterations on `DocPrinter` did mix conditions for `if (doc.DocKind is X)` and `else if (doc is Y y)` which didn't seem to have much of an impact.

before
```
| Method                | Mean     | Error    | StdDev   | Gen0      | Gen1      | Gen2     | Allocated |
|---------------------- |---------:|---------:|---------:|----------:|----------:|---------:|----------:|
| Default_CodeFormatter | 91.43 ms | 1.823 ms | 3.045 ms | 2333.3333 | 1333.3333 | 333.3333 |  41.45 MB |
```

after
```
| Method                | Mean     | Error   | StdDev  | Gen0      | Gen1      | Gen2     | Allocated |
|---------------------- |---------:|--------:|--------:|----------:|----------:|---------:|----------:|
| Default_CodeFormatter | 107.4 ms | 2.03 ms | 2.34 ms | 3000.0000 | 2333.3333 | 666.6667 |  42.53 MB |
```

TimothyMakkison:
> @TimothyMakkison this was inspired by your PR I reviewed earlier that changed some code to operate on Kind() instead of using IsInstanceOf

Thanks for checking this out!

> I confirmed that my initial test for the after seemed to be a fluke. 

Yeah, this is one of the more annoying things when optimising code. For timing stuff I recommend
- Looping the code under test, this is why some of my times are in seconds because I ran `FormatAsync` in a for loop 10 times.
- Running the benchmarks 3 times each for the changes and the baseline, back to back.
- Looking at the median values from the 3 runs
- Closing all unnecessary programs in the background.

Needless to say, this is a PITA 😅 I usually watch netflix or cook while doing this.

> Maybe this isn't worth exploring. I assume the overhead of adding DocKind doesn't provide enough of a benefit

Hmm, it's that we can't improve on this.
 I wonder if instead of having a `DocKind` field, we implement this as an abstract method.
- This does have the cost of a virtual call
- Removes the overhead of the constructor and the additional 8 bytes for each `Doc` type
- Does let `RunOn` use an optimised `switch` statement

PR #1529, opened 2025-03-07
perf: use `ValueListBuilder<T>` for constant buffers

TimothyMakkison:
Follow on from #1502, using `ValueListBuilder` to replace known `List<Doc>` sizes.

- Most of these changes aren't measurable in the benchmarks, many are insignificant, although I suspect this is partly due to the limited code found in the benchmark file.
- This PR will increase the performance benefits of #1524

@belav what do you think the best way of replacing a list that uses a collection initializer is?

 I can either add the items directly into the scratch buffer and manually update `Length`
```C#
var docs = new ValueListBuilder<Doc>
([
    Modifiers.Print(node.Modifiers, context),
    Token.Print(node.DelegateKeyword, context),
    null,null
]);
docs.Length = 2;
```

Or create the buffer and then append the items via params.
```C#
var docs = new ValueListBuilder<Doc>([null, null, null, null]);
docs.Append(
    Modifiers.Print(node.Modifiers, context),
    Token.Print(node.DelegateKeyword, context)
);
```
Or Append each item
```C#
var docs = new ValueListBuilder<Doc>([null, null, null, null]);
docs.Append(Modifiers.Print(node.Modifiers, context));
docs.Append(Token.Print(node.DelegateKeyword, context));
```

### DIFF
--- a/Src/CSharpier/DocTypes/Doc.cs
+++ b/Src/CSharpier/DocTypes/Doc.cs
@@ -56,6 +56,16 @@ internal abstract class Doc
 
     public static Doc Concat(params Doc[] contents) => new Concat(contents);
 
+    public static Doc Concat(ref ValueListBuilder<Doc> contents)
+    {
+        return contents.Length switch
+        {
+            0 => Null,
+            1 => contents[0],
+            _ => new Concat(contents.AsSpan().ToArray()),
+        };
+    }
+
     public static Doc Join(Doc separator, IEnumerable<Doc> enumerable)
     {
         var docs = new List<Doc>();
--- a/Src/CSharpier/SyntaxPrinter/AttributeLists.cs
+++ b/Src/CSharpier/SyntaxPrinter/AttributeLists.cs
@@ -13,7 +13,7 @@ internal static class AttributeLists
             return Doc.Null;
         }
 
-        var docs = new List<Doc>();
+        var docs = new ValueListBuilder<Doc>([null, null]);
         Doc separator = node
             is TypeParameterSyntax
                 or ParameterSyntax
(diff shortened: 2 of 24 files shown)


TimothyMakkison:
### Before
| Method                        | Mean     | Error   | StdDev  | Gen0      | Gen1      | Gen2     | Allocated |
|------------------------------ |---------:|--------:|--------:|----------:|----------:|---------:|----------:|
| Default_CodeFormatter_Tests   | 102.3 ms | 2.04 ms | 1.91 ms | 3333.3333 | 2333.3333 | 666.6667 |  31.13 MB |
| Default_CodeFormatter_Complex | 173.5 ms | 2.52 ms | 3.99 ms | 5000.0000 | 2000.0000 |        - |  54.57 MB |

### After
| Method                        | Mean      | Error    | StdDev   | Median    | Gen0      | Gen1      | Gen2     | Allocated |
|------------------------------ |----------:|---------:|---------:|----------:|----------:|----------:|---------:|----------:|
| Default_CodeFormatter_Tests   |  95.02 ms | 1.878 ms | 3.793 ms |  93.56 ms | 3000.0000 | 1666.6667 | 333.3333 |  30.81 MB |
| Default_CodeFormatter_Complex | 168.67 ms | 1.949 ms | 2.795 ms | 168.72 ms | 5000.0000 | 2000.0000 |        - |  53.08 MB |

belav (approved):
> Opted to append each item to keep similar formatting

That is the option I would have gone for. I think it keeps it cleaner. I also like consolidating the logic for dealing with `Concat` with 0, 1 or more items. Nice work!

belav (approved):

PR #1545, opened 2025-03-18
perf: replace `StringBuilder` with `ValueListBuilder<char>`

TimothyMakkison:
- Replaces `StringBuilder` with `ValueListBuilder<char>`
- Added some extension methods to make this easier
- Uses `SkipLocalsInit` to avoid the cost to zero `stackalloc char[]`

#### Benchmarks
I have no idea why, all the new changes I make are slower in the benchmarks, even when they should be a straight upgrade 🤔 
I'm running into this issue a lot with #1542
### Before
| Default_CodeFormatter_Tests   | 133.7 ms | 2.65 ms |  3.54 ms | 133.7 ms | 2000.0000 | 1000.0000 |  31.07 MB |
| Default_CodeFormatter_Complex | 284.9 ms | 5.60 ms | 13.95 ms | 279.1 ms | 5000.0000 | 2000.0000 |  54.59 MB |

### After
| Method                        | Mean     | Error   | StdDev   | Median   | Gen0      | Gen1      | Gen2      | Allocated |
|------------------------------ |---------:|--------:|---------:|---------:|----------:|----------:|----------:|----------:|
| Default_CodeFormatter_Tests   | 136.2 ms | 2.69 ms |  5.74 ms | 134.5 ms | 2000.0000 | 1000.0000 |         - |  30.71 MB |
| Default_CodeFormatter_Complex | 285.6 ms | 5.71 ms | 14.93 ms | 278.1 ms | 6000.0000 | 3000.0000 | 1000.0000 |  53.45 MB |

### DIFF
--- a/Src/CSharpier.Core/CSharp/SyntaxPrinter/CSharpierIgnore.cs
+++ b/Src/CSharpier.Core/CSharp/SyntaxPrinter/CSharpierIgnore.cs
@@ -1,6 +1,8 @@
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.RegularExpressions;
 using CSharpier.Core.DocTypes;
+using CSharpier.Core.Utilities;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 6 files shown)


belav (approved):
Did you figure out why your old benchmarks showed this as being slightly slower?

TimothyMakkison:
In retrospect I think I was being overly paranoid. 😓 This is a net positive for memory usage with a negligible improvement to speed. The benchmarks naturally vary a little

belav (approved):

PR #1548, opened 2025-03-21
perf: add `TrimStart(this StringBuilder`

TimothyMakkison:
Prevents duplication of the result string, saving memory in certain cases, ie. a preprocessor directive at the start of a file.

### Before
| Method                        | Mean      | Error    | StdDev   | Gen0      | Gen1      | Allocated |
|------------------------------ |----------:|---------:|---------:|----------:|----------:|----------:|
| Default_CodeFormatter_Tests   |  98.32 ms | 1.954 ms | 2.327 ms | 2000.0000 | 1000.0000 |  31.07 MB |
| Default_CodeFormatter_Complex | 206.40 ms | 3.671 ms | 5.716 ms | 5000.0000 | 2000.0000 |  54.58 MB |

### After
| Method                        | Mean      | Error    | StdDev   | Gen0      | Gen1      | Allocated |
|------------------------------ |----------:|---------:|---------:|----------:|----------:|----------:|
| Default_CodeFormatter_Tests   |  86.67 ms | 1.675 ms | 1.929 ms | 2000.0000 | 1000.0000 |  31.07 MB |
| Default_CodeFormatter_Complex | 182.89 ms | 3.469 ms | 6.766 ms | 5000.0000 | 2000.0000 |  52.47 MB |

### DIFF
--- a/Src/CSharpier/DocPrinter/DocPrinter.cs
+++ b/Src/CSharpier/DocPrinter/DocPrinter.cs
@@ -46,11 +46,11 @@ internal class DocPrinter
 
         this.EnsureOutputEndsWithSingleNewLine();
 
-        var result = this.Output.ToString();
         if (this.PrinterOptions.TrimInitialLines)
         {
-            result = result.TrimStart('\n', '\r');
+            this.Output.TrimStart('\n', '\r');
         }
+        var result = this.Output.ToString();
 
         return result;
     }
--- a/Src/CSharpier/Utilities/StringBuilderExtensions.cs
+++ b/Src/CSharpier/Utilities/StringBuilderExtensions.cs
@@ -21,6 +21,17 @@ internal static class StringBuilderExtensions
     }
 #endif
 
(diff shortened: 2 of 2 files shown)


belav (approved):
FYI I'm heading to the airport real soon so won't be around for the next 8 days. I may be able to get caught up before I take off.

Hoping to release all of these performance changes along with 1.0 in the next couple weeks, thanks again for all your hard work!

TimothyMakkison:
> FYI I'm heading to the airport real soon so won't be around for the next 8 days. I may be able to get caught up before I take off.

Happy journeys, hope you have a good holiday 😄 


> Hoping to release all of these performance changes along with 1.0 in the next couple weeks, thanks again for all your hard work!

Looking forward to it. I'm near the end of my changes. See #1551 for a general feel for the final result. I'll have to find a way to benchmark an actual run of multiple files, to see if I've overlooked something.

PR #1571, opened 2025-04-14
fix: Release Core project

Hona:
just noticed this while browsing around :D

image
image


belav (approved):

belav:
Thanks for taking care of that!

PR #1591, opened 2025-04-25
Use alternative library to parse csharpier- and git-ignore files

kevinboss:
Using gitignore parser library instead of building ignore rules ourselves improves csharpier performance drastically even with large repositories and a lot of ignore rules.

This would introduce a potentially breaking change as the rules are now case-sensitive by default since git is as well.

I had to remove one unit test as we would not parse the ignore-files ourselves anymore as well as alter one which was testing case-insensitive paths.

### DIFF
--- a/Src/CSharpier.Cli/IgnoreFile.cs
+++ b/Src/CSharpier.Cli/IgnoreFile.cs
@@ -1,13 +1,16 @@
-using System.Diagnostics.CodeAnalysis;
 using System.IO.Abstractions;
-using Ignore;
+using System.Text.RegularExpressions;
 
 namespace CSharpier.Cli;
 
 internal class IgnoreFile
 {
     private List<IgnoreWithBasePath> Ignores { get; }
-    private static readonly string[] alwaysIgnored = ["**/node_modules", "**/obj", "**/.git"];
+    private static readonly string alwaysIgnoredText = """
+        **/node_modules
+        **/obj
+        **/.git
+        """;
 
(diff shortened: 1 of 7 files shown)


Src/CSharpier.Cli/IgnoreFile.cs
```diff
@@ -39,58 +37,50 @@ public bool IsIgnored(string filePath)
...
         return new IgnoreFile(ignores);
     }
 
+    private static void AddDefaultRules(IgnoreWithBasePath ignore)
+    {
+        var stringBuilder = new StringBuilder();
+        foreach (var s in alwaysIgnored)
+        {
+            stringBuilder.AppendLine(s);
+        }
```
belav, line 76: `alwaysIgnored` could be a raw string to simplify this and for a minor performance improvement.

belav (changes requested):
Thanks for doing this work!

My quick performance test still showed times that aren't nearly as good as as `0.28.0` on my work repo, but I think there was something else that changed prior to 1.0.0.

I'm okay with the ignore files being case sensitive.

There are a couple of build failures around warnings + formatting, and one minor comment I have. Then this will be good to merge.

kevinboss:
@belav Should be done. I have some trouble getting your coding style rules to run in Rider so I rely on the CI builds to check formatting. It does look fine on my fork and I hope it does as well on the checks here.

belav (approved):
Looks good now, thanks again!

PR #1625, opened 2025-06-06
Xaml Support In Visual Studio

Tiptup300:
Add xaml to default to xml formatter in Core and added to xaml to supported language in visual studio


Src/CSharpier.Core/PrinterOptions.cs
```diff
@@ -44,6 +44,7 @@ public static Formatter GetFormatter(string filePath)
             || filePath.EndsWith(".targets", StringComparison.Ordinal)
             || filePath.EndsWith(".xml", StringComparison.Ordinal)
             || filePath.EndsWith(".config", StringComparison.Ordinal)
+            || filePath.EndsWith(".xml", StringComparison.Ordinal)
```
JoKrus, line 47:
```suggestion
            || filePath.EndsWith(".xaml", StringComparison.Ordinal)
```

Issue #1604, opened 2025-05-09
```xaml``` support as xml files are now supposed to be formatted


JoKrus:
Hi,

I was wondering, if ```*.xaml``` files are also planned to be supported for formatting as they are also xml structured files that are currently not being formatted by csharpier.

Is there a reason for it missing?

belav:
This was an oversight on my part because I'm not familiar with xaml. I see no reason why CSharpier can't format them.

I'll include them by default in the next minor release.

For now you can use the `overrides` to include `xaml` files - https://csharpier.com/docs/Configuration#configuration-overrides

JoKrus:
Thanks for the hint for the overrides. It worked nicely with the CLI but at least the Rider Plugin does not seem to accept this config entry.
I have not tested the VS code or VS plugin.

Also a brief issue: In the documentation for json, it is shown as `"files": ["*.cst"]` but this seems to be wrong, I had to change it to `"files": "*.cst"` for it to be a valid config file

Tiptup300:
Visual Studio does not show "Reformat with Csharpier" when you right click within a xaml file.

You have to run it via the cli. 

I also experienced the same issue as JoKrus when configuring using a .csharpierrc

belav:
> Also a brief issue: In the documentation for json, it is shown as `"files": ["*.cst"]` but this seems to be wrong, I had to change it to `"files": "*.cst"` for it to be a valid config file

Sorry about that! I updated the doc - `files` is a glob string pattern not an array.

PR #1626, opened 2025-06-06
Add XAML support to Visual Studio Extension

Tiptup300:
See https://github.com/belav/csharpier/issues/1604


Tiptup300:
@belav 

Was I supposed to put a fully cleaned file as the .test file?

belav:
> @belav
> 
> Was I supposed to put a fully cleaned file as the .test file?

That's correct. Think of them like a snapshot test. The input (test file) is the expected output.

It is also possible to have the input differ from the expected output but that's less common.

PR #1647, opened 2025-07-04
feat: access the internal array of `Stack` using `UnsafeAccessor`

TimothyMakkison:
Avoid `O(n)` lookup by using `UnsafeAccessor` to access the backing array directly.


TimothyMakkison:
Clsoed due to neglible performance improvement

PR #1665, opened 2025-07-19
Supporting stdin-filepath, dealing with gitignore when there is no path specified and handling non-c# files when no path is specified

### LINKED ISSUE #288: Support an option like --stdin-filepath
When CSharpier formats a file from stdin, it considers the current directory to be the location of the file for purposes of finding a config file and ignore file.

We should add an option `--stdin-filepath`
This will be considered the location of the file for purposes of finding the config file and the ignore file.

### LINKED ISSUE #1657: Formating via stdin is being ignored due to gitignore
**Environments:**

- CSharpier Version: 1.0.2
- Running CSharpier via: cli
- Operating System: linux
- .csharpierrc Settings: none
- .editorconfig Settings: none
- .gitignore: `*`

**Steps to reproduce:**

 - Create a `.gitignore` with `*` in it (ignore all files)
 - Write a `foobar.cs` file
 - Try to format it with `csharpier format < ./foobar.cs`

**Expected behavior:**

The formated file should be printed to stdout.

**Actual behavior:**

Nothing is printed.

belav:
closes #288
closes #1657

### DIFF
--- a/Src/CSharpier.Cli.Tests/CliTests.cs
+++ b/Src/CSharpier.Cli.Tests/CliTests.cs
@@ -352,6 +352,39 @@ public class CliTests
     }
 
     [Test]
+    public async Task Format_Should_Format_Piped_File_With_Config_And_Path()
+    {
+        await WriteFileAsync("Stdin/.csharpierrc", "printWidth: 10");
+
+        var formattedContent1 = "var x =\n    _________________longName;\n";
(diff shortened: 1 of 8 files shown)


DinoChiesa:
Can you explain how a person would use this new thing?  It looks like some change was made but I don't see an explanation of what the change was, or how it works.

PR #1667, opened 2025-07-25
docs: fix pre-commit task-runner.json

JamesYFC:
Running the suggested task-runner.json before would match no files, as husky .net does not support the {a,b} splitting pattern.

before:
image
after:
image


belav:
Whoops, that what I get for not testing. Thanks for the fix!

belav (approved):

PR #1681, opened 2025-08-16
Fixing edge case with null coalescing

### LINKED ISSUE #1573: Inconsistencies with null-coalescing wrapping on method chains
csharpier 0.30.6

**Input:**
```csharp
var test4 =
    await context
        .Images.Where(x => x.Id == ImageId.CreateUnique() && x.Created == DateTime.MaxValue)
        .FirstOrDefaultAsync(cancellationToken) ?? throw new Exception("This is an exception message");
```

**Output:**
```csharp
var test4 =
    await context
        .Images.Where(x => x.Id == ImageId.CreateUnique() && x.Created == DateTime.MaxValue)
        .FirstOrDefaultAsync(cancellationToken) ?? throw new Exception("This is an exception message");
```

**Expected behavior:**
```csharp
var test4 =
    await context
        .Images.Where(x => x.Id == ImageId.CreateUnique() && x.Created == DateTime.MaxValue)
        .FirstOrDefaultAsync(cancellationToken)
    ?? throw new Exception("This is an exception message");
```

On its own, this might seem fine, but when compared to a similar method where we modify different parts of the assignment, the inconsistency becomes apparent:

```csharp
// Fully inline: short method chain and short exception message.
var test1 = await context.Images.FirstAsync(cancellationToken) ?? throw new Exception("This is an exception message");

// Break after '=', everything else fits on one line.
var test2_1 =
    await context.Images.FirstOrDefaultAsync(cancellationToken) ?? throw new Exception("This is an exception message");

// Similar to test2_1, slightly longer exception message but still contained in a single line.
var test2_2 =
    await context.Images.FirstAsync(cancellationToken) ?? throw new Exception("This is a longer exception message");

// Breaks after '=', and '??' is on a new line due to method length.
var test3_1 =
    await context.Images.FirstOrDefaultAsync(x => x.Id == ImageId.CreateUnique(), cancellationToken)
    ?? throw new Exception("This is an exception message");

// Similar to test3_1, but it wraps due to message length.
var test3_2 =
    await context.Images.FirstAsync(cancellationToken)
    ?? throw new Exception("This is a much longer exception message which should be long enough to wrap on its own");

// Multiline formatting: Last part of method chain is short and the exception message is short, so '??' is not wrapped, inconsistent with others.
var test4 =
    await context
        .Images.Where(x => x.Id == ImageId.CreateUnique() && x.Created == DateTime.MaxValue)
        .FirstOrDefaultAsync(cancellationToken) ?? throw new Exception("This is an exception message");

// Multiline formatting: Last part of method chain is long, so '??' is wrapped.
var test5_1 =
    await context
        .Images.Where(x => x.Id == ImageId.CreateUnique() && x.Created == DateTime.MaxValue)
        .FirstOrDefaultAsync(x => x.Id == ImageId.CreateUnique(), cancellationToken)
    ?? throw new Exception("This is an exception message");

// Multiline formatting: Exception message is long, so '??' is wrapped.
var test5_2 =
    await context
        .Images.Where(x => x.Id == ImageId.CreateUnique() && x.Created == DateTime.MaxValue)
        .FirstOrDefaultAsync(cancellationToken)
    ?? throw new Exception("This is a much longer exception message which should be long enough to wrap on its own");
```

I also tried without assignment and swapped out `throw` for a static function with same results.

belav:
closes #1573

### DIFF
--- a/Src/CSharpier.Core/CSharp/SyntaxPrinter/SyntaxNodePrinters/BinaryExpression.cs
+++ b/Src/CSharpier.Core/CSharp/SyntaxPrinter/SyntaxNodePrinters/BinaryExpression.cs
@@ -84,15 +84,46 @@ internal static class BinaryExpression
             var binaryOnTheRight = binaryExpressionSyntax.Kind() == SyntaxKind.CoalesceExpression;
             if (binaryOnTheRight)
             {
-                // if the code is NOT of the form below group around Left + Line. Else no group keeps the ?? next to the )
-                /*
-                    CallMethod(
-                        longParameters,
-                        longParameters
-                    ) ?? expression
-                */
+                var chain = 0;
+                var possibleInvocation = binaryExpressionSyntax.Left;
+                while (possibleInvocation is not null)
+                {
+                    if (possibleInvocation is InvocationExpressionSyntax invocationExpression)
+                    {
+                        possibleInvocation = invocationExpression.Expression;
+                        chain++;
+                    }
+                    else if (
+                        possibleInvocation
+                        is MemberAccessExpressionSyntax memberAccessExpressionSyntax
+                    )
+                    {
+                        possibleInvocation = memberAccessExpressionSyntax.Expression;
+                        chain++;
+                    }
+                    else if (
+                        possibleInvocation
+                        is ConditionalAccessExpressionSyntax conditionalAccessExpressionSyntax
+                    )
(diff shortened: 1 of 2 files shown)


belav:
There are a number of edge cases to deal with, this may need a different fix - https://github.com/belav/csharpier-repos/pull/146/files

Issue #205, opened 2021-05-17
Break before operator on expression bodied methods and variable declarations.


belav:
I think we should consider breaking before the operator for the following statements.

```c#
public void Serialize(ReadOnlySpan<byte> value)
    => InternalSerialize(Encoding.Utf8.GetString(value));
    
var someReallyLongName
    = "someReallyLongValueThatBreaks11111111111111111111111111111111";
```
As compared to
```c#
public void Serialize(ReadOnlySpan<byte> value) =>
    InternalSerialize(Encoding.Utf8.GetString(value));
    
var someReallyLongName =
    "someReallyLongValueThatBreaks11111111111111111111111111111111";
````

**Comment by respel** [comment]
Both look OK to me visually. Also, prettier uses the latter.

Do you have a concrete case in mind where we can benefit from using the former?

**Comment by shocklateboy92** [comment]
Don't have a super strong preference, but I'm curious how you'd handle expression bodied lambdas like:
```cs
someLongAssObject.SomeLongAssFunction((someLongAssArg, someLongerAsserArg) => someLongAssArg.SomeLongAssMethod(someLongerAsserArg.SomeLongAssProperty))
```

respel:
@belav Interesting articles. 

There's one small catch though. In the links you gave, they are only talking about arithmetic and boolean operators, not assignment. Same goes for `black`.

martindisch:
I tested a change that should do this for expression-bodied methods in #1693. I'm not fully convinced that the implementation holds up for all scenarios if you look at the tests and am happy to close it again. Feel free to use the PR in support of the discussion here and change it as you see fit.

PR #1693, opened 2025-09-16
Break before operator on expression-bodied methods

martindisch:
An experimental change to support the discussion in #205. I have no strong feelings about whether this should be merged, but wanted to put it out there.

### DIFF
--- a/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/AssignmentExpressions.test
+++ b/Src/CSharpier.Tests/FormattingTests/TestFiles/cs/AssignmentExpressions.test
@@ -69,18 +69,18 @@ class ClassName
                 );
     }
 
-    public SomeExpressionBodyMethod() =>
-        nonChainFormatting = anotherVariable1 = (
-            someCondition________________________________
-            || someOtherCondition___________________________
-        );
+    public SomeExpressionBodyMethod()
+        => nonChainFormatting = anotherVariable1 = (
+                someCondition________________________________
+                || someOtherCondition___________________________
+            );
 
-    public SomeExpressionBodyMethod() =>
-        chainFormatting =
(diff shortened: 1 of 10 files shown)


Src/CSharpier.Tests/FormattingTests/TestFiles/cs/AssignmentExpressions.test
```diff
@@ -69,18 +69,18 @@ class ClassName
...
-            anotherVariable2 =
-                (
-                    someCondition________________________________
-                    || someOtherCondition___________________________
-                );
+    public SomeExpressionBodyMethod()
+        => nonChainFormatting = anotherVariable1 = (
+                someCondition________________________________
+                || someOtherCondition___________________________
+            );
```
martindisch, line 76: There is an extra space before the closing parenthesis here (check against where `nonChainFormatting` starts above) which freaks me out a bit.

domn1995:
I'm personally not a fan of this change as it would add an extra level of indentation that would otherwise not be there as can be seen in many of the snapshot diffs in this PR like [this](https://github.com/belav/csharpier/pull/1693/changes#diff-56d492ddd7f3bcdb6aec53bb142d419589c644c9fdfad0c09e9a1c0a7c3a682eR74) and [this](https://github.com/belav/csharpier/pull/1693/changes#diff-9a37ca2a93cd6f9e3dd36d6abbc8801a08d3cc604b0e3a546a544952b83e06a8R5).

This format would also strongly clash against lambdas with statement bodies like

```cs
var foo = () =>
{
    return 1;
};

// versus

var foo = ()
    =>
    {
        return 1;
    };
  
// or maybe this?

var foo = ()
    => {
        return 1;
      };
   // ^ where should this closing brace even go?
```

PR #1696, opened 2025-09-21
Fix condition causing FirstTargetFramework build property erasure

tdg5:
Many of my test projects fail to build when `csharpier` is added to them. They fail with this message:

```
Build FAILED.

"/Users/danny/src/tdg5/csharpier-build-issue-reproduction/App.Tests.csproj" (Rebuild target) (1:7) ->
"/Users/danny/src/tdg5/csharpier-build-issue-reproduction/App.Tests.csproj" (CSharpierFormatInner target) (1:8) ->
  /usr/local/share/dotnet/sdk/9.0.300/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.RuntimeIdentifierInference.targets(201,38):
    error MSB4184: The expression "[MSBuild]::VersionLessThan('', 6.0)" cannot be
    evaluated. Version string was not in a correct format.
    [/Users/danny/src/tdg5/csharpier-build-issue-reproduction/App.Tests.csproj]
```

For some reason, I only run into this problem with my test projects that include `xunit.v3`, but I'm fairly certain this is a `csharpier` issue.

A minimal reproduction of the issue can be found at [tdg5/csharpier-build-issue-reproduction](https://github.com/tdg5/csharpier-build-issue-reproduction).

I believe the issue is that the `FirstTargetFramework` build property is getting clobbered due to a typo in a condition that's meant to check if `FirstTargetFramework` already has a value, but fails to do so because of the typo.

This PR fixes the typo and my build failures.

Let me know if I can answer any questions or provide any other context. Thanks!

### DIFF
--- a/Src/CSharpier.MsBuild/build/CSharpier.MsBuild.targets
+++ b/Src/CSharpier.MsBuild/build/CSharpier.MsBuild.targets
@@ -10,7 +10,7 @@
     <CSharpierCommand Condition="'$(CSharpier_Check)' == 'true'">check</CSharpierCommand>
     <CSharpierArgs Condition="'$(CSharpier_LogLevel)' != ''">$(CSharpierArgs) --log-level $(CSharpier_LogLevel)</CSharpierArgs>
     <FirstTargetFramework Condition=" '$(TargetFrameworks)' == '' ">$(TargetFramework)</FirstTargetFramework>
-    <FirstTargetFramework Condition=" '$(FirstTargetFrameworks)' == '' ">$(TargetFrameworks.Split(';')[0])</FirstTargetFramework>
+    <FirstTargetFramework Condition=" '$(FirstTargetFramework)' == '' ">$(TargetFrameworks.Split(';')[0])</FirstTargetFramework>
     <NullOutput>NUL</NullOutput>
   </PropertyGroup>
 


tdg5:
For anyone else running into this issue, a workaround can be to define both `TargetFramework` and `TargetFrameworks`. I believe most build tools are supposed to ignore `TargetFrameworks` if `TargetFramework` is defined, so hopefully other build tools would not be affected by the additional `TargetFrameworks` property. 🤞🏻 

```
<TargetFramework>net9.0</TargetFramework>
<TargetFrameworks>$(TargetFramework)</TargetFrameworks>
```

tdg5:
Another workaround for this problem that might come with less side effects is to set a value for `FirstTargetFrameworks` so the buggy condition is skipped (as it should be):

```
<TargetFramework>net9.0</TargetFramework>
<FirstTargetFrameworks>workaround-for-csharpier-pr-1696</FirstTargetFrameworks>
```

belav (approved):
I was not able to reproduce this but this is definitely a typo, and I don't know why things would be working with the typo there. Thanks for finding the fix!

corngood:
I hit this specifically when building on mac, because that particular version check only applies when the target is osx-arm64.  I think that's probably why not everyone was able to reproduce it.

PR #1703, opened 2025-10-11
perf: use overload for `Any( SyntaxTriviaList` to prevent allocations

TimothyMakkison:
Semi reverts #1572 which overwrote #1485, I assume this was accidental. I noticed that you prefer explicit `Enumerable.Any` usage, should I convert this PR to use `ListExtensions.Any`?

## Benchmarks
Saves around 15% of memory usage
### Before
| Default_CodeFormatter_Tests   | 128.0 ms | 2.44 ms | 3.43 ms | 3000.0000 | 1000.0000 |  34.57 MB |
| Default_CodeFormatter_Complex | 260.3 ms | 5.11 ms | 7.17 ms | 5000.0000 | 2000.0000 |  53.05 MB |

### After
| Method                        | Mean     | Error   | StdDev   | Gen0      | Gen1      | Allocated |
|------------------------------ |---------:|--------:|---------:|----------:|----------:|----------:|
| Default_CodeFormatter_Tests   | 126.0 ms | 2.27 ms |  5.45 ms | 2000.0000 | 1000.0000 |  28.02 MB |
| Default_CodeFormatter_Complex | 245.2 ms | 4.63 ms | 10.44 ms | 4000.0000 | 2000.0000 |  43.88 MB |


belav (approved):
I'm not really sure how these ended up using `Enumerable.Any` as it is not something I prefer. I much prefer the extension methods especially if using them is more efficient!

PR #1713, opened 2025-10-19
fix: comment break indentation of simple lambdas

### LINKED ISSUE #1147: Incorrect formatting for simple lambda expression with comment
**Input:**
```C#
services.AddMetricsTrackingMiddleware(
    // Comment
    opts =>
    {
        opts.IgnoredHttpStatusCodes = Enumerable.Range(300, 9).ToArray();
    }
);
```
**Output:**
```C#
services.AddMetricsTrackingMiddleware(
    // Comment
    opts =>
{
    opts.IgnoredHttpStatusCodes = Enumerable.Range(300, 9).ToArray();
});
```
**Expected behavior:**
```C#
services.AddMetricsTrackingMiddleware(
    // Comment
    opts =>
    {
        opts.IgnoredHttpStatusCodes = Enumerable.Range(300, 9).ToArray();
    }
);
```

I sure this bug is caused by my changes in #1066. I totally forgot about cases with comments

TimothyMakkison:
Fixes #1147, again this is my first time fixing bugs, no idea if I'm doing it right

Lambdas seem pretty all over the place, I'm pretty sure `() =>` has some issues as well

### DIFF
--- a/Src/CSharpier.Core/CSharp/SyntaxPrinter/ArgumentListLikeSyntax.cs
+++ b/Src/CSharpier.Core/CSharp/SyntaxPrinter/ArgumentListLikeSyntax.cs
@@ -15,7 +15,10 @@ internal static class ArgumentListLike
     )
     {
         Doc? args;
-        if (arguments is [{ Expression: SimpleLambdaExpressionSyntax lambda1 }])
+        if (
+            arguments is [{ Expression: SimpleLambdaExpressionSyntax simpleLambda }]
+            && !simpleLambda.GetLeadingTrivia().Any(o => o.IsComment())
+        )
         {
             var groupId = context.GroupFor("LambdaArguments");
             args = Doc.Concat(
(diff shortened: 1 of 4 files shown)


belav:
This has possibly a single edge case that it still needs to handle - https://github.com/belav/csharpier-repos/pull/152/files

I think those all are this same issue
```c#
// input and expected output
        Assert.Collection(
            outputSteps,
            // First source output for diagnostics is unchanged.
            step => Assert.Equal(IncrementalStepRunReason.Unchanged, step.Reason),
            // Second source output for generated code is changed.
            step => Assert.Equal(IncrementalStepRunReason.Modified, step.Reason)
        );

// current changes cause it to do this
        Assert.Collection(
            outputSteps,
            // First source output for diagnostics is unchanged.
            step =>
                Assert.Equal(IncrementalStepRunReason.Unchanged, step.Reason),
            // Second source output for generated code is changed.
            step =>
                Assert.Equal(IncrementalStepRunReason.Modified, step.Reason)
        );

**Comment by TimothyMakkison** [comment]
Okay, I think I've fixed this issue but have replaced it with a new one, see `MemberChains.test`. I don't think this is a new issue #1613 parenthesized lambda expressions have the same behaviour.

I assume the fix is found in `ArgumentListLike.Print` but it may be beyond my current understanding, I'd appreciate your help here.

## Before
```C#
Select______________________________________(superLongName_________________________ =>
    true
);
```
## After
```C#
Select______________________________________(superLongName_________________________ => true
); // weird parentheses,
```

PR #1729, opened 2025-11-01
Update Dependencies

stefannikolei:
Updates all possible Dependencies. Skipped DiffEngine because of Breaking changes.

Removed System.Threading.Tasks.Extensions its not needed.

Enabled Transitive Pinning (https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#transitive-pinning) to make sure the same package is used in every project


stefannikolei:
> Looks good! I validated that DiffEngine is keeping the .actual.cs file as readonly and only allowed edits to the test files.

Ty for checking. 

Gonna look now to get it to the latest. Didn't want to go all the way. Wanted to fix the breaking change which was standing in my way last time.

PR #1737, opened 2025-11-04
perf: fast comparison using `FullSpan` and source code

TimothyMakkison:
- Compare the old and formatted code before using `SyntaxNodeComparer` (ie new version of csharpier, no changes but hash is cache is invalidated)
- Skip comparison if two `TypeDeclaration` aka (classes, struct, interfaces etc) or `MethodDeclaration` are identical
    - This could be done for all `SyntaxNodes` but may slow things down if most nodes have changed

The CLI spends up to 30% of the time doing comparisons, with this, we can skip cases with identical source code, and only compare types/methods that have changed. The performance degradation for lots of changes (i.e. first time a project is formatted) shouldn't be that great as `CompareFullSpan` is basically a fast `SequenceEqual` and only used for type declarations and methods.


belav (approved):
Looks good!

PR #1755, opened 2025-11-12
Update compatible C# version in index.js

davidwalschots:
C# 14 support was added in #1658, but the website was not updated to reflect this.


belav (approved):
Thanks for catching this!

Issue #1739, opened 2025-11-06
Run on save being disabled in Rider plugin causes toil


martindisch:
We're very happily using CSharpier in a large engineering organization. We have many .NET services living in different repositories, meaning the average developer is regularly checking out new repos. "Run on save" is disabled by default in the Rider plugin and it uses Rider's layered configuration, so settings are only stored per solution and it's not possible to globally enable run on save once and forever.

For us this means we regularly have developers committing unformatted code when they check out a new repository and forget to enable run on save. These mistakes are caught in CI, but they happen frequently and lead to unnecessary frustration.

I see two ways to make this a better experience:
1. Enable run on save by default. This makes sense to me, because I expect most people installing a code formatter plugin would want it to automatically format their code. It's also incredibly easy to implement, so I made a PR in case you think this would be a good idea https://github.com/belav/csharpier/pull/1738
2. Make it possible to globally configure the plugin, not only on a solution level. I'm not familiar with how the layered configuration has to be implemented to allow this.

Thanks for this amazing tool, it's a great developer experience booster :)

belav:
> Enable run on save by default. This makes sense to me, because I expect most people installing a code formatter plugin would want it to automatically format their code. It's also incredibly easy to implement, so I made a PR in case you think this would be a good idea https://github.com/belav/csharpier/pull/1738

I think that would probably frustrate more people that it benefits. VSCode, prettier plugins, etc - they all seem to be opt in for format on save.

> Make it possible to globally configure the plugin, not only on a solution level. I'm not familiar with how the layered configuration has to be implemented to allow this.

I'm on board with this and didn't even realize it didn't work that way until I tested it just now. I may have missed something when I implemented the settings that is required to allow them to be stored in the "this computer" layer.

martindisch:
Got it, will close the PR then. Making it globally configurable would be much appreciated, after all it already works this way for the Visual Studio plugin.

belav:
I tweaked #1762 a bit and pushed out a 2.3.0 update that has a global run on save option. It sometimes takes jetbrains a couple of days to review and approve before it will be available.

martindisch:
Thank you.

TimothyMakkison:
@belav sorry to bother you, but have you [read my comment?](https://github.com/belav/csharpier/pull/1798#issuecomment-3704190944) I'm concerned that users may think the plugin is broken or malfunctioning.

Hopefully I've misunderstood something and this won't be an issue

PR #1762, opened 2025-11-24
feat: add global run on save option

TimothyMakkison:
Full disclosure, I rarely use Java, no idea if this works, not sure how to debug plugins. I'm hoping the CI has a test for this.

Edit: Looks like I need to install Java 😭 

#1739

### DIFF
--- a/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierSettingsComponent.java
+++ b/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierSettingsComponent.java
@@ -20,9 +20,9 @@ public class CSharpierSettingsComponent implements SearchableConfigurable {
 
     private final Project project;
     private final String[] runOnSaveOptions = new String[] {
+        "True",
         "Use Global Setting",
         "False",
-        "True",
     };
     private ComboBox<String> runOnSaveComboBox = new ComboBox<String>(runOnSaveOptions);
     private JBCheckBox globalRunOnSaveCheckBox = new JBCheckBox("Run on Save (Global)");
@@ -143,11 +143,11 @@ public class CSharpierSettingsComponent implements SearchableConfigurable {
         String selectedItem = (String) runOnSaveComboBox.getSelectedItem();
 
         if (selectedItem == runOnSaveOptions[0]) {
-            return SolutionRunOnSaveOption.UseGlobalSetting;
-        } else if (selectedItem == runOnSaveOptions[1]) {
-            return SolutionRunOnSaveOption.False;
-        } else if (selectedItem == runOnSaveOptions[2]) {
             return SolutionRunOnSaveOption.True;
+        } else if (selectedItem == runOnSaveOptions[1]) {
+            return SolutionRunOnSaveOption.UseGlobalSetting;
+        } else if (selectedItem == runOnSaveOptions[2]) {
+            return SolutionRunOnSaveOption.False;
         }
 
         this.logger.debug("invalid runOnSaveComboBox selection: " + selectedItem);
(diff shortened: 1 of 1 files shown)


TimothyMakkison:
image

Got it working, setting persists between projects now 👍 

I opted to put both settings under the same header, did wonder if it should say `Run on save (Enabled globally)`.

belav:
Intellij has the idea of layers so implementing this as a global setting needs to be done that way instead of adding another option. I'm not sure what needs to be done but a global setting is just one that can be saved at the "my computer" layer. And then if needed the csharpier plugin code needs to make sure the more specific layer setting takes priority.

VS has a separate global setting because at least for 2022 and earlier there was nothing built-in that allowed it. I haven't looked into 2026 yet to see how the new settings work but I'm hoping it is closer to vscode.

martindisch:
Thanks for starting work on this :-) This will be super useful, because in our organization it unfortunately happens frequently enough that people open a new solution, don't have run on save enabled for it, and get frustrated by a build error in CI. The pain has gotten bad enough that our platform team is planning to build autoformatting into CI that will create an "autofix" branch to merge into the PR branch on formatting errors.

If we can get the Rider plugin to reliably work in _all_ repositories one opens, we can make this a much better experience.

martindisch:
I gave it another shot in https://github.com/belav/csharpier/pull/1766 and it seems to work. It doesn't use layers though, just changes everything to global settings.

TimothyMakkison:
> Intellij has the idea of layers so implementing this as a global setting needs to be done that way instead of adding another option. I'm not sure what needs to be done but a global setting is just one that can be saved at the "my computer" layer. 

I tried doing this but I couldn't find any examples - online documentation wasn't very helpful with a lot of dead links/github repos, AI's were a little better. I'm curious to know if @martindisch found anything relevant when doing #1766.
I found documentation for layered settings with Resharper, but nothing for jvm rider plugins.

> And then if needed the csharpier plugin code needs to make sure the more specific layer setting takes priority.

I could modify this PR to do this, I can't just have the project settings always override the global one, because the project setting defaults to `false`, preventing the global on save from ever doing anything. It might be possible to migrate to a newer system where we have a new Enum for project on save settings `True, False (default), Override False`. With the current `False` -> `False (default)` with users opting to use `Override False` as needed.


> VS has a separate global setting because at least for 2022 and earlier there was nothing built-in that allowed it. I haven't looked into 2026 yet to see how the new settings work but I'm hoping it is closer to vscode.

[Docs here](https://learn.microsoft.com/en-us/visualstudio/extensibility/visualstudio.extensibility/settings/settings?view=visualstudio), not sure if `VisualStudio.Extensibility` is as backwards compatible as the current version.

belav:
Based on the other PR plus some work I tried today, it doesn't seem like there is any way to get settings into `This Computer`. At least not while this is an intellij plugin.

Another possibility would be to have a whole section called `CSharpier (global settings)` and include all of the settings there. I don't know if it is worth that extra effort. So I think sticking with this makes sense unless you or @martindisch feel strongly about having a whole section with all of the options available to set globally.

I do think this should modify RunOnSave to be a nullable boolean. That way someone can always override the global value whether it is true or false.

The way VS handles it is a nullable boolean behind the scenes but a drop down with an empty option for the null value.

image

TimothyMakkison:
This is my current approach, I've included some migration logic to avoid the issues mentioned in my previous comment.
Its backed by an enum not sure if this is the best approach. I'd appreciate some naming advice
image

TimothyMakkison:
I've implemented my previous idea. I opted to add a new `ProjectRunOnSaveOption` setting. If the previous `runOnSave` was `True`, `projectRunOnSaveOption` becomes `ProjectRunOnSaveOption.RunOnSave` and resets `runOnSave`. Otherwise we default to `ProjectRunOnSaveOption.SoftNeverRun` for new and existing projects. This way Global run on save will be able to override projects by default but we also have the option to override this behaviour at the project level with `Never run on save`.


- Naming is not final, I'm not sure if users would understand the difference between `Don't run on save` and `Never run on save`.
- I do not regularly write Java any feedback would be appreciated
- It might be worth updating `runOnSave` when `projectRunOnSaveOption` changes on the off change someone downgrades versions, I doubt this would be a common problem 
- IntelliJ does have a build in settings migrator, I have no idea how to use this as the documentation is terrible. I'm not convinced it would actually be useful

TimothyMakkison:
Ok, I've updated this PR after the @belav's [feedback](https://github.com/belav/csharpier/pull/1766#issuecomment-3700056923)

- updated the front end
- changed varaible names to use solution
- reordered the dropdown layout.

image


Is it possible to release a beta version of the csharpier plugin? I'm a little worried about it breaking for everyone despite my best attempts with migration

martindisch:
Thanks for pushing this Timothy!

PR #1763, opened 2025-11-26
Fix Stream closed exception in Rider plugin

rcdailey:
Fixes the `java.io.IOException: Stream closed` exception that occurs when the CSharpier process fails to start in the Rider plugin.

## Problem
The `onExit` handler was attempting to read from the error stream after the process had exited, but streams are automatically closed when a process terminates.

## Solution
Read stderr continuously in a background thread during the process lifetime and capture output to a `StringBuilder`. The `onExit` handler then references the captured output instead of trying to read from the closed stream.

## Approach
This follows the pattern used in JetBrains' own code (`ForkedByModuleSplitter`) for handling process streams safely, which uses separate named threads to read stdout and stderr while the process is running.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,24 @@
+name: "Close stale issues and PRs"
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+jobs:
+  stale:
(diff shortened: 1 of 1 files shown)


belav (approved):
LGTM! I'll get a release out with this change shortly.

Issue #1670, opened 2025-07-31
Bug: insert_final_newline = false from .editorconfig is ignored


kudima03:
CSharpier currently always inserts a final newline at the end of .cs files, even when .editorconfig explicitly sets:
`insert_final_newline = false`

To reproduce:

1. Create a `.editorconfig` with:
```
[*.cs]
insert_final_newline = false
```
2. Format any .cs file using CSharpier:
`csharpier format .`
3. Observe that the file ends with a newline regardless of the setting.

When using `dotnet format`, the same .editorconfig setting in the same project is respected correctly — the file is saved without a trailing newline, as intended.

Environment:

- CSharpier version: 1.0.3
- Windows 11
- PowerShell 7.5.2

This makes it difficult to maintain consistency with other tools and repositories that intentionally avoid final newlines.

belav:
CSharpier's behavior is intentional and will not be changed. It is a best practice to end files with a new line.

kudima03:
@belav So CSharpier cannot be fully configured from .editorconfig? Only partially?

kudima03:
@belav I explored [documentation](https://csharpier.com/docs/Configuration) more carefully, it is not obvious from fast view that only 8 parameters can be applied from `.editorconfig`. May be it is worth to be marked more brightly.

PR #1764, opened 2025-11-28
feat(editorconfig): add support for insert_final_newline

cybercyst:
Currently `csharpier` insists in always adding a new line.

I found that there are existing issues that have been closed regarding this:

- https://github.com/belav/csharpier/issues/1670

Currently this causes a bit of problems in my workflow. The organization I work for sets `.editorconfig` with `insert_final_newline` set to false. This causes my PRs to often fail when linting is run, as `dotnet format` respects the `.editorconfig` file, and since the files have newlines, `dotnet format --verify-no-changes` fails.

I personally agree that we should have newlines, but I'm not in a position to change how the orginization I work for sets their configurations.

### DIFF
--- a/Src/CSharpier.Cli/EditorConfig/EditorConfigSections.cs
+++ b/Src/CSharpier.Cli/EditorConfig/EditorConfigSections.cs
@@ -53,6 +53,11 @@ internal class EditorConfigSections
             printerOptions.EndOfLine = endOfLine;
         }
 
+        if (resolvedConfiguration.InsertFinalNewline is { } insertFinalNewline)
+        {
+            printerOptions.InsertFinalNewline = insertFinalNewline;
+        }
+
         return printerOptions;
     }
 
@@ -63,6 +68,7 @@ internal class EditorConfigSections
         public int? TabWidth { get; }
         public int? MaxLineLength { get; }
         public EndOfLine? EndOfLine { get; }
+        public bool? InsertFinalNewline { get; }
         public string? Formatter { get; }
 
         public ResolvedConfiguration(List<Section> sections)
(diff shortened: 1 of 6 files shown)


belav:
CSharpier is not going to support this option. You are welcome to fork it if you can't convince your organization to change.

cybercyst:
I'm disappointed, but thanks for reviewing! Also thanks for Csharpier. It is an incredibly useful tool!

belav:
Thanks for being understanding! I was going to adopt the one change I commented on just now but then realized why the code is the way it is.

robv8r:
> CSharpier is not going to support this option. You are welcome to fork it if you can't convince your organization to change.

May I please ask why? I'm honestly curious and don't want to miss a learning opportunity.

**Just to recap**

If I correctly understand the intent of the PR, the author would like CSharpier to support the following `.editorconfig` snippet and remove any trailing newlines from XML files:

```editorconfig
[*.xml]
insert_final_newline = false
```

If I correctly understand the above quoted response, CSharpier is not going to support the above option.

**Questions**

- Is the above snippet **not** a recommended approach?  If so, why?
- Is there a different reason CSharpier chose not to support xml files "without final new lines"
- Is there a different way of accomplishing this? (Supporting XML files "without final new lines")

Thanks
--rob

belav:
@robv8r

CSharpier ends all files with a single new line. It is a standard practice - https://www.google.com/search?q=why+should+I+end+a+file+with+one+new+line

CSharpier will not add any options beyond the few that it already has. If someone doesn't like the way it does things they are welcome to fork it or not use it.

PR #1766, opened 2025-12-03
Save Rider plugin settings globally

### LINKED ISSUE #1739: Run on save being disabled in Rider plugin causes toil

martindisch:
This changes settings storage from the project level to the global settings, allowing users to configure CSharpier once, and not for each solution they open.

I was able to test it locally and it works, but I'm not an expert on IntelliJ plugins and can't entirely rule out that this breaks something somewhere. So far the only observed effect should be that people will have their settings reset to defaults with the update and will have to reconfigure the plugin once.

Closes #1739.

### DIFF
--- a/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessProvider.java
+++ b/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierProcessProvider.java
@@ -271,7 +271,7 @@ public class CSharpierProcessProvider implements DocumentListener, Disposable, I
             ICSharpierProcess csharpierProcess;
             if (
                 Semver.gte(version, serverVersion) &&
-                !CSharpierSettings.getInstance(this.project).getDisableCSharpierServer()
+                !CSharpierSettings.getInstance().getDisableCSharpierServer()
             ) {
                 csharpierProcess = new CSharpierProcessServer(customPath, version, this.project);
             } else if (Semver.gte(version, "0.12.0")) {
@@ -279,7 +279,7 @@ public class CSharpierProcessProvider implements DocumentListener, Disposable, I
 
                 if (
                     Semver.gte(version, serverVersion) &&
-                    CSharpierSettings.getInstance(this.project).getDisableCSharpierServer()
+                    CSharpierSettings.getInstance().getDisableCSharpierServer()
                 ) {
                     this.logger.debug(
                             "CSharpier server is disabled, falling back to piping via stdin"
--- a/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierSettings.java
+++ b/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierSettings.java
@@ -1,9 +1,9 @@
 package com.intellij.csharpier;
 
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.components.PersistentStateComponent;
 import com.intellij.openapi.components.State;
 import com.intellij.openapi.components.Storage;
-import com.intellij.openapi.project.Project;
 import com.intellij.util.xmlb.XmlSerializerUtil;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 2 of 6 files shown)


belav:
I don't agree with changing this to be global. Anyone that wants format on save in one solution but not another is going to have to flip the setting back and forth every time they switch solutions.

As is, you set it once the first time you clone a repo, and only if there isn't already a team shared setting.

rcdailey:
Is it possible to make this use settings layers so the user can choose what they want (local vs team-shared vs global)? Respectfully belav what you're saying is your opinion and all the more reason for user choice here. Rider has layers to support this flexibility, I'm just not sure if plugins can utilize it.

belav:
> Is it possible to make this use settings layers so the user can choose what they want (local vs team-shared vs global)?

@rcdailey that is my hope. My googling has indicated that it should be possible to implement this to use the global layer. 

> Respectfully belav what you're saying is your opinion

I don't think it is just an opinion and may have not expressed it well.

If this setting is only global then anyone who works in two solutions and wants format on save in one and not the other will be switching that global setting back and forth every time they switch solutions. If the setting stays as is (with local/team shared) then anyone who always wants it on (or off) will have to set that setting once per solution. The global only setting will be a major inconvenience for some users. The current setting is a minor inconvenience for some users.

rcdailey:
> If this setting is only global then anyone who works in two solutions and wants format on save in one and not the other will be switching that global setting back and forth every time they switch solutions

I understand where you are coming from. That is 100% valid and I'm sure it applies to a percentage of users. The point I was trying to make by calling it an opinion is that there is also a percentage of users that prefer not to have to configure this once per solution, and instead have it enabled by default everywhere. I think the author of this PR falls into the latter group.

Either way the ultimate solution IMO is to use Rider's settings layers, and it sounds like that's the approach you'd favor as well. Hopefully the OP is open to the work for this.

martindisch:
I have some bad news unfortunately. Given that this is a major inconvenience for teams in our company, I tried to give it another shot to get true layered config working. As @TimothyMakkison pointed out in https://github.com/belav/csharpier/pull/1762#issuecomment-3609413268, the documentation for Plugin configuration is atrocious, so I attempted to coax various LLMs into coming up with a solution. They always ended up with what I would call "user space layered config", meaning a layered config implemented on the plugin-level, for example with a dropdown in the plugin settings where a user can choose where to save settings (global or solution), or entirely duplicating the plugin settings into "CSharpier global" and "CSharpier local". All of these are unergonomic, subpar solutions and lead me to believe that maybe plugins really can't use layered config (only the IDE itself), or that it is entirely undocumented to a degree where nobody can find it. I also haven't seen another Rider plugin so far using true layered config where we could copy from.

I understand the hesitation to make this a global setting, given we have no data on what percentage of users would prefer which solution.

rcdailey:
@martindisch is this helpful? https://gist.github.com/rcdailey/a2d6845082cddfd2ae3475b7b7cd3527

TimothyMakkison:
I believe that documentation is for a resharper rider plugin written in C#. TMK the csharpier rider plugin is an intellij platform plugin written in Java.

martindisch:
Sadly not really, because as I understand it, this example is specific to Resharper and written in C#, while CSharpier is a "regular" Java plugin. But then again, I have zero expertise, maybe somebody else knows better.

rcdailey:
My understanding is that this is not a language specific concern. This documentation shows Application and Project level settings, which I assume maps to the respective settings layers in Rider: https://plugins.jetbrains.com/docs/intellij/settings-guide.html#declaring-application-settings

I lack actual experience with this though. Maybe someone ought to just attempt an implementation and test to verify?

TimothyMakkison:
I've updated #1762 to use a dropdown box for project level settings, giving a fine grained control for users at the project level. I'd appreciate feedback on the naming.

image

belav:
It isn't clear to me what the difference between `Don't run on save` and `Never run on save` is. Maybe just keep it simple with

- True
- False
- Use Global Setting

I'd also call it `Solution` instead of project. Rider already uses that verbiage for the setting layers in Rider.

PR #1770, opened 2025-12-07
fix: Null coalescing not own its line with complex left-hand side

### LINKED ISSUE #1769: The null coalescing operator is grouped in an unexpected place
CSharpier Version: 1.2.1

**Input:**

```cs
var x =
    someValue.SomeCall().SomeProperty.SomeProperty
    ?? someValue.SomeCall().SomeProperty.SomeProperty;

var x =
    someValue.SomeCall().SomeProperty?.SomeCall().SomeProperty
    ?? someValue.SomeCall().SomeProperty?.SomeCall().SomeProperty;

var x =
    someValue.SomeCall().A_______.B_______.C_______
    ?? someValue.SomeCall().A_______.B_______.C_______;
```

**Output:**

```cs
var x =
    someValue.SomeCall().SomeProperty.SomeProperty ?? someValue
        .SomeCall()
        .SomeProperty.SomeProperty;

var x =
    someValue.SomeCall().SomeProperty?.SomeCall().SomeProperty ?? someValue
        .SomeCall()
        .SomeProperty?.SomeCall()
        .SomeProperty;

var x =
    someValue.SomeCall().A_______.B_______.C_______ ?? someValue
        .SomeCall()
        .A_______.B_______.C_______;
```

**Expected behavior:**

```cs
var x =
    someValue.SomeCall().SomeProperty.SomeProperty
    ?? someValue.SomeCall().SomeProperty.SomeProperty;

var x =
    someValue.SomeCall().SomeProperty?.SomeCall().SomeProperty
    ?? someValue.SomeCall().SomeProperty?.SomeCall().SomeProperty;

var x =
    someValue.SomeCall().A_______.B_______.C_______
    ?? someValue.SomeCall().A_______.B_______.C_______;
```

From #1681, some null coalescing operators no longer have their right-hand expression on its own line.

In the `chain > 3` condition, the left-hand expression becomes a grouping target just by being slightly more complex.

I confirmed that the first and second examples above can be avoided by converting the condition to `chain > 4`, but I don't think that's a very good solution.

The reason why changes like #1681 were probably made is that, while addressing the pattern reported in #1573, they also had to handle existing formatting like the example below, and got stuck on how to define the left-hand side.

```cs
var coalesceIsSpecial =
    CallSomething(
        someLongValue__________________________________________,
        someLongValue__________________________________________
    ) ?? someOtherValue;
```

This pattern is tricky to handle, but while listing several plausible examples, I thought that "if the left-hand expression is an `InvocationExpression` and the `MemberAccessExpression` is built solely from property accesses, then group it" might work well.

```cs
var coalesceIsSpecial =
    CallSomething(
        someLongValue__________________________________________,
        someLongValue__________________________________________
    ) ?? someOtherValue;

var x =
    someValue
        .Property.CallLongMethod_____________________________________()
        .CallMethod(
            someLongValue__________________________________________,
            someLongValue__________________________________________
        )
    ?? someOtherValue;

var x =
    someValue.CallMethod(
        someLongValue__________________________________________,
        someLongValue__________________________________________
    ) ?? someOtherValue;

var x =
    someValue
        .Property.CallLongMethod_____________________________________()
        .CallMethod__________()
    ?? throw new Exception();

var x =
    someValue
        .Property.CallLongMethod_____________________________________()
        .CallMethod__________(someParameter)
    ?? throw new Exception();

var x =
    someValue
        .Property.CallLongMethod_____________________________________()
        .CallLongMethod___________________________________________________()
    ?? throw new Exception();
```

ogaken-1:
There was an issue where if the left-hand side of `??` was even slightly complex, the right-hand side would become part of the group and could not have its own line.

I reconsidered the conditions for grouping and changed it from judging the complexity of the left-hand side by a threshold to a condition where "the left-hand side is an InvocationExpression and consists solely of MemberAccessExpression that does not include an InvocationExpression".

fix #1769

### DIFF
--- a/Src/CSharpier.Core/CSharp/SyntaxPrinter/SyntaxNodePrinters/BinaryExpression.cs
+++ b/Src/CSharpier.Core/CSharp/SyntaxPrinter/SyntaxNodePrinters/BinaryExpression.cs
@@ -73,7 +73,11 @@ internal static class BinaryExpression
 
             // This group ensures that something like == 0 does not end up on its own line
             var shouldGroup =
-                binaryExpressionSyntax.Kind() != binaryExpressionSyntax.Parent!.Kind()
+                (
+                    binaryExpressionSyntax.Kind() != SyntaxKind.CoalesceExpression
+                    || ShouldGroupNullCoalescingSyntax(binaryExpressionSyntax)
+                )
+                && binaryExpressionSyntax.Kind() != binaryExpressionSyntax.Parent!.Kind()
                 && GetPrecedence(binaryExpressionSyntax)
                     != GetPrecedence(binaryExpressionSyntax.Parent!)
                 && binaryExpressionSyntax.Left.GetType() != binaryExpressionSyntax.GetType()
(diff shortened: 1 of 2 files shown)


ogaken-1:
Upon reviewing the PR that was created, I noticed there are some oversights. I'll work on the corrections later.

ogaken-1:
After reviewing the results in the csharpier-repos, I have added support for the following:
- ParenthesizedExpressionSyntax
- CastExpressionSyntax { Expression: ParenthesizedExpressionSyntax }
- ConditionalAccessExpressionSyntax

ogaken-1:
@belav 
I made additional changes thinking they would be an improvement, but perhaps they weren't?
I would be grateful for your opinion. Thank you.

belav:
Taking a look at the changes it made I think everythink looks good except this one - https://github.com/belav/csharpier-repos/pull/159/changes/e097ff1d5ab4cfc42a3ca55c92ceb38e3bfbd6c3#diff-bb147afabd02afcf48b4b4c09aad50f6bb117fea0bae21a1b2f1adef34f16beb

Your first pass had `?? false` on a new line but the most recent version keeps it on the same line.

I don't see a test case for it which is probably why it ended up being undone.

Was there something specific you didn't think was an improvement? I do like the way it groups it with the ending paren
```
) ?? result
```

belav:
Ideally I'd have an action for doing the formatting on csharpier-repos - I think I attempted it a while back and it wasn't super straightforward.

ogaken-1:
Added support for when the leftmost element is a parenthesized expression.

PR #1777, opened 2025-12-14
test: Fix failure caused by 1.0 assembly renaming

ogaken-1:
It appears it wasn't detected because it was skipped in CI, etc.

The test run via `dotnet test` was now reporting failures.


belav (approved):
Ah yeah missed this, thanks!

PR #1798, opened 2026-01-01
Implement global setting for run on save

### LINKED ISSUE #1739: Run on save being disabled in Rider plugin causes toil

belav:
closes #1739

### DIFF
--- a/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierSettingsComponent.java
+++ b/Src/CSharpier.Rider/src/main/java/com/intellij/csharpier/CSharpierSettingsComponent.java
@@ -8,7 +8,6 @@ import com.intellij.ui.components.JBLabel;
 import com.intellij.ui.components.JBTextField;
 import com.intellij.util.ui.FormBuilder;
 import java.awt.*;
-import java.util.Arrays;
 import javax.swing.*;
 import org.jetbrains.annotations.Nls;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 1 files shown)


TimothyMakkison:
Hey @belav , I see you switched from strategy of migrating to a new property and instead opted to extend the existing one to be `Boolean`. 

Did you confirm that this won't cause existing projects to ignore the global setting? My [comment here](https://github.com/belav/csharpier/pull/1762#issuecomment-3609413268) goes into more detail.
I'm assuming `Boolean` defaults to null so any newly opened projects will use the global settings and won't have this issue. 

I'm not sure how to test this, but I suspect it will be really annoying for users of it behaves this way

belav:
@TimothyMakkison I confirmed that things all work as I expect them to. If an existing project has a value set it will ignore the global value. That is how it is supposed to work.

PR #1814, opened 2026-02-11
Sarif formatting support

### LINKED ISSUE #1398: Enable error output that matchs what GH actions expect.
It is possible for CSharpier to output errors that provide file names + line numbers to GH actions. Then the GH action will add comments directly to the files in the PR.

Is it possible to make this configurable so that something like azure can also work?

The recent changes to CSharpier.MsBuild probably do the same thing for rider/visualstudio, should CSharpier output the format it expects directly instead?

dragosv:
## Description

Enable error output that matchs what GH actions expect (Sarif)

## Related Issue

Closes #1398 

### Checklist

- [ ] My code follows the project's code style
  - always `var`
  - always `this.`
  - no pointless comments
- [ ] I will not force push after a code review of my PR has started
- [ ] I have added tests that cover my changes

### DIFF
--- a/Src/CSharpier.Cli/ConsoleLogger.cs
+++ b/Src/CSharpier.Cli/ConsoleLogger.cs
@@ -1,10 +1,16 @@
 using Microsoft.Extensions.Logging;
 
+#if !NET9_0_OR_GREATER
+using Lock = object;
+#else
+using Lock = System.Threading.Lock;
+#endif
+
 namespace CSharpier.Cli;
 
 internal class ConsoleLogger(IConsole console, LogLevel loggingLevel) : ILogger
 {
-    private static readonly object ConsoleLock = new();
+    private static readonly Lock ConsoleLock = new();
 
     public void Log<TState>(
         LogLevel logLevel,
@@ -33,7 +39,7 @@ internal class ConsoleLogger(IConsole console, LogLevel loggingLevel) : ILogger
             if (logLevel >= LogLevel.Warning)
             {
                 console.ForegroundColor = GetColorLevel(logLevel);
-                console.Write($"{logLevel.ToString().ToUpper()}: ");
+                console.Write($"{logLevel.ToString().ToUpperInvariant()}: ");
                 console.ResetColor();
             }
 


belav:
Sarif log formatting is a bit overkill and I don't really see the benefit. There are simple annotations that can be used to have GH reference files.

dragosv:
> Sarif log formatting is a bit overkill and I don't really see the benefit. There are simple annotations that can be used to have GH reference files.

Sarif support for the check part of the tool would be really beneficial for quite a few use cases. One is the scenario where the format does not happen on the CI/CD, and the tool only checks that formatting was done on the client, ideally using a commit hook

belav:
> Sarif support for the check part of the tool would be really beneficial for quite a few use cases. One is the scenario where the format does not happen on the CI/CD, and the tool only checks that formatting was done on the client, ideally using a commit hook

What benefit is there to output a large amount of json that isn't human readable vs an exit code of 1 and plain text that tells the user which files aren't formatted?

Based on my understanding the sarif format is for reporting issues and used by static code analysis tools. With csharpier the file is formatted or it is not. And a simple command will format the file. There isn't a need for a detailed report of which files aren't formatted and which line is the problem.

dragosv:
Sarif support would bring parity with dotnet format. 

dotnet format can output results in the SARIF (Static Analysis Results Interchange Format) format. This is incredibly useful for integrating linting and formatting checks into CI/CD pipelines, such as GitHub Actions or Azure DevOps, where the results can be displayed directly in the "Security" or "Code Quality" tabs.

To generate a SARIF report, you use the --report flag and specify the directory where you want the file to be saved.

dotnet format --verify-no-changes --report ./artifacts/reports

belav:
> Sarif support would bring parity with dotnet format.

This is not a goal of csharpier.

Sarif formatting makes sense for dotnet format because it reports on all analyzers. That can include security issues which can't be addressed by just running a single command. If csharpier finds unformatted files the fix is simple. Showing unformated files on a security tab does not make sense. Showing it on a code quality tab maybe makes sense except that the intended usage of csharpier is to format the entire repository. A simple check with a pass fail covers that.

PR #1818, opened 2026-02-16
perf: batch individual file inputs for parallel formatting

ansemb:
### Problem

When passing 200+ individual file paths to CSharpier (e.g. with pre-commit hooks), formatting is significantly slower than passing a directory containing the same files.

> Note: this was run on Mac M1. Showing the ms is only meant as an indication of large difference between inputting a folder and multiple files.

**Directory**:
```
dotnet csharpier format path/to/project/dir
Formatted 567 files in 928ms.
```
**Files**:
```
dotnet csharpier format \
  path/to/project/dir/file1 \
  path/to/project/dir/sub/file2 \
  ...
```

**Before (293 files, individual paths):** Formatted 293 files in 40886ms.
**After (293 files, individual paths):** Formatted 293 files in 859ms.

### Root Cause

In `CommandLineFormatter.FormatPhysicalFiles`, each individual file path was processed inside the input loop with:
1. A **new `OptionsProvider`** per file (expensive — reads `.csharpierrc`, `.editorconfig`, `.gitignore` from disk and walks up directories)
2. A **new `FormattingCache`** per file (reads/writes a separate cache JSON each time)
3. **Sequential** formatting (no parallelism)

In contrast, directory inputs already used a single `OptionsProvider`, a single `FormattingCache`, and parallel formatting via `Task.WhenAll`.

### Solution

Individual file paths are now collected into a pending list during the input loop. After all inputs are processed:

1. A **common ancestor directory** is computed from all file paths using a new `GetCommonAncestor` helper
2. A **single `OptionsProvider`** is created, pre-seeded at the common ancestor. Config resolution still walks up from each file's own directory via its internal `ConcurrentDictionary` caches, so per-directory configs (e.g. different `.csharpierrc` in different subdirectories) are respected correctly.
3. A **single `FormattingCache`** is shared across all files
4. All files are formatted **in parallel** via `Task.WhenAll`, matching the directory code path

Directory inputs continue to work exactly as before.

### Tests Added

- `Multiple_Files_Should_Be_Formatted` — verifies files from different subdirectories are formatted
- `Multiple_Files_Should_Respect_Per_Directory_Config` — verifies per-directory `.csharpierrc` is respected when using a shared `OptionsProvider`
- `Mixed_Files_And_Directory_Should_Format_All` — verifies mixed directory + individual file inputs
- `GetCommonAncestor_Returns_Common_Path` — 5 parameterized cases covering same dir, nested, divergent, root-only, and empty-string fallback

### DIFF
--- a/Src/CSharpier.Cli/CommandLineFormatter.cs
+++ b/Src/CSharpier.Cli/CommandLineFormatter.cs
@@ -178,6 +178,8 @@ internal static class CommandLineFormatter
             writer = new FileSystemFormattedFileWriter(fileSystem);
         }
 
+        var pendingFiles = new List<(string DirectoryName, string ActualPath, string OriginalPath)>();
+
         for (var x = 0; x < commandLineOptions.DirectoryOrFilePaths.Length; x++)
         {
             var directoryOrFilePath = commandLineOptions.DirectoryOrFilePaths[x];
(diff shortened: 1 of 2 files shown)


belav:
This adds a lot of complexity when it seems like a simpler solution is to just call `dotnet csharpier format .` in the pre-commit hook if there are more than x file paths.

PR #1830, opened 2026-03-10
allow checking formatting with cache

sd-st:
## Description

I run csharpier check as part of my lint flow, which means I run it a lot. On one of my repositories it takes 200ms to run format and 8000ms to run check. This PR allows for `check` to use the cache (although it can be disabled, just like for `format`)

### Checklist

  - always `var`
  - always `this.`
  - no pointless comments
- [ ] I have added tests that cover my changes

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -9,10 +9,12 @@ body:
         Thanks for taking the time to report a bug! Please fill out the information below to help us investigate.
 
         **Before submitting**, please check:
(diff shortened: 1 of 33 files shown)


sd-st:
I think there's probably nothing to test here; it doesn't look like there were tests for the previous behaviour (and no tests for the CLI arguments in general as far as I can tell)

sd-st:
I do think it's a little unfortunate to have a different API surface between checking and formatting but I made that change

sd-st:
Thanks! I'm planning to try it out in CI so I'll report back if there are any issues.

PR #1831, opened 2026-03-10
Fix: ignore formatting comments in switch sections

### LINKED ISSUE #1824: [Bug]: Unable to ignore csharpier indentation on multi-line verbatim strings
### Description

When formatting my code, I have some verbatim strings where the first line is being indented in a way I don't want and adding `// csharpier-ignore` or `// csharpier-ignore-start` doesn't skip over over that block.

I can't use raw string literals because this file is in project still targeting .NET 4.7.2 and raw string literals are not available in C# 7.3.

### Steps to Reproduce

Have a file with the following content:

```cs
namespace StepDefinitions
{
    using TechTalk.SpecFlow;

    [Binding]
    public class MyTestSteps
    {
        public void IHaveAConfiguration()
        {
            // csharpier-ignore-start
            var table = CommonStepHelpers.TableFromString(
@"			| Field         | Value |
			| ValueA        | 1     |
			| ValueB        | AC08  |"
            );
            // csharpier-ignore-end

            // ... do things with table
        }
    }
}
```

Run csharpier on it.

### Expected Behavior

Using `// csharpier-ignore` should work, and so should the `// csharpier-ignore-start` and `// csharpier-ignore-end` pair.

### Actual Behavior

When ran through csharpier the first line of the string becomes indented and making the table no longer line up like this:

```cs
            // csharpier-ignore-start
            var table = CommonStepHelpers.TableFromString(
                @"			| Field         | Value |
			| ValueA        | 1     |
			| ValueB        | AC08  |"
            );
            // csharpier-ignore-end
```

ngbrown:
## Description

When code is in a `switch` case statement group (or `SyntaxKind.SwitchSection`), the ignore comments should work, both range and one-line. This PR adds a failing test, and then resolves the issue.

## Related Issue

Fixes #1824

### Checklist

  - always `var`
  - always `this.`
  - no pointless comments

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -9,10 +9,12 @@ body:
         Thanks for taking the time to report a bug! Please fill out the information below to help us investigate.
 
         **Before submitting**, please check:
-        - I've searched [existing issues](https://github.com/belav/csharpier/issues) to ensure this isn't a duplicate
-        - I've read the [documentation](https://csharpier.com/) and this isn't expected behavior
-        - I'm using the latest version of CSharpier ![CSharpier](https://img.shields.io/nuget/v/CSharpier)
-        - I'm using the latest version of CSharpier's IDE plugin and understand it is versioned independently of CSharpier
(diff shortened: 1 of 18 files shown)


belav (approved):
Looks good, I like that you ensured the `// csharpier-ignore` would only ignore the first statement in a case section.

PR #1848, opened 2026-04-16
fix: Tolerate missing directories when formatting from stdin

jennings:
When formatting code passed via stdin, and using `--stdin-path`, the path passed may not exist. This is the case when using CSharpier as a [jj] fix tool: Commits are not checked out, so the directory containing the file being formatted may not actually exist on disk.

When this happened, CSharpier was throwing a DirectoryNotFoundException from FindForDirectoryName in CSharpierConfigParser.

Instead, we can predict this exception and check for configuration in the parent directory, until we find one that does exist.

[jj]: https://jj-vcs.dev

### Checklist

  - always `var`
  - always `this.`
  - no pointless comments


belav (approved):
Looks good, thanks for the contribution!

Issue #1851, opened 2026-04-21
[Feature]: Error output should be be written to stderr


jennings:
### Description

I use CSharpier as a [jj fix tool](https://www.jj-vcs.dev/latest/cli-reference/#jj-fix), which means the current file is piped to CSharpier over stdin and the formatted result read from stdout.

If a file is malformed XML, CSharpier writes the following to stdout and CSharpier exits with code 0, causing jj to replace the entire file contents with this warning:

    Warning <path> - Appeared to be invalid xml so was not formatted.

I'm happy to submit a PR to change this, I just want to make sure which of these cases would be welcome changes:

- (Required for my use case) Exit with a non-zero exit code. I would just pick code `1` unless you want a different value (I don't think the CLI has a convention for this yet?)
- (Optional but I strongly recommend) Print errors and warnings to stderr, not stdout.
- (Not required for my particular use case, but I noticed this while testing this changed behavior) Write the "Formatted N files in Xms." message to stderr as well.

jennings:
Hm, I was going to fix this by making this change:

```diff
-fileIssueLogger.WriteWarning(codeFormattingResult.WarningMessage);
+fileIssueLogger.WriteError(codeFormattingResult.WarningMessage);
+Interlocked.Increment(ref commandLineFormatterResult.FailedCompilation);
```

But then I found the level checks in ConsoleLogger which writes errors and greater to stderr, and anything else to stdout.

My expectation was that CSharpier would use stderr and stdout in the Unix tradition:

- The data you're running the program for goes to stdout (in this case, the formatted file contents)
- All diagnostic information goes to stderr, regardless of "severity".

This works well because it means if you create a pipeline like `echo "$SOME_CODE" | csharpier format --stdin-path Foobar.csproj | some_other_program`, diagnostics always end up in the terminal and never get passed to `some_other_program`.

Would changing CSharpier to use stderr and stdout this way be welcome? Edit: I opened #1852 for this, if it's unwelcome it can just be closed, but I would be sad.

belav:
For a reason that I can't currently remember, I decided that invalid xml should be a warning while invalid csharp is an error. I think it was because there are some file extensions like `.config` that are xml in some codebases but not in others.

I think changing that, at least for when piping content to std-in, would solve your issue without changing the behavior for anyone else. I'm worried logging warnings to stderr would be a breaking change.

[...]

PR #1852, opened 2026-04-21
Write all log messages to stderr

jennings:
## Description

Diagnostic output is often expected to go to standard error, leaving standard output for program output (which is the formatted file when using `--write-stdout` or reading from stdin).

Changes ConsoleLogger to write all log messages to stderr, and updates all tests to match.

## Related Issue

Related to #1851, but isn't the complete fix (that issue needs the exit code changed)

### Checklist

  - always `var`
  - always `this.`
  - no pointless comments


jennings:
It occurs to me that me that [many people expect well-behaved daemons to log to stdout](https://12factor.net/logs). Perhaps logs should go to stderr when running as a program but to stdout when running as a server?

Issue #1627, opened 2025-06-09
.gitignore from parent folders impacts formatting of children


yonguelink:
My `{Home}` folder, on windows, is managed via a git repository so I have the following file structure:

`{Home}/.git/...`
`{Home}/.gitignore`
`{Home}/repos/{bunch of repositories}`

The `.gitignore` in the `{Home}` folder contains `repos` to avoid committing stuff from the repositories I have on different machines into this "Home" repo.

Since I updated to csharpier 1.0.2 (I went from `0.30.x` to `1.0.2`), all my files are ignored by csharpier:

![Image](https://github.com/user-attachments/assets/68847f8e-c0a4-4730-b608-3dbca53887f5)

After digging around, it's the `repos` in the `.gitignore` of my `{Home}` folder that's causing the issue.

**Environments:**

- CSharpier Version: 1.0.2
- Running CSharpier via: dotnet cli
- Operating System: Windows
- .csharpierrc Settings: Default
- .editorconfig Settings: Default

**Steps to reproduce:**

- Have a `.gitignore` somewhere in the directory chain of your files
- Said `.gitignore` ignores the folder containing your files
- Run the format or check commands

Or check the content of the following zip: 

[csharpier-reproduction.zip](https://github.com/user-attachments/files/20658774/csharpier-reproduction.zip)

If you uncomment the `repos` line in the root's `.gitignore` the formatting of the `/repos/demo.cs` file works.

**Expected behavior:**

The files should be formatted/checked.

I suggest that once a `.gitignore` file is found, to stop going further up in the chain.

**Actual behavior:**

The files are all ignored

belav:
Gitignore is designed to have a hierarchy. If csharpier were to stop processing files when it found the first one it would be a non standard implementation.

You can modify the csharierignore file to specially include any files that are being ignored by your root gitignore. The csharierignore file takes priority over any gitignore file.

yonguelink:
> If csharpier were to stop processing files when it found the first one it would be a non standard implementation.

I agree stopping at "the first one" is too simplistic, but git itself doesn't get confused by nested gitignore files in multiple repositories. I feel like this behavior should be applied to everything that uses the gitignore file. So it's more "if we find a `.git` folder (I assume that's not configurable in git? ...) we stop searching for more `.gitignore` files".

> You can modify the csharierignore file to specially include any files that are being ignored by your root gitignore. The csharierignore file takes priority over any gitignore file.

[...]

PR #1860, opened 2026-05-15
Stop .gitignore walk at git worktree boundary

Jacko1394:
# Issue
csharpier was not formatting any files, in a worktree under a .gitignored folder path

## Description

`IgnoreFile.FindIgnorePaths` walks up from the target directory collecting `.gitignore` files and stops when it encounters a `.git` directory. In a git worktree, `.git` is a *file* containing a `gitdir:` pointer rather than a directory, so the walk continued past the worktree root.

When the worktree lives at a path the parent repo's `.gitignore` excludes (e.g. tools that create worktrees inside an ignored folder), every file in the worktree is matched by the parent ignore rule and `dotnet csharpier format .` reports `Formatted 0 files`.

This change also stops the walk when `.git` exists as a file, matching git's own treatment of the worktree as a separate working tree. Added a regression test that mirrors the existing `Gitignore_Outside_Git_Is_Not_Used` but with `.git` as a file.

## Related Issue

Follow-up to #1627 / #1649, which introduced the `.git` directory boundary but did not cover worktrees.

### Checklist

  - always `var`
  - always `this.`
  - no pointless comments

### DIFF
--- a/Src/CSharpier.Cli/IgnoreFile.cs
+++ b/Src/CSharpier.Cli/IgnoreFile.cs
@@ -108,7 +108,8 @@ internal class IgnoreFile
 
     // this will return the ignore paths in order of priority
     // the first csharpierignore it finds at or above the path
-    // and then all .gitignores (at or above stopping once it encounters a .git directory) it finds in order from closest to further away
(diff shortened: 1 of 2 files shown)


belav (approved):
Thanks for the fix!

PR #1868, opened 2026-06-04
fix: preserve blank lines before csharpier-ignore comments

### LINKED ISSUE #1867: [Bug]: csharpier-ignore comment removes linespaces before block
### Description

When I use `// csharpier-ignore`, the newline before the ignore comment is always removed. I would expect it to follow the same rules applied when the comment is not there.

### Steps to Reproduce

```csharp
void X(){
    int i = 1;


    int x=1;


    // csharpier-ignore
    int y=1;
}
```

### Expected Behavior

The new line should be preserved:
```csharp
void X()
{
    int i = 1;

    int x = 1;

    // csharpier-ignore
    int y=1;
}

```

### Actual Behavior

The new line is removed:
```csharp
void X()
{
    int i = 1;

    int x = 1;
    // csharpier-ignore
    int y=1;
}
```

nkgotcode:
## Summary

Fixes #1867

Preserves leading blank-line trivia for ignored statement nodes when formatting files that use `// csharpier-ignore` comments, and adds regression tests.

## Test plan

- [x] Updated `CSharpierIgnore` expected snapshots
- [x] Added `LineEndingEdgeCase` coverage
- [ ] `dotnet test` on `CSharpier.Tests`

Closes #1867

Made with [Cursor](https://cursor.com)


Src/CSharpier.Core/CSharp/SyntaxPrinter/CSharpierIgnore.cs
```diff
@@ -114,12 +119,30 @@ PrintingContext context
...
-            statements.Add(unFormattedCode.ToString().Trim());
+            statements.Add(
+                PrintWithoutFormattingDoc(firstUnformattedNode, unFormattedCode.ToString())
+            );
         }
 
         return statements;
     }
 
+    public static Doc PrintNodeWithoutFormatting(SyntaxNode syntaxNode, PrintingContext context)
```
belav: These three methods have almost the same name and I have no idea what the difference between them is and when to call which method.

Src/CSharpier.Tests/FormattingTests/LineEndingEdgeCase.cs
```diff
@@ -43,4 +43,82 @@ void MethodName()
             .Code.Should()
             .Be(unformattedCode.ReplaceLineEndings(endOfLine == EndOfLine.LF ? "\n" : "\r\n"));
     }
+
+    [Test]
+    public async Task CSharpier_Ignore_Preserves_Preceding_Blank_Line()
```
belav: These tests have no reason to be in this class if all they are doing is ensuring that a blank line is not removed. That is what the `csharpierignore.test` file is for.

belav (changes requested):

nkgotcode:
Pushed e177a63: renamed the unformatted-print helpers to distinguish Doc wrapping from raw code normalization, and moved the blank-line regression cases into `Src/CSharpier.Tests/FormattingTests/TestFiles/cs/CSharpierIgnore.test` / `.expected.test`. I could not run `dotnet test` in this container because the repo pins SDK `10.0.201` and only `8.0.421` is installed here.

PR #1870, opened 2026-06-05
perf: use string interpolation for string creation

TimothyMakkison:
Use string interpolation to create a string in `PrintingContext.GroupFor`. Very small improvement


belav:

PR #1876, opened 2026-06-13
Avoid chopping single trailing lambda

### LINKED ISSUE #1865: [Feature]: Avoid chopping single trailing lambda
### Description

This may be a duplicate of #1613 but after reading that ticket, I am having a hard time understanding it's intentions. 

The behavior I am after is:
> When a method call is too long, one parameter is a lambda, and the last parameter is a lambda: prefer adding a new line somewhere after the `=>` instead of chopping the parameters

When I tried to introduce CSharpier to my company this was the only issue raised. Everything else was a nitpick but the lambda-configuration formatting was seen as "icky". I believe the above rule is sufficient to cover the case and could apply to expression-bodied and method-bodied lambdas just fine.

## Example

Often, this shows up with configuration or fluent libraries. You'll provide 0-4 parameters and then end with a lambda to do "whatever else". This pattern is very common:

- [`IConfiguration.Get<T>`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationbinder.get?view=net-11.0-pp#microsoft-extensions-configuration-configurationbinder-get-1(microsoft-extensions-configuration-iconfiguration-system-action((microsoft-extensions-configuration-binderoptions))))
- [`EntityTypeBuilder.OwnsMany`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.metadata.builders.entitytypebuilder-1.ownsmany?view=efcore-10.0#microsoft-entityframeworkcore-metadata-builders-entitytypebuilder-1-ownsmany(system-string-system-string-system-action((microsoft-entityframeworkcore-metadata-builders-ownednavigationbuilder))))
- [`HostApplicationBuilder.ConfigureContainer`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.hostapplicationbuilder.configurecontainer?view=net-11.0-pp#microsoft-extensions-hosting-hostapplicationbuilder-configurecontainer-1(microsoft-extensions-dependencyinjection-iserviceproviderfactory((-0))-system-action((-0))))

My personal gripe is with a state machine library which would greatly suffer if methods were used to configure states and transitions. Keeping the configuration in one solid chunk makes it easy to read it instead of scrolling to other methods.

### Current Behavior

A simple state machine with that library is currently formatted with CSharpier as:

```csharp
var stateMachine = StateMachine
    .Configure<string, string>()
    .WithInitialState("A", 1)
    .WithState(
        "A",
        state =>
            state
                .WithParameter<int>()
                .WithTransition(
                    "To B",
                    t =>
                        t.WithMappedParameter<string>(value => value.ToString())
                            .If(mapped => mapped.Length > 0)
                            .To("B")
                )
    )
    .WithState("B", state => state.WithParameter<string>())
    .Build();
```

There is a _ton_ of unnecessary whitespace being added due to aggressive chopping which directly makes the state machine harder to read. 

### Better Behavior

Ideally, the state machine configuration could just split the lambdas before chopping, which prevents long lines while still being easy to read. I'm not sure if this is _exactly_ how it would look, but the idea of not chopping the parameters already saves bloating whitespace:

```csharp
var stateMachine = StateMachine
    .Configure<string, string>()
    .WithInitialState("A", 1)
    .WithState("A", state =>
        state
            .WithParameter<int>()
            .WithTransition("To B", t => 
                t
                    .WithMappedParameter<string>(value => value.ToString())
                    .If(mapped => mapped.Length > 0)
                    .To("B")
            )
    )
    .WithState("B", state => state.WithParameter<string>())
    .Build();
```

### Ideal Behavior

Depending on how #1676 works out this could even reduce indentation more:

```csharp
var stateMachine = StateMachine
    .Configure<string, string>()
    .WithInitialState("A", 1)
    .WithState("A", state => state
        .WithParameter<int>()
        .WithTransition("To B", t => t
            .WithMappedParameter<string>(value => value.ToString())
            .If(mapped => mapped.Length > 0)
            .To("B")
        )
    )
    .WithState("B", state => state.WithParameter<string>())
    .Build();
```

## Out of Scope

It feels unreasonable to do this when there are multiple lambdas or parameters after the lambda.

With multiple lambdas you're probably always better off chopping because the API was designed with shorter lambdas in mind (not long configuration chains). For example, this looks great and I **wouldn't** want this to be changed:

```csharp
messageBus.Subscribe(
    "orders/created",
    Retries.AtLeastOnce,
    e => e.Connected += OnConnected,
    e => e.Disconnected += OnDisconnected
);
```

If there are parameters after the lambda (typical for a `CancellationToken`) then chopping is almost always fine too. I feel like this is outside the scope of a linter and the writer should just create a method - there is often little readability benefit to using a lambda. This is typical for `InvokeAsync` or `Task.Run` sort of scenarios:

```csharp
Task.Run(
    () =>
    {
        var order = await this.orderRepository.GetOrder(orderId, token);
        var orderUpdated = new OrderUpdatedEvent { Order = order };
        await this.messageBus.Publish(orderUpdated, token);
    },
    token
);

// Just use a method group!
Task.Run(PublishOrderUpdatedEvent, token);
```

## Next Steps

If this seems like this has potential I could fork CSharpier and (try to) implement this. I am unfamiliar with the code base but it feels like the [`ArgumentListLikeSyntax`](https://github.com/belav/csharpier/blob/f8e15a29c09bfc59d31db60d7076e2420a06b8a8/Src/CSharpier.Core/CSharp/SyntaxPrinter/ArgumentListLikeSyntax.cs) already has handling of lambda-edge-cases and may be a suitable place to start with. 

P.S: I absolutely love this tool and feel that it has saved me countless hours. I will continue to use this personally because it just makes programming so much easier.

mattwzawislak:
## Description

Adds in the ability to avoid chopping single trailing lambdas when the argument list up to and including the "=>" symbol doesn't need to be chopped.

This would take a snippet of code like 

```csharp
logging.AddFile(
    Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "server{0}.log"),
    o =>
    {
        o.MaxRollingFiles = 1;
        // ...
    }
);
```

And turn it into this (assuming configured line length isn't exceeded):

```csharp
logging.AddFile(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "server{0}.log"), o =>
{
    o.MaxRollingFiles = 1;
    // ...
});
```

Without breaking directives or comments:
```csharp
logging.AddFile(
#if DEBUG
    Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "server{0}.log"),
#else
    "server{0}.log",
#endif
    o =>
    {
        o.MaxRollingFiles = 1;
        // ...
    }
);
```

And avoids chopping when the lambda is not the last argument or there are multiple arguments:
```csharp
logging.AddFile(
    "server{0}.log",
    cfg => cfg.Section("logging"),
    o =>
    {
        o.MaxRollingFiles = 1;
        // ...
    }
);
```

## Related Issue

Closes #1865

### Checklist

  - always `var`
  - always `this.`
  - no pointless comments


Src/CSharpier.Core/CSharp/SyntaxPrinter/ArgumentListWithTrailingLambda.cs
```diff
@@ -0,0 +1,75 @@
...
+            return chop;
+        }
+
+        var wrap = Doc.Concat(
+            Doc.ForceFlat(flatParts),
+            lambdaBody,
+            bodyIsBraced ? Doc.Null : Doc.SoftLine
+        );
+
+        return Doc.ConditionalGroup(wrap, wrap, chop);
```
belav, line 73: I think this should just be ` return Doc.ConditionalGroup(wrap, chop);`

Passing the same doc twice just ends up doing more work in the case where `chop` is what is used.

Src/CSharpier.Tests/FormattingTests/TestFiles/cs/ParenthesizedLambdaExpressions.test
```diff
@@ -104,8 +104,26 @@ public class ClassName
             CallAnotherMethod______________________________________________________123()
         );
 
+        CallMethod(CallAnotherMethod_________________(), () =>
+            CallAnotherMethod_________________()
+        );
```
belav, line 109: I don't agree with this breaking this way, and prettier also avoids breaking this.

```c#
// input & expected output
        var antiforgeryMiddleware = new AntiforgeryMiddleware(
            antiforgeryService.Object,
            hc => Task.CompletedTask
        );

// this pr
        var antiforgeryMiddleware = new AntiforgeryMiddleware(antiforgeryService.Object, hc =>
            Task.CompletedTask
        );
```

This is what I think should be breaking, and prettier does it this way. The pr works this way currently but I don't see a test for it.
```c#

// input & expected output
        Assert.Collection(receiveAuthStateDiff.Edits, edit =>
        {
            Assert.Equal(RenderTreeEditType.PrependFrame, edit.Type);
            AssertFrame.Text(
                batch.ReferenceFrames[edit.ReferenceFrameIndex],
                "Authenticated: True; Name: Bert; Pending: False; Renders: 1"
            );
        });
```

I just did a quick scan of https://github.com/belav/csharpier-repos/pull/172/changes and ran into that first case right away.

belav (changes requested):
The code looks good but I do disagree with the simple lambdas breaking. If you can make that change I'll redo the csharpier-repos PR to see if there are any edgecases that need to be handled.

Thanks for taking this on!

Src/CSharpier.Core/CSharp/SyntaxPrinter/ArgumentListWithTrailingLambda.cs
mattwzawislak: This is because the first `wrap` is `PrintMode.Flat` and the second is `PrintMode.Break` where it breaks the lambda. `DocPrinter.ProcessGroup` skips the first option for a `ConditionalGroup`

Src/CSharpier.Tests/FormattingTests/TestFiles/cs/ParenthesizedLambdaExpressions.test
mattwzawislak: This is an important part of this PR since it is where the lambdas will usually start to blow up:

Before:
```csharp
public void ConfigureOptions()
{
    services.Configure<MyOptions>(
        "Primary",
        options =>
            options.SetName("primary").SetTimeout(TimeSpan.FromSeconds(30)).EnableRetries()
    );
}
```

With this change:
```csharp
public void ConfigureOptions()
{
  services.Configure<MyOptions>("Primary", options =>
      options.SetName("primary").SetTimeout(TimeSpan.FromSeconds(30)).EnableRetries()
  );
}
```

In either way, CSharpier will start to add line breaks if the chaining goes on too long but the start of that method call is what changes and the method calls are on 1 lower indent level.

Is there a point where you think the body of the lambda is large enough to justify wrapping instead of chopping?

Src/CSharpier.Core/CSharp/SyntaxPrinter/ArgumentListWithTrailingLambda.cs
belav: Ah yeah, basically all of that conditional group code was ported from prettier and I had completely forgotten how it actually gets processed.

Src/CSharpier.Tests/FormattingTests/TestFiles/cs/ParenthesizedLambdaExpressions.test
belav: Poking around prettier some more it looks like they are actually inconsistent with how they deal with lambdas depending on if the lambda ends in a method or a property.

```c#
new AntiforgeryMiddleware(
  antiforgeryService.Object,
  (hc) => Task.CompletedTask____________________________________________,
);

new AntiforgeryMiddleware(antiforgeryService.Object, (hc) =>
  Task.CompletedTask____________________________________________(),
);
```

And with your example this pr definitely improves things.

I think ideally csharpier would do this - but I don't know that it is possible. It would need to understand that if the lambda is long enough that it is going to break after the `=>` to not break before `options`

```c#
services.Configure<MyOptions>(
    "Primary",
    options => options.SetName("primary").SetTimeout(TimeSpan.FromSeconds(30)).EnableRetries()
);

services.Configure<MyOptions>("Primary", options => 
    options.SetName("primary").SetTimeout(TimeSpan.FromSeconds(30)).EnableRetries________()
);
```

I think I've been convinced. Let me just review more of the csharpier-repos code to see if there are any other edge cases to look at.

mattwzawislak: I like your idea about the the "if the lambda is long enough that it is going to break after the =>". I could try playing around with this to see if that is reasonable. My biggest concern is perf - already this change has impacted the perf since it is trying a lot of options so it might be not too much extra time spent.

belav: I took a look through that large PR again and I think it is good as is. I don't know that it is worth the extra perf impact to do, and I'm still unsure how it would work.

belav (approved):

Issue #1856, opened 2026-05-02
[Feature]: Ignore using directives sorting


Lukis90:
### Description

If I use `// csharpier-ignore-start/end` block on using directives, the directives will still be sorted. We have provisioned teplate code, that has using directives provisioned based on project name and the sorting for those using directives can differ everytime. Is there a reason why block ignore is not implemented for using directives?

Currently

```
// csharpier-ignore-start
using CC.test;
using BB.test;
using AA.test;
// csharpier-ignore-end

public class ClassName {
    public string ShortPropertyName {
        get;
        set; 
    }

    public void LongUglyMethod(string longParameter1, string longParameter2, string longParameter3) { 
        this.LongUglyMethod("1234567890", "abcdefghijklmnopqrstuvwxyz", "ABCDEFGHIJKLMNOPQRSTUVWXYZ");
    }
}
```
will be formatted to:
```
// csharpier-ignore-start
using AA.test;
using BB.test;
using CC.test;

// csharpier-ignore-end

public class ClassName
{
    public string ShortPropertyName { get; set; }

    public void LongUglyMethod(string longParameter1, string longParameter2, string longParameter3)
    {
        this.LongUglyMethod(
            "1234567890",
            "abcdefghijklmnopqrstuvwxyz",
            "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
        );
    }
}
```
I was expecting:
```
// csharpier-ignore-start
using CC.test;
using BB.test;
using AA.test;
// csharpier-ignore-end

public class ClassName
{
    public string ShortPropertyName { get; set; }

    public void LongUglyMethod(string longParameter1, string longParameter2, string longParameter3)
    {
        this.LongUglyMethod(
            "1234567890",
            "abcdefghijklmnopqrstuvwxyz",
            "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
        );
    }
}
```

PR #1877, opened 2026-06-13
feat: added support for ignoring using directives

TimothyMakkison:
Added support for using ranged ignores in using directives solving #1856

- The logic is pretty messy, but it seems to play well with `#if` directives
- This does not support ranged ignores which bridge the usings section and the code
  - This might be an easy addition by adding a field to `PrintingContext`


belav (approved):
> This does not support ranged ignores which bridge the usings section and the code

I believe that limitation exists almost everywhere. I assume it will get pretty messy to make sure it works everywhere and I don't know that it is worth it.

Thanks for getting this figured out!

PR #1881, opened 2026-07-03
Support closed and union syntax via preview Roslyn

### LINKED ISSUE #1880: [Feature]: Support `closed` modifier
C# 15 adds the [closed](https://learn.microsoft.com/en-in/dotnet/csharp/language-reference/keywords/closed) modifier

- `SyntaxPrinter/Modifiers.cs` need to support ordering `closed`, I can't find any documentation online about how Microsoft recommends ordering this
- A test needs to be added to verify that this works

Probably a good first issue

wuzzeb:
## Description

Add closed to modifier ordering and route Roslyn's preview UnionDeclaration syntax kind through the existing base type declaration printer. Add formatting coverage for closed record classes and union declarations.

For modifier ordering, the current IDE0036 default-order documentation does not yet list the new closed modifier, so this PR makes a local ordering choice rather than copying an official one. I placed closed with the inheritance/type-shape modifiers, directly after abstract and before sealed, because the C# preview docs describe closed as a class modifier for closed hierarchies, and it is related to inheritance constraints. That keeps it near the modifiers it semantically overlaps with while leaving the existing documented order otherwise unchanged.

## Microsoft.CodeAnalysis.CSharp

Unfortunately, the published nuget for Microsoft.CodeAnalysis.CSharp does not yet support either closed or union.  To implement it temporarily, I reference the packages from the .NET 11 preview 5 SDK.  Thus, we should not merge this PR as-is.  The idea is to just have something ready. 

When Microsoft.CodeAnalysis.CSharp is updated on nuget, we can take out the `RoslynPreviewSdkReferences.props` file and all the overrides and just use the nuget package, and then the PR will be ready to merge.

This draft intentionally references the .NET 11 Preview 5 SDK Roslyn assemblies for net10.0 builds because matching Microsoft.CodeAnalysis.CSharp NuGet packages are not yet published. Replace RoslynPreviewSdkReferences.props with a normal NuGet package update before merging.

## Related Issue

Fixes #1880 

## Checklist

  - always `var`
  - always `this.`
  - no pointless comments

### DIFF
--- a/RoslynPreviewSdkReferences.props
+++ b/RoslynPreviewSdkReferences.props
@@ -1,0 +1,19 @@
+<Project>
+  <PropertyGroup>
+    <RoslynPreviewSdkPath>/usr/share/dotnet/sdk/11.0.100-preview.5.26302.115/Roslyn/bincore</RoslynPreviewSdkPath>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <!--
+      DRAFT PR HACK: use the Roslyn bits from .NET 11 Preview 5 so CSharpier can
+      parse C# 15 closed/union syntax before matching NuGet packages are published.
+      Replace this import with Microsoft.CodeAnalysis.CSharp from NuGet once available.
+    -->
+    <Reference Include="Microsoft.CodeAnalysis">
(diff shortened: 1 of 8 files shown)


wuzzeb:
Also, ` SyntaxKind.UnionDeclaration` does not yet exist in the preview 5 version, so had to temporarily use the cast string.... but the parsing and formatting does work (see the test)

belav:
This is being done in #1889

PR #1896, opened 2026-08-06
feat: refactored benchmarks and added `CliBenchmarks`

TimothyMakkison:
- Added `CliBenchmarks` containing benchmarks for formatting a directory and checking a list of files.
  - Uses the `CSharpier` project to format and check. This is more of a benchmark to measure the time to get the files as opposed to a formatting benchmark as the files will become cached.
  - As Soda found in #1893, `CSharpier` is very bad at handing multiple files, each iteration of `CliBench` took 50 seconds for me 😑
- `Benchmarks` is getting cluttered so I split it up into `CSharpBenchmarks` and `XmlBenchmarks`

Relevant to #1893


TimothyMakkison:
Should work, but I'm a little confused.

When I run `dotnet csharpier format .` at the root of my csharpier repo it takes 3 seconds, even when the files haven't changed and thus should be cached
`Formatted 447 files in 3881ms.`

`FormatCli` should be identical and yet it runs faster than in terminal
`Formatted 447 files in 872ms.`

I assume this discrepancy is because the method is allowed to warmup beforehand 🤔

TimothyMakkison:
Renamed the benchmarks and added `Format` and `FormatNoCache`

belav (approved):

PR #1897, opened 2026-08-08
perf: remove string allocation using `Replace(char` overload

TimothyMakkison:
Each call to `StringExtensions.NormalisePath` would allocate a new `string` of `Path.DirectorySeparatorChar`
From profiling `csharpier format` on the csharpier repo #1896
image


TimothyMakkison:
Closed in favour of #1898

Issue #1888, opened 2026-07-17
Support for net11 and c# 15. Drop net8/9


belav:
Collection expression arguments
Union types
Closed modifier

TimothyMakkison:
They [recently added](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/labeled-break-continue) labeled `continue` and `break` statements, curious to see how Roslyn handles this

PR #1898, opened 2026-08-08
perf: refactor ignore logic to prevent duplicate work and allocations

TimothyMakkison:
Builds on #1897, supports #1893
#### IgnoreList
- Convert `alwaysIgnoredRules` to use `IgnoreRule[]`, avoiding the cost of compiling the regex for each construction of `IgnoreList`
- Instead of concatenating `alwaysIgnoredRules` and the parsed rules we simply evaluate one after the other to avoid allocating a `List` or larger array

#### IgnoreRule
- Use `TrimStart(ReadOnlySpan<char>` overload to avoid reallocating a `string` for every call
- Use preprocessor to use the more modern `Split` api that returns a `Range` instead of creating substrings.
Related #1896

#### StringExtensions
- Keeps the change from #1897, avoids allocating a new string each call

### Benchmarks - from #1896
Note that timing isn't very accurate due to low iteration count
#### Before

Method | Mean | Error | StdDev | Gen0 | Gen1 | Gen2 | Allocated
-- | -- | -- | -- | -- | -- | -- | --
FormatCli | 737.4 ms | 116.9 ms | 30.36 ms | 6000.0000 | - | - | 57.53 MB
CheckFiles | 43,895.1 ms | 639.7 ms | 98.99 ms | 53000.0000 | 52000.0000 | 2000.0000 | 459.28 MB

#### After

Method | Mean | Error | StdDev | Gen0 | Gen1 | Gen2 | Allocated
-- | -- | -- | -- | -- | -- | -- | --
FormatCli | 643.6 ms | 63.36 ms | 16.46 ms | 2000.0000 | - | - | 24.9 MB
CheckFiles | 33,135.1 ms | 842.61 ms | 130.39 ms | 45000.0000 | 44000.0000 | 2000.0000 | 393.78 MB

Pretty large performance gains for `FormatCli`, curious to see how large the real world impact will be.

### DIFF
--- a/Src/CSharpier.Cli/DotIgnore/IgnoreList.cs
+++ b/Src/CSharpier.Cli/DotIgnore/IgnoreList.cs
@@ -8,14 +8,14 @@ namespace CSharpier.Cli.DotIgnore;
 
 internal class IgnoreList(string basePath)
 {
-    private static readonly string[] alwaysIgnoredText =
+    private static readonly IgnoreRule[] alwaysIgnoredRules =
     [
-        "**/bin",
-        "**/node_modules",
-        "**/obj",
-        "**/.git",
+        new("**/bin"),
+        new("**/node_modules"),
+        new("**/obj"),
+        new("**/.git"),
     ];
-    private readonly List<IgnoreRule> rules = [];
+    private IgnoreRule[] rules = [];
 
     public static async Task<IgnoreList> CreateAsync(
         IFileSystem fileSystem,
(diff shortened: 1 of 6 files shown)


TimothyMakkison:
After #1888 it would be worth switching most of these methods over to using `ReadOnlySpan<char>` while using `GetAlternateLookup`. 
In `StringExtensions.NormalisePath` should be changed to accept `ReadOnlySpan<char>`, checking for `:` either allocating a new string or returning a trimmed value.
There are other areas where `SearchValues` could replace `string[]` for lookups

belav:

TimothyMakkison:
> CSharpier is my passion project. I don't enjoy reviewing code and trying to get it to conform to what I'd prefer to see. I much prefer to spend my free time working on CSharpier myself. As such I will no longer be accepting external contributions. Feel free to open an issue for any problems you encounter.

Fair enough, most of my other changes were fairly small, but please consider merging or looking into this PR. This does have some fairly substantial real world improvements. The relevant changes are the four files below. Together they do noticeably speed up the format command.

At the the very least convert `string[] alwaysIgnoredText` into `IgnoreRule[] alwaysIgnoredRules` 

image

PR #1899, opened 2026-08-08
perf: move `IsGeneratedCodeFile` to `Cli` project and use `Path(Span)` methods

TimothyMakkison:
## Description
- Move `IsGeneratedCodeFile` method into `CSharpier.Cli` as its the only project that calls it (and is .NET only avoiding the need for preprocessors)
- Change the method signature to `IsGeneratedCodeFile(ReadOnlySpan<char> filePath)`, this is subtle, but it lets the methods like `Path.GetFileName` use the `ReadOnlySpan<char>` overload so this entire method goes from allocating several substrings to allocating zero objects

### Profiling Check of the Csharpier repo
image

## Related Issue
#1898


belav:

